PR #53, opened 2019-09-23
'Solution'

Serhii-P:


Serhii-P:
https://serhii-p.github.io/react_add-todo-form/

irenhh:
as soon as I change value of an input, the error caused by empty submit should disappear, not it doesn't

PR #55, opened 2019-09-27
Develop

ibezkorovayniy:
https://ibezkorovayniy.github.io/react_add-todo-form/

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -2,18 +2,83 @@ import React from 'react';
 import './App.css';
 
 import users from './api/users';
+import todos from './api/todos';
+import TodoList from './components/TodoList/TodoList';
+import NewTodo from './components/NewTodo/NewTodo';
 
-function App() {
-  return (
-    <div className="App">
-      <h1>Static list of todos</h1>
+class App extends React.Component {
+  state = {
+    todosWithUsers: todos.map(todo => (
+      {
+        ...todo,
+        user: users.find(user => user.id === todo.userId),
+      })),
+    showAddDialog: false,
+  };
 
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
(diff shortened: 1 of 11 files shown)


irenhh:
when I fill the input the error has to disappear immediately

now I can add empty task with spaces as its title

src/App.js
```diff
@@ -2,18 +2,83 @@ import React from 'react';
...
+
+  closeForm = ({ target }) => {
+    if (!target.closest('.add-button') && !target.closest('.addDialog')) {
+      this.setState({ showAddDialog: false });
+    }
+  };
+
+  render() {
+    const { todosWithUsers } = this.state;
+    const { showAddDialog } = this.state;
```
irenhh: add this destructuring to the prev one

src/components/NewTodo/NewTodo.js
```diff
@@ -0,0 +1,144 @@
...
+  state = initialState;
+
+  getFormValue() {
+    return Object.entries(this.state)
+      .reduce((acc, entry) => (
+        {
+          ...acc,
+          [entry[0]]: entry[1].value,
+        }
+      ), {});
```
irenhh:
```
return Object.entries(this.state)
  .reduce((acc, entry) => ({
    ...acc,
    [entry[0]]: entry[1].value,
  }), {});
```

src/components/required.js
```diff
@@ -0,0 +1,3 @@
+export const required = (name, value) => (value
+  ? null
+  : `Field ${name} is required`);
```
irenhh:
```
export const required = (name, value) => (
  value
    ? null
    : `Field ${name} is required`
);
```

irenhh: `required` is not a name for function

PR #130, opened 2020-02-21
Develop

SashaPylypyuk:
https://sashapylypyuk.github.io/react_add-todo-form/

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,17 +1,15 @@
 import React from 'react';
 import './App.css';
+import { TodoList } from './components/TodoList/TodoList';
 
 import users from './api/users';
+import todos from './api/todos';
 
 function App() {
   return (
     <div className="App">
       <h1>Add todo form</h1>
-
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
+      <TodoList users={users} todos={todos} />
     </div>
   );
 }
--- a/src/components/NewTodo/NewTodo.js
+++ b/src/components/NewTodo/NewTodo.js
@@ -1,0 +1,124 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export class NewTodo extends React.Component {
+  state = {
+    users: this.props.users,
+    title: '',
+    userId: 0,
+    id: 3,
+    errorTitle: false,
+    errorUser: false,
+  }
+
+  onSubmit = (e) => {
+    e.preventDefault();
+
+    const { userId, id, title } = this.state;
+
+    if (title === '') {
+      this.setState({
+        errorTitle: true,
+      });
+
+      return;
+    }
+
+    if (userId === 0) {
+      this.setState({
+        errorUser: true,
+      });
+
+      return;
+    }
+
+    const todo = {
+      userId, id, title,
+    };
+
+    this.props.addTodo(todo);
+    this.setState({
+      id: id + 1,
+      title: '',
+      userId: 0,
+    });
+  };
+
+  onChange = (e) => {
+    this.setState({
+      title: e.target.value,
+    });
+  };
+
+  onSelectName = (e) => {
+    this.setState({
+      userId: e.target.value,
+      errorUser: false,
+    });
+  };
+
+  clearErrorTitle = () => {
+    this.setState({
+      errorTitle: false,
+    });
+  }
+
+  render() {
+    const { users } = this.state;
+
+    return (
+      <>
+        <form onSubmit={this.onSubmit}>
+          <input
+            value={this.state.title}
+            onChange={this.onChange}
+            onSelect={this.clearErrorTitle}
+            type="text"
+            placeholder="Enter the name of todo"
+            className="form__input"
+          />
+          <select
+            onChange={this.onSelectName}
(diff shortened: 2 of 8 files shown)


src/App.js
```diff
@@ -1,17 +1,15 @@
...
 function App() {
   return (
     <div className="App">
       <h1>Add todo form</h1>
-
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
+      <TodoList users={users} todos={todos} />
```
irenhh: create one array with todo and corresponding user in each element

src/components/TodoList/TodoList.js
```diff
@@ -0,0 +1,53 @@
...
+  addTodo = (todo) => {
+    this.setState(prevState => ({
+      todos: [...prevState.todos, todo],
+    }));
+  }
+
+  render() {
+    return (
+      <>
+        <NewTodo users={this.state.users} addTodo={this.addTodo} />
```
irenhh: destructure this.state

src/components/TodoList/TodoList.js
```diff
@@ -0,0 +1,53 @@
...
+        <table className="table">
+          <thead>
+            <tr className="table__head">
+              <th>Id</th>
+              <th>Title</th>
+              <th>User Id</th>
+            </tr>
+          </thead>
+          {this.state.todos.map(todo => (
+            <tr className="table__row">
```
irenhh: where is key attribute?

src/components/NewTodo/NewTodo.js
```diff
@@ -0,0 +1,125 @@
...
+    errorUser: false,
+  }
+
+  onSubmit = (e) => {
+    e.preventDefault();
+
+    const { userId, id, title } = this.state;
+
+    if (title === '') {
+      this.setState({
```
irenhh: use setState once for one method

irenhh: write all the code inside setState callback

src/components/NewTodo/NewTodo.js
```diff
@@ -0,0 +1,125 @@
...
+      return;
+    }
+
+    const todo = {
+      userId, id, title,
+    };
+
+    this.props.addTodo(todo);
+    this.setState({
+      id: id + 1,
```
irenhh: don't mutate state

src/components/NewTodo/NewTodo.js
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  render() {
+    const { users } = this.state;
+
+    return (
+      <>
+        <form onSubmit={this.onSubmit} className="form">
+          <input
+            value={this.state.title}
```
irenhh: destructure

src/App.js
SashaPylypyuk: I don't understand why we need one array for todos and users. In "TodoList" we push users in "NewTodo" and must show them in the list, in array where we have todos and users it's hard make. I think we don't need one array

src/components/NewTodo/NewTodo.js
SashaPylypyuk: i can't validate properties in setState's callback

src/components/NewTodo/NewTodo.js
SashaPylypyuk: Image 1

src/components/NewTodo/NewTodo.js
irenhh: why? use prevState

src/components/NewTodo/NewTodo.js
irenhh: just prevState.id + 1

src/components/NewTodo/NewTodo.js
```diff
@@ -0,0 +1,118 @@
...
+    if (errorUser || errorTitle) {
+      return;
+    }
+
+    const todo = {
+      userId, id, title,
+    };
+
+    this.props.addTodo(todo);
+    this.setState(prevState => ({
```
irenhh: you're still using two setStates in one method

SashaPylypyuk: Так в цьому методі неможливо використовувати 1 setState, спочатку йде валідація потім добавлення в масив і потім обнулення через setState, чесно я не розумыю як тут використовувати 1 setState

irenhh:
```
  all your validations
  when you want to return new state object just write
  return {
    new state object
  }
})
```

src/components/NewTodo/NewTodo.js
```diff
@@ -0,0 +1,124 @@
...
+        return {
+          errorUser: true,
+        };
+      }
+
+      const todo = {
+        userId, id, title,
+      };
+
+      this.props.addTodo(todo);
```
irenhh: don't call addTodo in setState

irenhh: make a separate validation function and return some object with errors or just one value for example and then check it in this method and return setState with errors on addTodo in `else` statement

PR #148, opened 2020-05-02
solution

BorisovecViktor:
[DEMO LINK](https://BorisovecViktor.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -2,18 +2,47 @@ import React from 'react';
 import './App.css';
 
 import users from './api/users';
+import todosFromServer from './api/todos';
 
-function App() {
-  return (
-    <div className="App">
-      <h1>Add todo form</h1>
+import TodoList from './components/TodoList';
+import NewTodo from './components/NewTodo';
 
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
-    </div>
-  );
+class App extends React.Component {
+  state = {
+    todos: todosFromServer.map(todo => ({
+      ...todo,
+      user: users.find(user => user.id === todo.userId),
+    })),
+  }
+
+  addTodo = (id, title, completed, user) => {
+    this.setState(prev => ({
+      todos: [
+        ...prev.todos,
+        {
+          id,
+          title,
+          completed,
+          user,
+        },
+      ],
+    }));
+  }
+
+  render() {
+    const { todos } = this.state;
+
+    return (
+      <div className="App">
+        <div className="todo__add">
+          <NewTodo users={users} addTodo={this.addTodo} id={todos.length + 1} />
+        </div>
+        <div className="todo__list">
+          <TodoList todos={todos} />
+        </div>
+      </div>
+    );
+  }
 }
 
 export default App;
--- a/src/components/NewTodo.js
+++ b/src/components/NewTodo.js
@@ -1,0 +1,142 @@
+import React from 'react';
+import './NewTodo.css';
+import PropTypes from 'prop-types';
+
+class NewTodo extends React.Component {
+  state = {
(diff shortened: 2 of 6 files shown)


src/App.js
```diff
@@ -2,18 +2,47 @@ import React from 'react';
...
+        },
+      ],
+    }));
+  }
+
+  render() {
+    const { todos } = this.state;
+
+    return (
+      <div className="App">
```
irenhh, line 36: class names start with lowercase letter

src/components/NewTodo.js
```diff
@@ -0,0 +1,142 @@
+import React from 'react';
+import './NewTodo.css';
+import PropTypes from 'prop-types';
+
+class NewTodo extends React.Component {
+  state = {
+    id: this.props.id,
```
irenhh, line 7: generate an id when you create a new todo, if you pass it as props and write directly to the state, you cause two additional unnecessary renders

src/components/TodoList.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import './TodoList.css';
+import PropTypes from 'prop-types';
+
+class TodoList extends React.Component {
```
irenhh, line 5: do you really need class component here?

src/components/TodoList.js
```diff
@@ -0,0 +1,28 @@
...
+          <p>{todo.user.name}</p>
+          <p>{todo.completed ? 'completed' : 'not completed'}</p>
+        </div>
+      ))
+    );
+  }
+}
+
+TodoList.propTypes = {
+  todos: PropTypes.arrayOf(PropTypes.object).isRequired,
```
irenhh, line 25: describe the object that you get, you use its properties so it would be better to know what type they are of

src/components/NewTodo.js
```diff
@@ -0,0 +1,142 @@
...
+        <br />
+
+        <button type="submit">Add todo</button>
+      </form>
+    );
+  }
+}
+
+NewTodo.propTypes = {
+  users: PropTypes.arrayOf(PropTypes.object).isRequired,
```
irenhh, line 137: the same
you can create description once and export it to use somewhere else

src/components/NewTodo.js
```diff
@@ -0,0 +1,142 @@
...
+
+  handleSubmit = (e) => { // form submit
+    e.preventDefault();
+
+    if (this.state.title.length < 4) {
+      this.setState({
+        errorTitle:
+  <div className="todo__error">
+    Please enter the title
+  </div>,
```
irenhh, line 53: fix indentations

src/components/NewTodo.js
```diff
@@ -0,0 +1,142 @@
...
+      });
+    }
+
+    if (this.state.userId === '') {
+      this.setState({
```
irenhh, line 58: there is not a good idea to use several setStates in one method, it's an async function and you can get an unexpected result;
write setState once and put all your validations inside it

irenhh: like this
```
  if (...) {...}
  if (...) {...}
})
```

src/components/NewTodo.js
```diff
@@ -0,0 +1,142 @@
...
+  handleSubmit = (e) => { // form submit
+    e.preventDefault();
+
+    this.setState((prev) => {
+      const nextState = {};
+
+      if (prev.title.trim() === '') {
+        nextState.errorTitle
+          = `Please enter the title`;
+      }
```
irenhh, line 53:
```suggestion
      if (prev.title.trim() === '') {
        nextState.errorTitle = `Please enter the title`;
      }
```

PR #160, opened 2020-05-04
Drozhzhyn solution

DrozhzhynY:
- https://drozhzhyny.github.io/react_add-todo-form/

### DIFF
--- a/src/components/NewTodo/NewTodo.js
+++ b/src/components/NewTodo/NewTodo.js
@@ -1,0 +1,209 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import './NewTodo.css';
+
+class NewTodo extends React.Component {
+  state = {
+    title: '',
+    name: '',
+    completed: '',
+    userId: '',
+    id: this.props.todoId,
+    status: '',
+    showTitleError: false,
+    showUserError: false,
+    showStatusError: false,
+  }
+
+  addTodoText = (event) => {
+    this.setState(
+      {
+        title: event.target.value.replace(/[^a-zа-я0-9\s]/g, ''),
+        showTitleError: false,
+      },
+    );
+  }
+
+  addUser = (id) => {
+    const { users } = this.props;
+
+    this.setState(
+      {
+        userId: +id,
+        name: users.find(user => user.id === +id).name,
+        showUserError: false,
+      },
+    );
+  }
+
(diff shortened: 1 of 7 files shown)


src/components/NewTodo/NewTodo.js
```diff
@@ -0,0 +1,209 @@
...
+                  Please enter the title
+                </div>
+              )
+            }
+          </label>
+          <br />
+          <label className="form__item">
+            <span className="form__name">Chose user</span>
+            <select
+              className={showUserError
```
irenhh, line 145: use classnames lib for this purpose, it's really convenient to use classed dynamically with it
https://www.npmjs.com/package/classnames

src/components/NewTodo/NewTodo.js
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  render() {
+    const { users } = this.props;
+    const { title,
+      showTitleError,
+      userId,
+      showUserError,
+      status,
+      showStatusError } = this.state;
```
irenhh, line 118:
```suggestion
    const { 
      title,
      showTitleError,
      userId,
      showUserError,
      status,
      showStatusError 
    } = this.state;
```

src/components/NewTodo/NewTodo.js
```diff
@@ -0,0 +1,209 @@
...
+    this.setState(
+      {
+        userId: +id,
+        name: users.find(user => user.id === +id).name,
+        showUserError: false,
+      },
+    );
+  }
+
+  statusTodo = (event) => {
```
irenhh, line 39: name of methods or functions should be verbs

src/components/NewTodo/NewTodo.js
```diff
@@ -0,0 +1,209 @@
...
+    const target = event.target.value;
```
irenhh, line 40:
```suggestion
    const value = event.target.value;
```
`value` is a more logical option in this case

src/components/NewTodo/NewTodo.js
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    addTodos({
+      name,
+      title,
+      userId,
+      id,
+      completed,
+    });
+    this.reset();
```
irenhh, line 108: add new line before this call

PR #166, opened 2020-05-05
not good solution

MykolaMak:
[DEMO LINK](https://MykolaMak.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/Todo/Todo.js
+++ b/src/components/Todo/Todo.js
@@ -1,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import User from '../User/User';
+import './Todo.css';
+
+const Todo = ({ title, user }) => (
+  <>
+    <h3 className="todo__title">
+      {title}
+    </h3>
+    <User user={user} />
+  </>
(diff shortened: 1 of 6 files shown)


src/components/Todo/Todo.js
```diff
@@ -0,0 +1,20 @@
...
+    </h3>
+    <User user={user} />
+  </>
+);
+
+export default Todo;
+
+Todo.propTypes = {
+  title: PropTypes.string.isRequired,
+  user: PropTypes.shape({}).isRequired,
```
irenhh, line 19: describe shape or write `object`

src/components/TodoList/TodoList.js
```diff
@@ -0,0 +1,30 @@
...
+      </li>
+    ))}
+  </ul>
+);
+
+export default TodoList;
+
+TodoList.propTypes = {
+  list: PropTypes.arrayOf(
+    PropTypes.object.isRequired,
```
irenhh, line 28: describe that values of this object that you use here

PR #211, opened 2020-09-04
solution

Andrii256:
[DEMO LINK](https://andrii256.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,17 +1,110 @@
 import React from 'react';
 import './App.css';
+import ClassNames from 'classnames';
+import { TodoList } from './componenets/TodoList';
 
-import users from './api/users';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-function App() {
-  return (
+class App extends React.Component {
+  state = {
+    todos: todosFromServer.map(todo => ({
+      ...todo,
+      user: usersFromServer.find(user => user.id === todo.userId),
+    })),
+    isHidden: true,
+    errorMessage: '',
+  }
+
+  timeToShowError = false;
+
+  validation = (event) => {
+    const form = event.target.closest('#formAdder');
+
+    if (!form.title.value && this.timeToShowError) {
+      this.setState({
+        isHidden: false,
+        errorMessage: 'enter title',
+      });
+
+      return false;
+    }
+
+    if (!form.userSelect.value && this.timeToShowError) {
+      this.setState({
+        isHidden: false,
+        errorMessage: 'chose user',
+      });
+
+      return false;
+    }
+
+    this.setState({
+      isHidden: true,
+      errorMessage: '',
+    });
+
+    return true;
+  }
+
+  Add = (event) => {
+    event.preventDefault();
+
+    this.timeToShowError = true;
+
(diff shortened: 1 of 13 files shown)


leonbohdan (approved):
Interesting solution.

src/App.js
```diff
@@ -1,17 +1,110 @@
...
+      user: usersFromServer.find(user => user.id === todo.userId),
+    })),
+    isHidden: true,
+    errorMessage: '',
+  }
+
+  timeToShowError = false;
+
+  validation = (event) => {
+    const form = event.target.closest('#formAdder');
```
irenhh, line 22: we've talked about it, no DOM

src/App.js
```diff
@@ -1,17 +1,110 @@
...
+
+    this.setState({
+      isHidden: true,
+      errorMessage: '',
+    });
+
+    return true;
+  }
+
+  Add = (event) => {
```
irenhh, line 50: why uppercase?

src/App.js
```diff
@@ -1,17 +1,110 @@
...
+    event.preventDefault();
+
+    this.timeToShowError = true;
+
+    if (!this.validation(event)) {
+      return;
+    }
+
+    const newTodo = {
+      id: [...this.state.todos].sort((a, b) => b.id - a.id)[0].id + 1,
```
irenhh, line 60: it's too complicated solution for setting a new id, sorting is a heavy process

src/App.js
```diff
@@ -1,17 +1,110 @@
...
+  state = {
+    todos: todosFromServer.map(todo => ({
+      ...todo,
+      user: usersFromServer.find(user => user.id === todo.userId),
+    })),
+    isHidden: true,
+    errorMessage: '',
+  }
+
+  timeToShowError = false;
```
irenhh, line 19: why not in state?

irenhh (changes requested):
don't allow empty spaces

src/App.js
Andrii256: because it didnt work in state. I was looking for bug 2 hours and then I understood that there are bug in react

src/App.js
Andrii256: but it garanted that Id will unical

src/App.js
Andrii256: because 
image

irenhh: this is the name for the button, not the method

src/App.js
irenhh: use uuid library, this would be a better solution

irenhh: what if your array is of 1000 items?..

src/App.js
irenhh: what kind of bug was it?

Andrii256: i dont know, but calling of set state didnt work

Andrii256: https://youtu.be/FiLo9wXdehw

irenhh: this is because async setState

src/App.js
irenhh: not fixed

PR #276, opened 2020-10-30
solution

andlap24:
[DEMO LINK](https://andlap24.github.io/react_add-todo-form/)


casualjackie (approved):
Я бы объединил FormSelect и SelectOption, а TodoList с TodoForm. Особенно последний, маленький компонент, занимающий 7 строк, как по мне не достоен жизни в отдельном файле, с учетом его несложной логики, которую вряд-ли нужно будет менять или прям много в неё дописывать, но тут все на вкус и цвет наверное, оставлю вердикт за ментором) В остальном мне всё нравиться ;) Андрей 👍🏻

aliceblunt8 (changes requested):
Lift state up to the App component, which will render Form and Todolist
Also, I think it's unnecessary to move FormSelect to another component, cause it's a form element, let's stick them together
Also fix this - after an empty title error when user starts typing, the error message is still visible
image

src/components/TodoForm/TodoForm.jsx
```diff
@@ -0,0 +1,151 @@
...
+  render() {
+    const { users } = this.props;
+    const { username, todoTitle, titleError, usernameError, todoList } = this.state;
+
+    return (
+      <>
+        <form
+          className="App__form form"
+          onSubmit={event => (
+            this.onSubmit(event)
```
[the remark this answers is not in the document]
andlap24: Alice, if i move state from TodoForm up to App, then there will be 10 arguments to transfer between components( I should move that for sure ?

src/components/TodoForm/TodoForm.jsx
```diff
@@ -0,0 +1,160 @@
...
+      todoTitle,
+      titleError,
+      usernameError,
+      todoList } = this.state;
+
+    return (
+      <>
+        <form
+          className="App__form form"
+          onSubmit={this.onSubmit}
```
aliceblunt8, line 100:
```suggestion
          onSubmit={this.handleSubmit}
```

src/App.js
```diff
@@ -3,17 +3,26 @@ import './App.css';
...
-      </p>
-    </div>
-  );
-}
+      </span>
+    </p>
+
+    <TodoForm
+      users={users}
+    />
```
aliceblunt8, line 24: Todolist should be here

```suggestion
    <TodoForm
      users={users}
    />
    <TodoList
      todosList={todos}
    />
```

aliceblunt8 (changes requested):

src/components/TodoForm/TodoForm.jsx
```diff
@@ -0,0 +1,148 @@
...
+import './TodoForm.css';
+
+import { SelectOption } from '../SelectOption';
+
+export class TodoForm extends React.PureComponent {
+  state = {
+    todoTitle: '',
+    username: '',
+    todoUserId: 0,
+    todoId: this.props.todoId,
```
aliceblunt8, line 12: it's a very bad practice to use props in state
https://medium.com/@justintulk/react-anti-patterns-props-in-initial-state-28687846cc2e

aliceblunt8: just use todos.length + 1 for id, when you create a new todo, no need to put todoId in state

src/components/TodoForm/TodoForm.jsx
```diff
@@ -0,0 +1,148 @@
...
+    titleError: false,
+    usernameError: false,
+  }
+
+  addTodoUserId = (id) => {
+    this.setState({ todoUserId: Number(id) });
+  }
```
aliceblunt8, line 19: abundant

src/components/TodoForm/TodoForm.jsx
```diff
@@ -0,0 +1,148 @@
...
+    const { todoTitle, username } = this.state;
+
+    if (todoTitle === '' && username === '') {
+      this.setState({
+        usernameError: true,
+        titleError: true,
+      });
+
+      return;
+    }
```
aliceblunt8, line 41: two separate checks are enough, no need to check for both in one if

```suggestion
```

src/components/TodoForm/TodoForm.jsx
```diff
@@ -0,0 +1,148 @@
...
+    this.addTodo();
+
+    this.setState({
+      todoTitle: '',
+      username: '',
+      titleError: false,
+    });
+  }
+
+  addTodo() {
```
aliceblunt8, line 68: let's move todo creating into the App component
addTodo will create new todo and add it to the state
and you will it in handleSubmit and pass everything you need

src/components/TodoForm/TodoForm.jsx
```diff
@@ -0,0 +1,148 @@
...
+          onSubmit={this.handleSubmit}
+        >
+          <select
+            className="form__select select"
+            name="username"
+            id="username"
+            value={username}
+            onChange={(event) => {
+              this.addTodoUserId(event.target.value);
+              this.handleChange(event.target);
```
aliceblunt8, line 101: modify handleChange func, so you can use it in both input and select

aliceblunt8 (changes requested):

src/components/TodoForm/TodoForm.jsx
andlap24: all right, understand )))

aliceblunt8: yes, the whole func

aliceblunt8 (approved):

PR #312, opened 2021-01-21
Add todo form - solution

VinnieJ-2k20:
- [DEMO LINK](https://VinnieJ-2k20.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,19 +1,164 @@
 import React from 'react';
+import { TodoList } from './components/TodoList';
 import './App.css';
 
 import users from './api/users';
+import todos from './api/todos';
 
-function App() {
-  return (
-    <div className="App">
-      <h1>Add todo form</h1>
+const todosWithUsers = todos.map(todo => ({
+  ...todo,
+  user: findUser(todo.userId),
+}));
 
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
-    </div>
-  );
+function findUser(userId) {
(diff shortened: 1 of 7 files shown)


src/App.js
```diff
@@ -1,19 +1,164 @@
...
+  ...todo,
+  user: findUser(todo.userId),
+}));
+
+function findUser(userId) {
+  return users.find(user => user.id === userId);
 }
 
-export default App;
+const titleCharLimit = 16;
```
KirillLutsenko, line 17: А я пропустил эту проверку в задании, вот они плюсы ревью чужих заданий)

KirillLutsenko (approved):
:+1:

src/App.js
```diff
@@ -1,19 +1,164 @@
...
+    event.preventDefault();
+
+    if (!this.validateInputs()) {
+      return;
+    }
+
+    this.setState(({ todosList, selectedUser, taskTitle, nextId }) => {
+      const newTask = {
+        userId: +selectedUser,
+        user: findUser(+selectedUser),
```
varseniuk, line 78: Let's not pass the whole 'user' object if we need only it's name

varseniuk (changes requested):
Code is good, but let's not pass the whole 'user' object to new todo, only the name

varseniuk (commented):
Made small update to my comment about 'name' and 'userId'

aliceblunt8 (approved):
gj

PR #313, opened 2021-01-21
added solution

MarchD:
[DEMO LINK](https://MarchD.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,19 +1,131 @@
 import React from 'react';
+import { TodoList } from './components/TodoList';
 import './App.css';
+import todosFromApi from './api/todos';
+import usersFromApi from './api/users';
 
-import users from './api/users';
+function getUserById(UserId) {
+  return usersFromApi.find(user => user.id === UserId);
+}
 
-function App() {
-  return (
-    <div className="App">
-      <h1>Add todo form</h1>
+const preparedTodos = todosFromApi.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
 
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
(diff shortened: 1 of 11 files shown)


olesiatk (approved):
Good job!

src/App.js
```diff
@@ -1,19 +1,131 @@
...
+      errorUser: false,
+      errorTitle: false,
+    });
+
+    if (userId === 0) {
+      return this.setState({ errorUser: true });
+    }
+
+    if (taskTitle.trim().length < 1) {
+      return this.setState({ errorTitle: true });
```
vsm2k17, line 47: how did you expect this should work? why do you return the result of setState? and return it to where?

src/App.js
```diff
@@ -1,19 +1,131 @@
...
+
+const preparedTodos = todosFromApi.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
+
+class App extends React.Component {
+  state = {
+    todos: preparedTodos,
+    users: usersFromApi,
```
vsm2k17, line 19: I think you can avoid storing users in a state

vsm2k17 (changes requested):

src/App.js
```diff
@@ -1,19 +1,135 @@
...
+
+    if (taskTitle.trim().length < 1) {
+      this.setState({ errorTitle: true });
+
+      return;
+    }
+
+    this.setState((state) => {
+      const newTask = {
+        id: Math.random(),
```
vsm2k17, line 55: better to use previous id + 1, think about the situation when math.random generate you the same id as you already have

vsm2k17 (approved):
good job

PR #318, opened 2021-01-21
Dev

Ruslanwp:
[DEMO LINK](https://Ruslanwp.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,19 +1,119 @@
 import React from 'react';
 import './App.css';
+import { TodosList } from './components/TodosList/TodosList';
+import { UsersList } from './components/UsersList/UsersList';
 
-import users from './api/users';
+import usersFromApi from './api/users';
+import todosFromApi from './api/todos';
 
-function App() {
-  return (
-    <div className="App">
-      <h1>Add todo form</h1>
+const preparedTodos = todosFromApi.map(todo => ({
+  ...todo,
+  user: usersFromApi.find(user => user.userId === todo.id),
+}));
 
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
-    </div>
-  );
+class App extends React.Component {
+  state = {
+    todos: preparedTodos,
+    selectedUserId: 0,
+    title: '',
+    errorTitle: '',
+    errorSelectUser: '',
+  }
+
+  handleSelectUser = (event) => {
+    const {
+      name, value,
+    } = event.target;
+
+    this.setState({
+      [name]: +value,
+      errorSelectUser: '',
+    });
+  }
+
+  handleChangeTitle = (event) => {
+    const {
+      name, value,
+    } = event.target;
+
+    this.setState({
+      [name]: value,
+      errorTitle: '',
+    });
+  }
+
+  render() {
+    const {
+      todos,
+      title,
+      selectedUserId,
+      errorSelectUser,
+      errorTitle,
+    } = this.state;
+
+    return (
+      <div className="App">
+        <h1>Add todo form</h1>
+
+        <p>
+          <span>Users: </span>
+          {todos.length}
+        </p>
+
+        <form onSubmit={((event) => {
+          event.preventDefault();
+
+          if (!selectedUserId) {
+            this.setState({
+              errorSelectUser: 'Please choose a user',
+            });
+          }
+
+          if (!title) {
+            this.setState({
+              errorTitle: 'Please choose a user',
+            });
+          }
+
+          if (selectedUserId && title) {
+            this.setState((prevState) => {
+              const newGood = {
+                userId: prevState.selectedUserId,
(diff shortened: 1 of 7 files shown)


Serhii-Alieksieievych (changes requested):
https://prnt.sc/xdzm9k

We have 3 tasks, 10 users in the array, and only one user is mentioned as the contractor...
Use user name instead of user id...

Serhii-Alieksieievych (approved):

src/App.js
```diff
@@ -1,19 +1,120 @@
...
+          }
+
+          if (selectedUserId && title) {
+            this.setState((prevState) => {
+              const newGood = {
+                userId: prevState.selectedUserId,
+                id: todos.length + 1,
+                title: prevState.title,
+                completed: false,
+                user: usersFromApi[prevState.selectedUserId - 1],
```
varseniuk, line 85: We don't need the whole 'user' object. Just its name

varseniuk (changes requested):
empty

We can add an empty todo (using spaces). Let's fix this

src/App.js
varseniuk: Small update:
Actually, we don't need to pass the 'name' only 'userId'.
```
 {
    userId: 1,
    id: 1,
    title: 'delectus aut autem',
    completed: false,
  },
```
So we should pass 'userId' to our new TODO-object, as task description says:

> each TODO should have a title and userId selected from a list of given users

Also, you may pass the 'Name' of the user to the TodoList to make it looks prettier, but not to add it to the new 'todo' object and it's not essential for this task

varseniuk (commented):
Added comment about 'name' and 'userId'

src/App.js
```diff
@@ -1,19 +1,129 @@
...
+  handleChangeTitle = (event) => {
+    const {
+      name, value,
+    } = event.target;
+
+    this.setState({
+      [name]: value,
+      errorTitle: '',
+    });
+  }
```
aliceblunt8, line 43: let's combine it into ONE function

src/components/User/User.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { TypeUser } from '../../types';
+
+export const User = ({ id, name }) => (
+  <option
+    value={+id}
```
aliceblunt8, line 6:
```suggestion
    value={id}
```

src/components/UsersList/UsersList.js
```diff
@@ -0,0 +1,27 @@
...
+    name="selectedUserId"
+    value={selectedUserId}
+    onChange={handle}
+  >
+    <option value="">
+      Choose a user
+    </option>
+
+    {users.map(user => (
+      <User {...user} key={user.id} />
```
aliceblunt8, line 17: it's a bad idea to move select's option to the separate component, let's move it back here

src/components/UsersList/UsersList.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { User } from '../User/User';
+import { TypeUser } from '../../types';
+
+export const UsersList = ({ users, handle, selectedUserId }) => (
+  <select
+    name="selectedUserId"
+    value={selectedUserId}
+    onChange={handle}
```
aliceblunt8, line 10:
```suggestion
    onChange={handleChange}
```

src/App.js
```diff
@@ -1,19 +1,129 @@
...
+    return (
+      <div className="App">
+        <h1>Add todo form</h1>
+
+        <p>
+          <span>Todos: </span>
+          {todos.length}
+        </p>
+
+        <form onSubmit={((event) => {
```
aliceblunt8, line 63: move onSubmit body to the separate function handleSubmit

src/App.js
```diff
@@ -1,19 +1,129 @@
...
+
+          const trimedTitle = title.trim();
+
+          if (trimedTitle.length === 0) {
+            this.setState({
+              errorTitle: 'Please enter the title',
+            });
+
+            return;
+          }
```
aliceblunt8, line 86: DRY

```suggestion
          if (!title.trim()) {
            this.setState({
              errorTitle: 'Please enter the title',
            });
          }
```

src/App.js
```diff
@@ -1,19 +1,129 @@
...
+
+          if (selectedUserId && title) {
+            this.setState((prevState) => {
+              const newGood = {
+                userId: prevState.selectedUserId,
+                id: todos.length + 1,
+                title: prevState.title,
+                user: usersFromApi[prevState.selectedUserId - 1],
```
aliceblunt8, line 94: create this variable outside the setState (now it looks messy)

aliceblunt8 (changes requested):

PR #319, opened 2021-01-21
Create Add Todo Form on React

KirillLutsenko:
[DEMO LINK](https://KirillLutsenko.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,0 +1,50 @@
+import React from 'react';
+import './App.css';
+
+import users from './api/users';
+import todosFromApi from './api/todos';
+import { TodoList } from './components/TodoList';
+import { TodoForm } from './components/TodoForm/TodoForm';
+
+const preparedTodos = todosFromApi.map(todo => ({
+  ...todo,
+  user: users.find(user => (todo.userId === user.id)),
+}));
+
+class App extends React.Component {
+  state = {
+    todos: preparedTodos,
+  }
+
+  addTodo = (title, user) => {
+    this.setState((state) => {
+      const newTodo = {
+        userId: user.id,
+        id: state.todos.length + 1,
+        title,
+        completed: false,
+        user,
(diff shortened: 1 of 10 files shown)


VinnieJ-2k20 (approved):
Good job and cool design! 
Just need to figure out the correct error display mode.

src/App.jsx
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  addTodo = (title, user) => {
+    this.setState((state) => {
+      const newTodo = {
+        userId: user.id,
+        id: state.todos.length + 1,
+        title,
+        completed: false,
+        user,
```
varseniuk, line 26: No need to pass the whole 'user' object. You need only its name

src/components/User/User.jsx
```diff
@@ -0,0 +1,13 @@
...
+import { UserType } from '../../types';
+import './User.css';
+
+export const User = ({ name }) => (
+  <div className="user__name">
+    <b>Name</b>
+    {': '}
+    {`${name}`}
+  </div>
+);
```
varseniuk, line 11: It's not very rational to create a separate component only just for displaying one string

varseniuk (changes requested):

src/App.jsx
varseniuk: Small update:
Actually, we even don't need to pass the 'name' (only 'userId').
```
 {
    userId: 1,
    id: 1,
    title: 'delectus aut autem',
    completed: false,
  },
```
So we should pass 'userId' to our new TODO-object, as task description says:

> each TODO should have a title and userId selected from a list of given users

Also, you may pass the 'Name' of the user to the TodoList to make it looks prettier, but not to add it to the new 'todo' object and it's not essential for this task

varseniuk (commented):
Updated comments

PR #320, opened 2021-01-21
Develop

vitalya2626:
[DEMO LINK](https://vitalya2626.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,19 +1,55 @@
+
 import React from 'react';
 import './App.css';
 
 import users from './api/users';
+import todos from './api/todos';
+import { NewForm } from './components/NewForm';
+import { ToDoList } from './components/TodoList';
 
-function App() {
-  return (
-    <div className="App">
-      <h1>Add todo form</h1>
+const preparedTodos = todos.map(todo => ({
+  ...todo,
+  user: users.find(user => user.id === todo.userId),
+}));
 
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
-    </div>
-  );
+class App extends React.Component {
+  state = {
+    todosList: preparedTodos,
+  }
+
+  addTodo = (title, newUser) => {
+    this.setState((state) => {
+      const newTodo = {
+        id: state.todosList.length + 1,
+        completed: false,
+        title,
+        userId: newUser.id,
+        user: newUser,
+      };
+
+      return {
+        todosList: [
+          ...state.todosList,
+          newTodo,
+        ],
+      };
+    });
+  }
+
+  render() {
+    return (
+      <div className="App">
+        <h1 className="app-title">Add todo form</h1>
+        <NewForm
+          onAdd={this.addTodo}
+          users={users}
+        />
+        <ToDoList
+          todos={this.state.todosList}
+        />
+      </div>
+    );
+  }
 }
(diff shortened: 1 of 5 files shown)


NikitaTudakov (approved):
Unbelievable!

src/components/TodoList.js
```diff
@@ -0,0 +1,45 @@
...
+
+      <tbody className="tbody">
+        {todos.map(todo => (
+          <tr key={todo.id}>
+            <td>
+              {todo.title}
+            </td>
+
+            <td>
+              {todo.user.name}
```
varseniuk, line 24: We don't need to pass the whole 'user' object here, just it's name

varseniuk (changes requested):
Let's pass only user's name, not the whole object

varseniuk (changes requested):
Now you have an empty field 'user name' in your [table](https://prnt.sc/xhrdz7). You may remove it at all and display only 'userId' in your table.

```
 {
    userId: 1,
    id: 1,
    title: 'delectus aut autem',
    completed: false,
  },
```
So we should pass only 'userId' to our new TODO-object, as task description says:

each TODO should have a title and userId selected from a list of given users

Also, if you want, you may pass somehow the 'Name' of the user to the TodoList to make it looks prettier, but not to add it to the new 'todo' object and it's not essential for this task so don't waste time on this

src/App.js
```diff
@@ -1,19 +1,55 @@
...
 import './App.css';
 
 import users from './api/users';
+import todos from './api/todos';
+import { NewForm } from './components/NewForm';
+import { ToDoList } from './components/TodoList';
+
+const preparedTodos = todos.map(todo => ({
+  ...todo,
+  user: users.find(user => user.id === todo.userId),
```
aliceblunt8, line 12: store just userId in todos, no need to store the whole user object

src/App.js
```diff
@@ -1,19 +1,55 @@
...
+
+  addTodo = (title, newUser) => {
+    this.setState((state) => {
+      const newTodo = {
+        id: state.todosList.length + 1,
+        completed: false,
+        title,
+        userId: newUser.id,
+        user: newUser,
+      };
```
aliceblunt8, line 28:
```suggestion
    this.setState((prevState) => {
      const newTodo = {
        id: prevState.todosList.length + 1,
        completed: false,
        title,
        userId: newUser.id,
        user: newUser,
      };
```

src/App.js
```diff
@@ -1,19 +1,55 @@
...
+
+      return {
+        todosList: [
+          ...state.todosList,
```
aliceblunt8, line 32:
```suggestion
          ...prevState.todosList,
```

src/App.js
```diff
@@ -1,19 +1,55 @@
...
-      </p>
-    </div>
-  );
+  render() {
+    return (
+      <div className="App">
+        <h1 className="app-title">Add todo form</h1>
+        <NewForm
+          onAdd={this.addTodo}
+          users={users}
```
aliceblunt8, line 45: you can import users inside the component, no need to pass them as a prop

aliceblunt8 (approved):
@vitalya2626 good job, approved, but please check my comments ;)

PR #321, opened 2021-01-21
Completed

Serhii-Alieksieievych:
[DEMO LINK](https://Serhii-Alieksieievych.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,19 +1,146 @@
 import React from 'react';
 import './App.css';
-
+import todos from './api/todos';
 import users from './api/users';
 
-function App() {
-  return (
-    <div className="App">
-      <h1>Add todo form</h1>
+class App extends React.Component {
+  state = {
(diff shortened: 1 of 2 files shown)


tamara-ostapets (approved):
cool 👀

src/App.js
```diff
@@ -1,19 +1,146 @@
...
+  };
+
+  render() {
+    const { usersForRender, userId, title } = this.state;
+    const todosForRender = this.state.todos
+      .map(todo => ({
+        ...todo,
+        user: usersForRender
+          .find(user => user.id === todo.userId),
+      }));
```
vsm2k17, line 23: 1) I don't think that you need to add the user to your todo, you need only the name
2) if you do need to create something, create it outside your component, so it happens only once

vsm2k17 (approved):
good job

PR #323, opened 2021-01-21
Solution

AlexysUa:
[DEMO LINK](https://alexysua.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoForm/TodoForm.jsx
+++ b/src/components/TodoForm/TodoForm.jsx
@@ -1,0 +1,133 @@
+import React from 'react';
+import { Form, Button, Alert } from 'react-bootstrap';
+import './TodoForm.css';
+import PropTypes from 'prop-types';
+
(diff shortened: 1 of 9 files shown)


src/components/TodoForm/TodoForm.jsx
```diff
@@ -0,0 +1,133 @@
...
+              Add Todo
+            </Button>
+          </div>
+        </Form>
+      </>
+    );
+  }
+}
+
+TodoFrom.propTypes = {
```
Ruslanwp, line 121: you shouldn't write propTypes for properties that you don't use, it is not necessary

Ruslanwp (approved):

src/components/TodoForm/TodoForm.jsx
Ruslanwp: great Job!

vsm2k17 (approved):
good job

PR #325, opened 2021-01-21
solution

5Mountains:
- [x] [DEMO LINK](https://5Mountains.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,19 +1,129 @@
 import React from 'react';
 import './App.css';
 
+import todos from './api/todos';
 import users from './api/users';
+import { TodoList } from './components/TodoList';
 
-function App() {
-  return (
-    <div className="App">
-      <h1>Add todo form</h1>
+function getuserByID(userId) {
+  return users.find(user => user.id === userId);
+}
 
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
-    </div>
-  );
+const preparedTodos = todos.map(todo => ({
+  ...todo,
+  user: getuserByID(todo.userId),
+}));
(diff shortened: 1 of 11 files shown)


sasha39612 (approved):

sasha39612:
Great job!

src/App.js
```diff
@@ -1,19 +1,131 @@
...
-      <h1>Add todo form</h1>
-
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
-    </div>
-  );
+function getuserByID(userId) {
+  return users.find(user => user.id === userId);
```
varseniuk, line 9: You don't need the whole 'user' object, just his name

src/components/User/User.jsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const User = ({ name }) => (
+  <span>{name}</span>
+);
```
varseniuk, line 6: It's very irrational to create element with only one 'span' that only displays the name of the user. You should put this span in your 'Todo' component

varseniuk (changes requested):
empty

We can add empty todo (using spaces). Fix this

src/App.js
5Mountains: fixed

src/components/User/User.jsx
5Mountains: made it

5Mountains:
Nice catch, by the way)

varseniuk (approved):
👍

PR #330, opened 2021-01-22
solution

HaidaiAndrii:
[DEMO LINK](https://HaidaiAndrii.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,19 +1,115 @@
 import React from 'react';
 import './App.css';
-
+import { TodoList } from './Components/TodoList/TodoList';
 import users from './api/users';
+import todos from './api/todos';
 
-function App() {
-  return (
-    <div className="App">
-      <h1>Add todo form</h1>
+const preparedTodos = [...todos].map(todo => ({
+  ...todo,
+  user: users.find(user => user.id === todo.userId).name,
+}));
 
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
-    </div>
-  );
+class App extends React.Component {
+  state = {
+    todos: preparedTodos,
+    title: '',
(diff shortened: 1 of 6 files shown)


Serhii-Alieksieievych (approved):

src/App.js
```diff
@@ -1,19 +1,115 @@
 import React from 'react';
 import './App.css';
-
+import { TodoList } from './Components/TodoList/TodoList';
 import users from './api/users';
+import todos from './api/todos';
+
+const preparedTodos = [...todos].map(todo => ({
+  ...todo,
+  user: users.find(user => user.id === todo.userId).name,
```
aionova, line 9:
```suggestion
  userName: users.find(user => user.id === todo.userId).name,
```

src/App.js
```diff
@@ -1,19 +1,115 @@
...
+  addTask = () => {
+    if (this.state.title.trim().length > 0 && this.state.userId > 0) {
+      this.setState(state => ({
+        todos: [...state.todos,
+          {
+            userId: state.userId,
+            id: state.id,
+            title: state.title,
+            completed: false,
+            user: users.find(user => user.id === state.userId),
```
aionova, line 41: no need to pass the whole user object here. You can pass only userName here

src/Components/User/User.js
```diff
@@ -0,0 +1,16 @@
...
+
+export const User = ({ name }) => (
+  <>
+    name:
+    {' '}
+    <span>
+      {name}
+    </span>
+  </>
+);
```
aionova, line 12: no need to create a separate component for 1 row of code

aionova (changes requested):
Good start!

PR #337, opened 2021-01-23
todo List

kerpichonka:
[DEMO LINK](https://kerpichonka.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,19 +1,128 @@
 import React from 'react';
+import { TodoList } from './components/TodoList/TodoList';
 import './App.css';
 
 import users from './api/users';
+import todos from './api/todos';
 
-function App() {
-  return (
-    <div className="App">
-      <h1>Add todo form</h1>
+const todosFromServer = todos.map(todo => ({
+  ...todo,
+  name: users.find(user => user.id === todo.userId).name,
+}));
 
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
-    </div>
-  );
+class App extends React.Component {
+  state = {
+    people: users,
+    todoList: todosFromServer,
+    task: '',
+    userName: '',
+    id: todosFromServer[todosFromServer.length - 1].id + 1,
+    errorUser: false,
+    errorTask: false,
+  };
+
+  handlerChange = (event) => {
+    const { value, name } = event.target;
+
+    this.setState({
+      [name]: value.replace(/[^ a-z ]/gi, ''),
+      errorUser: false,
+      errorTask: false,
+    });
+  }
+
+  hendlerSubmit = (event) => {
+    event.preventDefault();
+
+    if (!this.state.task) {
+      this.setState({ errorTask: true });
+
+      return;
+    }
+
+    if (!this.state.userName || this.state.userName === 'Choose a user') {
+      this.setState({ errorUser: true });
+
+      return;
+    }
+
+    this.setState((state) => {
+      const newTodo = {
+        name: state.userName,
+        userId: state.people.find(user => user.name === state.userName).id,
+        title: state.task,
+        completed: false,
+        id: state.id,
+      };
+
+      return {
+        id: state.id + 1,
+        todoList: [...state.todoList, newTodo],
+        task: '',
+        userName: '',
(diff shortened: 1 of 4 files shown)


dtsehelnyk (approved):

src/App.js
```diff
@@ -1,19 +1,128 @@
...
-
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
-    </div>
-  );
+const todosFromServer = todos.map(todo => ({
+  ...todo,
+  name: users.find(user => user.id === todo.userId).name,
```
aliceblunt8, line 10:
```suggestion
  userName: users.find(user => user.id === todo.userId).name,
```

src/App.js
```diff
@@ -1,19 +1,128 @@
...
+}));
+
+class App extends React.Component {
+  state = {
+    people: users,
```
aliceblunt8, line 15: no need to save users in state, just import it where you need it

src/App.js
```diff
@@ -1,19 +1,128 @@
...
+    todoList: todosFromServer,
+    task: '',
+    userName: '',
+    id: todosFromServer[todosFromServer.length - 1].id + 1,
```
aliceblunt8, line 19: redundant, let's do calculate it when setting state or just use Date.now()

src/App.js
```diff
@@ -1,19 +1,128 @@
...
+    errorUser: false,
+    errorTask: false,
+  };
+
+  handlerChange = (event) => {
```
aliceblunt8, line 24:
```suggestion
  handleChange = (event) => {
```

aliceblunt8: it's a function, use verbs

src/App.js
```diff
@@ -1,19 +1,128 @@
...
+    const { value, name } = event.target;
+
+    this.setState({
+      [name]: value.replace(/[^ a-z ]/gi, ''),
+      errorUser: false,
+      errorTask: false,
+    });
+  }
+
+  hendlerSubmit = (event) => {
```
aliceblunt8, line 34:
```suggestion
  handleSubmit = (event) => {
```

src/App.js
```diff
@@ -1,19 +1,128 @@
...
+    event.preventDefault();
+
+    if (!this.state.task) {
+      this.setState({ errorTask: true });
+
+      return;
+    }
+
+    if (!this.state.userName || this.state.userName === 'Choose a user') {
+      this.setState({ errorUser: true });
```
aliceblunt8, line 44: let's destructure vars from this.state

src/App.js
```diff
@@ -1,19 +1,128 @@
...
+                <p className="form__error">Please enter the title</p>)}
+              <label htmlFor="input-task">
+                <input
+                  type="text"
+                  name="task"
+                  id="input-task"
+                  value={task}
+                  onChange={this.handlerChange}
+                  placeholder="Write task"
+                  className="form__input"
```
aliceblunt8, line 94: try to add some value to this input, it doesn't work

aliceblunt8: I think the problem might be in your handleChange function, maybe the replace method is redundant in this case

aliceblunt8 (changes requested):

src/App.js
kerpichonka: It is task condition https://prnt.sc/xr69p1

src/App.js
kerpichonka: interesting, in my case it all works

src/App.js
```diff
@@ -1,19 +1,128 @@
...
+
+    if (!userName || userName === 'Choose a user') {
+      this.setState({ errorUser: true });
+
+      return;
+    }
+
+    this.setState((state) => {
+      const newTodo = {
+        name: state.userName,
```
aliceblunt8, line 52:
```suggestion
        userName: state.userName,
```

user's name is undefined, cause you first save the value to 'name' prop, but read in Todo as 'userName'

image

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #427, opened 2021-05-26
Solution

igor-stasiv:
[DEMO LINK](https://igor-stasiv.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoList/TodoList.jsx
+++ b/src/components/TodoList/TodoList.jsx
@@ -1,0 +1,144 @@
+import React from 'react';
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+
+import './TodoList.scss';
+import { Todo } from './Todo';
+
+export class TodoList extends React.Component {
+  state = {
+    todos: [...this.props.todos],
+    title: '',
+    userName: '',
+    isTitleAdd: true,
+    isUserSelected: true,
+  };
+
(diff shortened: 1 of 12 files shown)


src/components/TodoList/TodoList.jsx
```diff
@@ -0,0 +1,144 @@
...
+      }
+
+      if (!title || title.length > 35 || title.length < 5) {
+        return { isTitleAdd: false };
+      }
+
+      if (!userName) {
+        return { isUserSelected: false };
+      }
+
```
Ihor-Karpyn, line 54: Maybe move these logic to handlers

Ihor-Karpyn (approved):

src/components/TodoList/TodoList.jsx
varseniuk: Well, these checks should be triggered on form submit, so let them be here

src/components/TodoList/TodoList.jsx
```diff
@@ -0,0 +1,144 @@
...
+      return ({
+        todos: [...todos, {
+          id: todos[todos.length - 1].id + 1,
```
varseniuk, line 57:
```suggestion
          id: todos.length + 1,
```

varseniuk (approved):
Really good job! 🔥

PR #441, opened 2021-05-27
Solution

SergeyZub:
[DEMO LINK](https://SergeyZub.github.io/react_add-todo-form/)


varseniuk (changes requested):
All is good, but we can add an empty todo (using spaces). Lrt's fix this

empty

varseniuk (approved):
Good!

PR #447, opened 2021-05-27
Solution

Taras-Konovalov:
[DEMO LINK](https://Taras-Konovalov.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,19 +1,114 @@
 import React from 'react';
 import './App.css';
+import UserList from './Components/UserList/UserList';
 
 import users from './api/users';
+import todos from './api/todos';
 
-function App() {
-  return (
-    <div className="App">
-      <h1>Add todo form</h1>
+const list = todos.map(todo => ({
+  ...todo,
+  user: users.find(user => user.id === todo.userId).name,
+}));
 
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
-    </div>
-  );
+class App extends React.Component {
+  state = {
+    todos: [...list],
+    title: '',
+    name: '',
+    todoError: false,
+    userError: false,
+  }
+
+  handleChangeTitle = (event) => {
+    this.setState({
+      title: event.target.value,
+    });
+  }
+
+  handleChangeName = (event) => {
+    this.setState({
+      name: event.target.value,
+    });
+  }
+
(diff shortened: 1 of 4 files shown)


den-bulaev (changes requested):
before creating a todo check if a user was selected, if not show an error message next to the select (**Please choose a user**). 'UserError' it seems like mystery for users.

hide a message immediately after any change of the field with an error

src/App.js
```diff
@@ -1,19 +1,126 @@
...
+  handleFormSubmit = (event) => {
+    if (this.state.title === '') {
+      this.setState({ todoError: true });
+    }
+
+    if (this.state.name === '') {
+      this.setState({ userError: true });
+    }
+
+    if (this.state.name !== '' && this.state.title !== '') {
```
VinnieJ-2k20, line 55:
```suggestion
    if (this.state.name && this.state.title) {
```

src/App.js
```diff
@@ -1,19 +1,126 @@
...
+
+      this.setState({
+        name: '',
+        title: '',
+        todoError: false,
+        userError: false,
+      });
+    }
+
+    event.preventDefault();
```
VinnieJ-2k20, line 77: This should be at the very start of the method.
If you'll have any async actions at the start(e.g. server requests), there may be some unexpected consequences.

src/Components/UserList/UserList.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const UserList = ({ userList }) => (
```
VinnieJ-2k20, line 4: Well, it's not `UserList`, it's rather `TodoList`.
Rename the component.

src/api/todos.js
```diff
@@ -8,7 +8,7 @@ const todos = [
     completed: false,
   },
   {
-    userId: 1,
+    userId: 2,
```
VinnieJ-2k20, line 11: This file simulates data received from server. You should not change it in any way.

VinnieJ-2k20 (changes requested):
Nice solution, just need some changes.

Upon adding new todo user `select` should be cleared.
Also added some more comments below.

VinnieJ-2k20 (approved):

PR #452, opened 2021-05-27
solution

Nazar-Kudatsky:
[DEMO LINK](https://Nazar-Kudatsky.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,19 +1,134 @@
 import React from 'react';
 import './App.css';
-
+import TodoList from './components/TodoList/TodoList';
+import todos from './api/todos';
 import users from './api/users';
 
-function App() {
+const preparedTodos = todos.map(todo => ({
+  ...todo,
+  user: users.find(user => user.id === todo.userId),
+}));
+
+class App extends React.Component {
+state = {
+  newTodo: '',
+  todos: preparedTodos,
+  user: '',
+  userError: false,
+  titleError: false,
+}
+
+handleChange = (event) => {
+  this.setState({
+    newTodo: event.target.value,
+    titleError: false,
+  });
+}
+
+handleSubmit = (event) => {
+  event.preventDefault();
+
+  if (this.state.newTodo.trim().length === 0 && this.state.user.length === 0) {
+    this.setState({
+      titleError: true,
+      userError: true,
+    });
+
+    return;
+  }
+
+  if (this.state.user.length === 0) {
+    this.setState({ userError: true });
+
+    return;
+  }
+
+  if (this.state.newTodo.trim().length === 0) {
+    this.setState({ titleError: true });
+
+    return;
+  }
+
+  this.setState(state => ({ todos: [
+    {
+      completed: false,
+      title: state.newTodo,
+      user: users.find(person => person.name === state.user),
+      userId: users.find(person => person.name === state.user).id,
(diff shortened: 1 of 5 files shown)


scarlettjohanssons (approved):

scarlettjohanssons:
Good job

src/App.js
```diff
@@ -1,19 +1,134 @@
...
+
+    return;
+  }
+
+  this.setState(state => ({ todos: [
+    {
+      completed: false,
+      title: state.newTodo,
+      user: users.find(person => person.name === state.user),
+      userId: users.find(person => person.name === state.user).id,
```
mykola-yarchuk, line 57: 1) why you need user AND userId, if you have user , you also have an id

mykola-yarchuk: 2. you can find user in advance and use it here, to prevent double invoke of find method

src/components/Todo/Todo.js
```diff
@@ -0,0 +1,29 @@
...
+
+  return (
+    <>
+      <h2>
+        {title}
+      </h2>
+      <p>
+        {`Status: ${status}`}
+      </p>
+      <User {...user} />
```
mykola-yarchuk, line 18: if you need only name then pass only name

mykola-yarchuk (changes requested):

src/App.js
```diff
@@ -1,19 +1,133 @@
...
+class App extends React.Component {
+state = {
+  newTodo: '',
+  todos: preparedTodos,
+  user: '',
+  userError: false,
+  titleError: false,
+}
+
+handleChange = (event) => {
```
VinnieJ-2k20, line 21:
```suggestion
handleTitleChange = (event) => {
```

src/App.js
```diff
@@ -1,19 +1,133 @@
...
+    this.setState({ userError: true });
+
+    return;
+  }
+
+  if (this.state.newTodo.trim().length === 0) {
+    this.setState({ titleError: true });
+
+    return;
+  }
```
VinnieJ-2k20, line 50: You can simplify this check.
First, check if there are errors and then change state accordingly.
Also, it is better to move form validation to separate method.

src/App.js
```diff
@@ -1,19 +1,133 @@
...
+
+  this.setState(state => ({ todos: [
+    {
+      completed: false,
+      title: state.newTodo,
+      user: users.find(person => person.name === state.user),
+      id: state.todos.length + 1,
```
VinnieJ-2k20, line 57: id ≠ length. More reliable way is to find max id and then add 1

src/components/Todo/Todo.js
VinnieJ-2k20: +1. Change it.

VinnieJ-2k20 (changes requested):
Almost done.
Left some comments on code optimisation.
Also need to fix one thing - user in select doesn't clear on form submit.

src/App.js
Nazar-Kudatsky: > You can simplify this check.
> First, check if there are errors and then change state accordingly.
> Also, it is better to move form validation to separate method.

Лінтер шаленіє, якщо не вийде з ним, то залишу як є

VinnieJ-2k20 (approved):

PR #453, opened 2021-05-27
Solution without styles

ViPiven:
[DEMO LINK](https://ViPiven.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -2,18 +2,115 @@ import React from 'react';
 import './App.css';
 
 import users from './api/users';
+import todos from './api/todos';
+import TodoList from './components/TodoList/TodoList';
 
-function App() {
-  return (
-    <div className="App">
-      <h1>Add todo form</h1>
+const findUser = userId => users.find(user => user.id === userId);
 
-      <p>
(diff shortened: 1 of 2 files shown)


ched88xtc (commented):
- hide a message immediately after any change of the field with an error

ched88xtc (approved):
LGTM!

src/App.js
```diff
@@ -2,18 +2,117 @@ import React from 'react';
...
+
+    this.setState({
+      newTitle: '',
+      newUserId: 0,
+    });
+  }
+
+  render() {
+    //  eslint-disable-next-line
+    const { todos, newTitle, newUserId, titleError, userError } = this.state;
```
varseniuk, line 65:
```suggestion
    const {
      todos,
      newTitle,
      newUserId,
      titleError,
      userError
    } = this.state;
```

varseniuk (changes requested):
All is good, but we can add an empty todo (using spaces). Let's fix this

Снимок экрана 2021-05-28 в 11 06 12

varseniuk (approved):
👍

PR #454, opened 2021-05-27
todo form implemented

illisogori:
[DEMO LINK](https://illisogori.github.io/react_add-todo-form/)


vadym-liubych (approved):
Nice solution!

src/components/Todo/Todo.jsx
```diff
@@ -0,0 +1,34 @@
...
+export const Todo = ({ title, completed, user }) => (
+  <div className="Todo">
+    <h2 className="Todo__title">
+      {title}
+    </h2>
+    <p className="Todo__user">{user.name}</p>
+    {completed
+      ? (
+        <span
+          className="Todo__status Todo__status--done"
```
varseniuk, line 14: You could create only one `span` instead of two with text `{completed ? 'Completed' : 'Not completed'}` and use `classnames` library

src/components/TodoForm/TodoForm.jsx
```diff
@@ -0,0 +1,88 @@
...
+        todo: '',
+        userId: 0,
+      },
+    );
+  }
+
+  todoHandler = (event) => {
+    this.setState({
+      todoError: !event.target.value,
+      todo: event.target.value,
```
varseniuk, line 43: We can add an empty todo (using spaces)

empty

varseniuk (changes requested):
All is good, but let's make impossible to add empty todos

varseniuk (approved):

PR #458, opened 2021-05-27
solution

vadym-liubych:
[DEMO LINK](https://vadym-liubych.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,19 +1,49 @@
 import React from 'react';
+import { TodoList } from './components/TodoList';
+import { TodoForm } from './components/TodoForm';
 import './App.css';
 
 import users from './api/users';
+import todos from './api/todos';
 
-function App() {
-  return (
-    <div className="App">
-      <h1>Add todo form</h1>
+function getUserById(userId) {
+  return users.find(user => user.id === userId);
+}
 
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
-    </div>
-  );
+const preparedTodos = todos.map(todo => ({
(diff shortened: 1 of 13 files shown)


src/App.js
```diff
@@ -1,19 +1,49 @@
...
-        <span>Users: </span>
-        {users.length}
-      </p>
-    </div>
-  );
+function getUserById(userId) {
+  return users.find(user => user.id === userId);
+}
+
+const preparedTodos = todos.map(todo => ({
```
illisogori, line 13:
```suggestion
const preparedTodos = todos.map(({userId})=> ({
```

illisogori (approved):
Good job.

src/App.js
varseniuk: Good point

src/components/TodoList/TodoList.js
```diff
@@ -0,0 +1,43 @@
...
+      {todos.map(todo => (
+        <li key={todo.id}>
+          <p>
+            <User user={todo.user} />
+          </p>
+          <p>
+            {`${todo.title} - `}
+            <span>
+              {todo.completed ? (
+                <span className="todo__status--done">Done</span>
```
varseniuk, line 18: You could create only one `span` instead of two with text {todo.completed ? 'Done' : 'Doing'} and use `classnames` library for switching classes

varseniuk (changes requested):
All is good, but we can add an empty todo (using spaces). Let's fix this

varseniuk (approved):

PR #474, opened 2021-05-28
solution

Namart:
[DEMO LINK](https://Namart.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -2,16 +2,18 @@ import React from 'react';
 import './App.css';
 
 import users from './api/users';
+import todos from './api/todos';
+import { TodoList } from './components/TodoList';
+
+const preparedTodos = todos.map(todo => ({
+  ...todo,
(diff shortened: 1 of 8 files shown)


src/App.js
```diff
@@ -2,16 +2,18 @@ import React from 'react';
...
 function App() {
   return (
     <div className="App">
-      <h1>Add todo form</h1>
-
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
+      <TodoList todos={preparedTodos} users={users} />
```
aliceblunt8, line 16: no need to pass users, just import it in file where you need it

aliceblunt8: prepare todos before passing it to TodoList, each todo should have a user property with object 
also put todos in App's state and remove them from TodoList state

aliceblunt8 (approved):

PR #477, opened 2021-05-28
React add todo form

rauzok:
[DEMO LINK](https://rauzok.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -28,10 +28,11 @@ class App extends React.Component {
       taskError: true,
     }));
 
-  eventUserOnChange = event => this.setState({
-    user: event.target.value,
-    userError: true,
-  });
+  eventUserOnChange = event => (
+    this.setState({
+      user: event.target.value,
+      userError: true,
+    }));
 
   addNewTask = (event) => {
     event.preventDefault();
@@ -39,14 +40,10 @@ class App extends React.Component {
 
     if (!title) {
       this.setState({ taskError: false });
-    } else {
-      this.setState({ taskError: true });
     }
 
     if (!user) {
       this.setState({ userError: false });
-    } else {
-      this.setState({ userError: true });
     }
 
     if (title && user) {


src/App.jsx
```diff
@@ -0,0 +1,122 @@
...
+      this.setState({ taskError: false });
+    } else {
+      this.setState({ taskError: true });
+    }
+
+    if (!user) {
+      this.setState({ userError: false });
+    } else {
+      this.setState({ userError: true });
+    }
```
YuriiLiso, line 50:
```suggestion
    if (!title) {
      this.setState({ taskError: false });
    }
    
    if (!user) {
      this.setState({ userError: false });
    } 
```
If the condition not passed, the default value is already ```true```.

YuriiLiso (commented):
Good Looks! Need some changes

YuriiLiso (approved):

src/App.jsx
```diff
@@ -0,0 +1,119 @@
...
+
+import users from './api/users';
+import todos from './api/todos';
+
+import { TodosList } from './components/todosList';
+
+const TodosWithUsers = todos.map(todo => ({
+  ...todo,
+  user: users.find(user => user.id === todo.userId).name,
+  isVisible: true,
```
VinnieJ-2k20, line 12: Why do we need it here? It is not used anywhere.

src/App.jsx
```diff
@@ -0,0 +1,119 @@
...
+}));
+
+class App extends React.Component {
+  state = {
+    renderList: [...TodosWithUsers],
+    title: '',
+    user: '',
+    completed: false,
+    taskError: true,
+    userError: true,
```
VinnieJ-2k20, line 22: I would change logic
`error === true` should mean that there is indeed an error present.
```suggestion
    taskError: false,
    userError: false,
```

VinnieJ-2k20 (approved):
Good job 👍 

Added some comments on slight optimisation.

PR #478, opened 2021-06-01
solution

stas-dukach:
[DEMO LINK](https://stas-dukach.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,19 +1,109 @@
 import React from 'react';
+import { TodoList } from './components/TodoList';
 import './App.css';
 
 import users from './api/users';
 
-function App() {
-  return (
-    <div className="App">
-      <h1>Add todo form</h1>
+class App extends React.Component {
+  state = {
+    userName: '',
+    todo: '',
+    todos: [],
+    userSelected: false,
+    todoSelected: false,
+  }
 
-      <p>
-        <span>Users: </span>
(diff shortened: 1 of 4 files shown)


src/App.js
```diff
@@ -1,19 +1,109 @@
 import React from 'react';
+import { TodoList } from './components/TodoList';
 import './App.css';
 
 import users from './api/users';
```
VinnieJ-2k20, line 5: You also need to import existing todos and render them in `TodoList`.

src/App.js
```diff
@@ -1,19 +1,109 @@
...
+      todoSelected: !state.todo,
+    }));
+
+    if (!this.state.userName) {
+      return;
+    }
+
+    if (!this.state.todo) {
+      return;
+    }
```
VinnieJ-2k20, line 30: Let's simplify your code.

```suggestion
    if (!this.state.userName || !this.state.todo) {
      return;
    }
```

VinnieJ-2k20 (changes requested):
Almost there.
Need to fix couple of things.
I would also made impossible to add todo with just spaces in title.

VinnieJ-2k20 (changes requested):
You fixed one item. Good.
Now just fix other two comments:
image

VinnieJ-2k20 (approved):

PR #553, opened 2021-09-13
solutiion add-todo-form

AndreySerhovetc:
[DEMO LINK](https://AndreySerhovetc.github.io/react_add-todo-form/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # React add TODO form
-- Replace `<your_account>` with your Github username in the [DEMO LINK](https://<your_account>.github.io/react_add-todo-form/)
+- Replace `<your_account>` with your Github username in the [DEMO LINK](https://AndreySerhovetc.github.io/react_add-todo-form/)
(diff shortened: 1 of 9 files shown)


src/Components/Form/AddNewTodo.tsx
```diff
@@ -0,0 +1,132 @@
...
+      this.setState(currentState => ({
+        isSelectUser: !currentState.isSelectUser,
+      }));
+
+      return;
+    }
+
+    const newTodo = {
+      userId,
+      id: this.props.tasks.length + 1,
```
sasha-morozov, line 67: The mentor said that it's an incorrect way to assign a new Id to the property 'id'. Consider using array methods to find the element with the biggest id in the array or the math method to find the biggest integer.

sasha-morozov (approved):
Overall, the code is good. Left one suggestion

danheim (approved):

PR #555, opened 2021-09-13
Solution

Crystallord:
[DEMO LINK](https://crystallord.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,0 +1,145 @@
+import React from 'react';
+import { uuid } from 'uuidv4';
+import { TodoInfo } from '../TodoInfo/TodoInfo';
+
+import './TodoList.scss';
+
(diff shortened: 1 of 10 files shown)


AndreySerhovetc (approved):

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+    const newTodo: Todo = {
+      uuid: uuid(),
+      user: this.props.users.find(user => user.name === this.state.name) || null,
+      userId: 1,
+      title: this.state.title,
+      completed: false,
+    };
+
+    if (!this.state.error) {
```
Elena-brv, line 65: make destructuring for all similar variables

Elena-brv (approved):
if it shows an error about the user first, even if the title is empty, it would be better to place select first, then input 
image

PR #559, opened 2021-09-13
React - Forms - React add TODO form

george-w-boo:
[DEMO LINK](https://george-w-boo.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,19 +1,148 @@
 import React from 'react';
-import './App.css';
+import 'bootstrap/dist/css/bootstrap.min.css';
+import './App.scss';
 
+import { uuid } from 'uuidv4';
 import users from './api/users';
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,19 +1,148 @@
...
+  selected: string;
+  isFilled: boolean;
+  isSelected: boolean;
+}
+
+class App extends React.Component<{}, State> {
+  state = {
+    todos: todosWithUsers,
+    users,
+    todoTitle: '',
```
DmytroMykoliv, line 29: I am not sure if you need to keep users in the state. They have storage on API, so you can use them without keeping them in state

DmytroMykoliv (approved):
Good work )

danheim (approved):

PR #567, opened 2021-09-14
react_add-todo-form

andrii-hryhoruk:
[DEMO LINK](https://andrii-hryhoruk.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,0 +1,155 @@
+import React from 'react';
+import { uuid } from 'uuidv4';
+import { UserInfo } from '../UsersInfo';
+import { TodoInfo } from '../TodoInfo';
+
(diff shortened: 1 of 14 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,155 @@
...
+      });
+    }
+
+    if (!name) {
+      this.setState({
+        isTitle: true,
+      });
+    }
+
+    const maxId = Math.max.apply(null, todos.map(todo => todo.id)) + 1;
```
sasha-morozov, line 66: It seems that uuid is redundant here since you're using Math.max.apply to find the biggest id among the objects.

sasha-morozov (approved):
Great job!

Elena-brv (changes requested):
the functionality works wrong. The error occurs only about the title. Even if title is not empty
image

danheim (approved):

PR #645, opened 2021-11-15
solution without optional

podlesnyi-pavel:
[DEMO LINK](https://podlesnyi-pavel.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,19 +1,153 @@
 import React from 'react';
-import './App.css';
+import './App.scss';
 
 import users from './api/users';
+import todos from './api/todos';
+import { Todo } from './types/TodoType';
+import { TodosList } from './components/TodosList';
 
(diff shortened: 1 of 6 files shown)


DarinaSavytska (approved):
Good job 👍

src/App.tsx
```diff
@@ -1,19 +1,49 @@
...
+      userId,
+      completed: false,
+    };
+
+    this.setState({
+      todosList: [
+        ...todosList,
+        newTodo,
+      ],
+    });
```
VinnieJ-2k20, line 33: Use functional setState here.

VinnieJ-2k20 (changes requested):
Looks good, but let's use functional setState when we rely on data from previous state.

VinnieJ-2k20 (approved):
Good job 👍

PR #664, opened 2021-11-27
Create TODO-list

dmtrhrytsak:
[TODO-list](https://dmtrhrytsak.github.io/react_add-todo-form/)


src/App.scss
```diff
@@ -0,0 +1,75 @@
+* {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0;
```
aionova, line 4: -_-

src/App.scss
```diff
@@ -0,0 +1,75 @@
...
+body {
+  font-family: Arial, Helvetica, sans-serif;
+}
+
+ul {
+  list-style: none;
+}
+
+.App {
+  max-width: 1440px;
```
aionova, line 16: you'd better use variables for magic numbers

src/App.scss
```diff
@@ -0,0 +1,75 @@
...
+  margin: 0 auto;
+  padding: 60px;
+}
+
+.AddTodo {
```
aionova, line 21: creating a separate file for each BEM-block is a good rule to follow

src/App.scss
```diff
@@ -0,0 +1,75 @@
...
+      background-color: #7b9b92;
+    }
+  }
+
+  &__error {
+    margin-top: 12px;
+    padding: 10px;
+    border-radius: 2px;
+
+    background-color: #ffdeea;
```
aionova, line 72: variables for colors

src/App.tsx
```diff
@@ -1,19 +1,183 @@
...
+import { User } from './types/User';
+
+const preparedTodos: Todo[] = todosFromServer.map((todo) => {
+  const todoWithOwner = {
+    ...todo,
+    user: users.find((user) => user.id === todo.userId) || null,
+  };
+
+  return todoWithOwner;
+});
```
aionova, line 18:
```suggestion
const preparedTodos: Todo[] = todosFromServer.map((todo) => ({
  ...todo,
  user: users.find((user) => user.id === todo.userId) || null,
}));
```

src/App.tsx
```diff
@@ -1,19 +1,183 @@
...
+  userSelected: boolean;
 };
 
+class App extends React.Component<{}, State> {
+  state: State = {
+    todos: [...preparedTodos],
+    taskDetails: '',
+    user: '',
+    taskDetailsProvided: true,
+    userSelected: true,
```
aionova, line 34:
```suggestion
    taskDetailsProvided: true,
    isUserSelected: true,
```

src/App.tsx
```diff
@@ -1,19 +1,183 @@
...
+      title: taskDetails,
+      completed: false,
+      user: currentUser,
+    };
+
+    this.setState((state) => {
+      return {
+        todos: [...state.todos, newTodo],
+      };
+    });
```
aionova, line 80:
```suggestion
    this.setState((state) => ({
      todos: [...state.todos, newTodo],
    }));
```

src/App.scss
aionova: in scss we usually don't use camel-case or whatever this is)

aionova (changes requested):

src/App.scss
dmtrhrytsak: -.-

dmtrhrytsak: image

dmtrhrytsak: I’d call it neither best nor bad practice

dmtrhrytsak: But since the performance doesn't matter now, it's just easier to reset them like this

aionova: https://blog.bitsrc.io/using-the-css-star-selector-to-zero-out-margins-and-paddings-bf4c074b83dc

We prefer using reset and normalize files instead. In production code you don't use * selector for that. If that's crucial for you, use it

dmtrhrytsak: Okay)

aionova (approved):
GJ!

PR #673, opened 2021-12-14
Add todo form

Vitaliy-0:
[DEMO](https://vitaliy-0.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,19 +1,169 @@
 import React from 'react';
-import './App.css';
+import './App.scss';
 
+import { TodoList } from './components/TodoList';
+import { todos as todosFromServer } from './api/todos';
 import users from './api/users';
+import { Todo } from './types/Todo';
 
-const App: React.FC = () => {
-  return (
-    <div className="App">
-      <h1>Add todo form</h1>
+const preparedTodos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: users.find(user => user.id === todo.userId) || null,
+}));
 
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
-    </div>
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,19 +1,169 @@
...
+        isTitleEmpty: false,
+      });
+    }
+  };
+
+  onSubmit = (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
+    const { userId, todoTitle } = this.state;
+
+    if (userId === 0) {
+      this.setState(({ isUserUnselected: true }));
```
Val-space, line 53:
```suggestion
      this.setState({ isUserUnselected: true });
```

src/App.tsx
```diff
@@ -1,19 +1,169 @@
...
+    }
+
+    if (todoTitle.trim() === '') {
+      this.setState(({ isTitleEmpty: true }));
```
Val-space, line 57:
```suggestion
      this.setState({ isTitleEmpty: true });
```

src/App.tsx
```diff
@@ -1,19 +1,169 @@
...
+  render() {
+    const {
+      todos,
+      todoTitle,
+      userId,
+      isUserUnselected,
+      isTitleEmpty,
+    } = this.state;
+
+    const visibleTodos = [...todos];
```
Val-space, line 110: gmm...

Val-space (approved):
👌🏼

src/App.tsx
ihor-jpeg: not needed

ihor-jpeg (changes requested):

AlinaMatuschak (approved):

PR #678, opened 2021-12-15
Solution

volodymyr-soltys97:
[DEMO LINK](https://volodymyr-soltys97.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,19 +1,139 @@
 import React from 'react';
-import './App.css';
+import './App.scss';
+import { TodoList } from './components/TodoList/TodoList';
+import { Todo } from './types/types';
 
 import users from './api/users';
+import todos from './api/todos';
 
-const App: React.FC = () => {
-  return (
-    <div className="App">
-      <h1>Add todo form</h1>
+interface State {
+  todos: Todo[],
+  title: string,
+  select: number,
+  choosenTitle: boolean,
+  choosenPerson: boolean,
+}
 
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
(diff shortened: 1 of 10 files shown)


skachkov23 (approved):
GJ

src/App.tsx
```diff
@@ -1,19 +1,139 @@
...
+
+  checkErrors = () => {
+    this.setState(state => {
+      // eslint-disable-next-line @typescript-eslint/no-shadow
+      const { title, select } = state;
+
+      return {
+        choosenTitle: Boolean(title),
+        choosenPerson: Boolean(select),
+      };
```
ihor-jpeg, line 63: where are you using this return value?

src/App.tsx
```diff
@@ -1,19 +1,139 @@
...
+
+    if (title && select) {
+      this.setState(state => ({
+        todos: [...state.todos, {
+          title: state.title,
+          id: state.todos.length + 1,
+          userId: state.select,
+        }],
+      }));
+    }
```
ihor-jpeg, line 40: you could optimize and not let the code go further if there are errors. like an early return statement

src/App.tsx
```diff
@@ -1,19 +1,139 @@
...
+              type="text"
+              name="title"
+              placeholder="Enter the title"
+              className="form__title--width"
+              maxLength={35}
+              value={this.state.title}
+              onChange={this.handleChange}
+            />
+
+            {!this.state.choosenTitle
```
ihor-jpeg, line 95: destructurring

ihor-jpeg (approved):
Good work! Have a look at the comments

PR #680, opened 2021-12-15
Solution

olya-shyian:
[DEMO LINK](https://olya-shyian.github.io/react_add-todo-form/)


i-stanko (approved):

src/App.tsx
```diff
@@ -1,19 +1,46 @@
...
-        {users.length}
-      </p>
-    </div>
-  );
-};
+const preparedTodos = todos.map(todo => {
+  const user = users.find(person => person.id === todo.userId) || null;
+
+  return { ...todo, user };
+});
```
ihor-jpeg, line 14: very concise and clean 👍

src/Components/AddTodo/AddTodo.tsx
```diff
@@ -0,0 +1,140 @@
...
+      this.setState({ titleError: 'Please, enter valid title' });
+
+      return;
+    }
+
+    if (userId === 0) {
+      this.setState({ userError: 'Please, choose user' });
+
+      return;
+    }
```
ihor-jpeg, line 64: you can think about moving this to a separate function and call it validateInput or something

src/Components/AddTodo/AddTodo.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+    const newTodo = {
+      userId,
+      id: this.props.todos.length + 1,
```
ihor-jpeg, line 68: this will only work if the ids match the indices. it might not always be the case

src/Components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+  return (
+    <div className="list">
+      {todos.map((todo: Todo) => (
+        <div
+          className="list__item"
+          key={todo.id}
+        >
+          <TodoInfo todo={todo} />
+        </div>
```
ihor-jpeg, line 21: here you are wrapping the TodoInfo into a div, but in the TodoInfo component you are using the fragment <></>. maybe move this div to the TodoInfo and get rid of the fragment?

ihor-jpeg (approved):
Good work! But have a look at the comments

PR #687, opened 2021-12-16
Solution

aldmpr:
[DEMO LINK](https://aldmpr.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,18 +2,48 @@ import React from 'react';
 import './App.css';
 
 import users from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { NewTodo } from './components/NewTodo';
 
-const App: React.FC = () => {
-  return (
-    <div className="App">
-      <h1>Add todo form</h1>
+const preparedTodos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  id: '',
+  user: users.find(person => person.id === todo.userId) || null,
+}));
 
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
-    </div>
-  );
-};
+interface State {
+  todos: Todo[];
+}
+
+class App extends React.Component<{}, State> {
+  state: State = {
+    todos: preparedTodos,
+  };
+
+  addNewTodo = (todo: Todo) => {
+    this.setState(prevState => ({
+      todos: [...prevState.todos, todo],
+    }));
+  };
+
+  render() {
+    const { todos } = this.state;
+
+    return (
+      <div className="App">
+        <div className="todos text-center">
+          <h1 className="mb-5">Add todo form</h1>
+          <NewTodo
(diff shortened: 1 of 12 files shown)


SerhiiKirik (approved):
Well done 🚀

ihor-jpeg (changes requested):
Good work! But peredelai

src/components/NewTodo.tsx
```diff
@@ -0,0 +1,130 @@
...
+      }));
+
+      return;
+    }
+
+    this.setState(prevState => {
+      const newTodo = {
+        user: users.find(user => user.id === userId) || null,
+        userId,
+        id: todos.length + 1,
```
AlinaMatuschak, line 59: id is a maximal id in the array + 1
What if tomorrow will you add the delete method?

src/components/NewTodo.tsx
```diff
@@ -0,0 +1,130 @@
...
+    if (!userId || !title.trim()) {
+      this.setState(prevState => ({
+        isUserEmpty: !prevState.userId,
+        isTitleEmpty: !prevState.title.trim(),
+      }));
+
+      return;
+    }
+
+    this.setState(prevState => {
```
AlinaMatuschak, line 55: Why do you use setState here? Does your callback return something?

src/components/NewTodo.tsx
```diff
@@ -0,0 +1,130 @@
...
+    return (
+      <form
+        onSubmit={this.handleSubmit}
+        className="form-inline"
+        noValidate
+      >
+        <h4 className="mb-5">Add a new task:</h4>
+
+        <div className="mb-5">
+          <input
```
AlinaMatuschak, line 87: add labels where they are needed

AlinaMatuschak (changes requested):

src/App.tsx
```diff
@@ -2,18 +2,156 @@ import React from 'react';
...
+    }));
+  };
+
+  validateSubmit = () => {
+    const { title, user } = this.state;
+
+    this.setState({
+      isTitleEmpty: title === '',
+      isUserEmpty: user === '',
+    });
```
AlinaMatuschak, line 74:
```suggestion
    this.setState({
      isTitleEmpty: !!title,
      isUserEmpty: !!user,
    });
```

src/App.tsx
```diff
@@ -2,18 +2,156 @@ import React from 'react';
...
+    } = this.state;
+
+    return (
+      <div className="App">
+        <div className="todos text-center">
+          <h1 className="mb-5">Add todo form</h1>
+          <form
+            onSubmit={(event) => {
+              event.preventDefault();
+              this.validateSubmit();
```
AlinaMatuschak, line 98: your validateSubmit method return form valid value, but you don't catch this
don't call addTodo and clearForm if form inputs don’t valid

AlinaMatuschak (approved):

PR #697, opened 2021-12-16
solution without styles

dlugash:
DEMO LINK](https://dlugash.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,19 +1,152 @@
+/* eslint-disable react/no-unused-state */
 import React from 'react';
 import './App.css';
-
+import todos from './api/todos';
 import users from './api/users';
 
-const App: React.FC = () => {
-  return (
-    <div className="App">
-      <h1>Add todo form</h1>
+import { TodoList } from './components/TodoList';
+import { Todo } from './types';
 
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
-    </div>
-  );
-};
+const preparedTodos: Todo[] = todos.map((todo) => {
+  return {
+    ...todo, // old properties
+    user: users.find(user => user.id === todo.userId) || null, // + user or null
+  };
+});
+
+interface State {
+  todos: Todo[];
+  newTodoTitle: string;
+  userId: number;
+  // isUserSelected: boolean;
+  // isNewTodoProvided: boolean;
+}
+
+export class App extends React.Component<{}, State> {
+  state: State = {
+    todos: [...preparedTodos],
+    newTodoTitle: '',
+    userId: 1,
+    // isUserSelected: false,
+    // isNewTodoProvided: false,
+  };
+
+  handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    // event.preventDefault();
+    const text = event.target.value;
+
+    this.setState({
+      newTodoTitle: text,
+    });
+
+    // eslint-disable-next-line no-console
+    console.log(this.state.newTodoTitle);
+
+    // if (text) {
+    //   this.setState({
+    //     isNewTodoProvided: true,
+    //   });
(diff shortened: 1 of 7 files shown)


VladDyrenkov (approved):
GJ, and delete the comment in the files

src/App.tsx
```diff
@@ -1,19 +1,124 @@
...
+  };
+
+  handleSubmit = () => {
+    const { userId, newTodoTitle } = this.state;
+
+    if (!newTodoTitle) {
+      return;
+    }
+
+    const nextID = Math.max(...todos.map(todo => todo.id)) + 1;
```
ArtemTopolskyi, line 57: :fire:

src/App.scss
```diff
@@ -0,0 +1,68 @@
...
+.App {
+  display: flex;
+  flex-direction: column;
+
+  &__form {
+    margin: auto;
+
+    &--form {
+      // height: 80px;
+      // width: 300px;
```
ArtemTopolskyi, line 20: don't left comments

src/App.tsx
```diff
@@ -1,19 +1,124 @@
...
+interface State {
+  todos: Todo[];
+  newTodoTitle: string;
+  userId: number;
+}
+
+export class App extends React.Component<{}, State> {
+  state: State = {
+    todos: [...preparedTodos],
+    newTodoTitle: '',
```
ArtemTopolskyi, line 26: bad variable name

src/App.tsx
```diff
@@ -1,19 +1,124 @@
...
+    userId: 1,
```
ArtemTopolskyi, line 27: image

ArtemTopolskyi (changes requested):
create separate folder for every component

src/App.tsx
```diff
@@ -1,19 +1,125 @@
...
+  todos: Todo[];
+  todoTitle: string;
+  userId: number;
+}
+
+export class App extends React.Component<{}, State> {
+  state: State = {
+    todos: [...preparedTodos],
+    todoTitle: '',
+    userId: 1,
```
ArtemTopolskyi, line 27: если создать todo, не выбрав юзера - все равно создастся объект с первым юзером, потому что захардкожен id 1. Это нужно исправить

ArtemTopolskyi (changes requested):

src/App.tsx
```diff
@@ -1,19 +1,125 @@
...
+  handleSubmit = () => {
+    const { userId, todoTitle } = this.state;
+
+    if (!todoTitle) {
+      return;
+    }
+
+    const nextID = Math.max(...todos.map(todo => todo.id)) + 1;
+
+    const currentUser = users.find(user => user.id === userId);
```
illiarerun, line 59: Check if this is undefined, and if it is, you shouldn't create a user, that way you won't need to hardcode 0 or 1 as id

src/App.tsx
```diff
@@ -1,19 +1,125 @@
...
+  return {
+    ...todo,
+    user: users.find(user => user.id === todo.userId) || null,
+  };
+});
+
+interface State {
+  todos: Todo[];
+  todoTitle: string;
+  userId: number;
```
illiarerun, line 20: And it'd be better to set it as the null at the start
```suggestion
  userId: number | null;
```

src/App.tsx
```diff
@@ -1,19 +1,125 @@
...
+      user: currentUser,
+    };
+
+    this.setState((state) => ({
+      todos: [
+        ...state.todos,
+        newTodo,
+      ],
+      todoTitle: '',
+      userId: 1,
```
illiarerun, line 75: And here also

illiarerun (changes requested):

PR #700, opened 2021-12-16
Solution

alinaShchavinska:
[DEMO LINK](https://alinaShchavinska.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -12,7 +12,7 @@ const preparedTodos: Todo[] = todos.map(todo => ({
 }));
 
 type State = {
-  todosFromServer: Todo[],
+  todos: Todo[],
   title: string,
   user: User | null,
   selectUserName: string,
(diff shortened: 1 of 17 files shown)


tmrsl (approved):
GJ!

src/App.tsx
```diff
@@ -1,19 +1,135 @@
...
+
+  addUser = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    const { value } = event.target;
+
+    const selectedUser = users.find(user => user.name === value);
+
+    this.setState({
+      selectUserName: value,
+      user: selectedUser || null,
+    });
```
ihor-jpeg, line 48: maybe we should not add a todo if the user is not found? who will be doing the todo then?

ihor-jpeg (approved):
Good work! But have a look at the comments

PR #722, opened 2022-01-25
My solution

SavelovSerhii:
https://savelovserhii.github.io/react_add-todo-form/


src/App.tsx
```diff
@@ -1,19 +1,114 @@
...
+            }}
+            className="App__input"
+          >
+            <option value="">Choose a user</option>
+            {serverUsers.map((user) => {
+              return (
+                <option value={user.id}>{user.name}</option>
+              );
+            })}
+          </select>
```
JHesher, line 95:
```suggestion
            <option value="">Choose a user</option>
            {serverUsers.map((user) => (
                <option value={user.id}>{user.name}</option>
              ))}
          </select>
```

JHesher (approved):

src/App.tsx
Olexandr6: Agree with @JHesher, it looks better

src/App.css
```diff
@@ -1,3 +1,32 @@
...
+  display: flex;
+  flex-direction: column;
+  padding: 5px;
+  width: 300px;
+  background-color: rgb(245, 222, 179);
+  border: 3px solid black;
+  border-radius: 16px;
+}
+
+.App__input {
```
Olexandr6, line 28: Use SCSS, instead. In the project, scss package already included, so you should use it

src/App.tsx
```diff
@@ -1,19 +1,114 @@
...
+  render() {
+    return (
+      <div className="App">
+        <h1 className="App__title">Add todo form</h1>
+        <form
+          onSubmit={(event) => {
+            event.preventDefault();
+            this.addTodo();
+            this.setState({ todoTitle: '', currentUser: 0 });
+          }}
```
Olexandr6, line 67: Create separate handler function for this logic

src/App.tsx
```diff
@@ -1,19 +1,114 @@
...
+          className="App__form"
+        >
+          <input
+            type="text"
+            name="title"
+            placeholder="Title"
+            value={this.state.todoTitle}
+            onChange={(event) => {
+              this.setState({ todoTitle: event.target.value, titleError: '' });
+            }}
```
Olexandr6, line 77: Create a separate handler function for this

src/App.tsx
```diff
@@ -1,19 +1,114 @@
...
+            className="App__input"
+          />
+          {this.state.titleError}
+          <select
+            name="users"
+            value={this.state.currentUser}
+            onChange={(event) => {
+              this.setState({ currentUser: +event.target.value, userError: '' });
+            }}
```
Olexandr6, line 86: The same case...

Olexandr6 (changes requested):
Check comments 👇

src/App.tsx
```diff
@@ -1,19 +1,120 @@
...
+            name="users"
+            value={this.state.currentUser}
+            onChange={this.handleUser}
+            className="App__input"
+          >
+            <option value="">Choose a user</option>
+            {serverUsers.map((user) => {
+              return (
+                <option value={user.id}>{user.name}</option>
+              );
```
Denchepornyuk, line 99: remove return from map

Denchepornyuk (approved):

PR #732, opened 2022-01-26
done

Volodymyr1806:
[DEMO LINK](https://Volodymyr1806.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,26 +4,24 @@ import { TodoList } from './components/TodoList/TodoList';
 import todosFromServer from './api/todos';
 import users from './api/users';
 
-const newTodos: Todo[] = todosFromServer.map(todo => ({
+const todosWithUser: TodoWithUser[] = todosFromServer.map(todo => ({
   ...todo,
   user: users.find(user => user.id === todo.userId) || null,
 }));
 
-type Props = {};
-
 type State = {
-  todos: Todo[],
+  todos: TodoWithUser[],
   newTitle: string,
-  selectedId: string,
+  selectedUserId: string,
   hasTitleError: boolean,
   hasUserSelectedError: boolean,
 };
 
-class App extends React.PureComponent<Props, State> {
+class App extends React.PureComponent<{}, State> {
   state: State = {
-    todos: [...newTodos],
+    todos: [...todosWithUser],
     newTitle: '',
-    selectedId: '0',
+    selectedUserId: '0',
     hasTitleError: false,
     hasUserSelectedError: false,
   };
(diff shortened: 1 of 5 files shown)


Valeria-Valuyskaya (approved):

src/App.tsx
```diff
@@ -1,19 +1,138 @@
...
+      }));
+    } else if (value.length > 1) {
+      this.setState((state) => ({
+        newTitle: state.newTitle,
+      }));
+    } else {
+      this.setState(() => ({
+        newTitle: '',
+      }));
+    }
```
Olexandr6, line 49: It's the unnecessary condition, cause if it's empty title, you can use previous condtion

Olexandr6, line 49: I mean:
```suggestion
    if (regexp.test(value)) {
      this.setState((state) => ({
        ...state,
        newTitle: value,
        hasTitleError: false,
      }));
    }
    
	  this.setState(prevState => ({
	     newTitle: prevState.newTitle,
	  });
```

Olexandr6: Test it, it should work correctly

src/components/UserInfo/UserInfo.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import './UserInfo.scss';
+
+export const UserInfo: React.FC<User> = ({ name, email }) => (
+  <div className="shortInfoOfUser">
+    <strong className="shortInfoOfUser__name">{name}</strong>
+    <span
+      className="shortInfoOfUser__email"
+    >
```
Olexandr6, line 9:
```suggestion
    <span className="shortInfoOfUser__email">
```

Olexandr6 (changes requested):
In general, good solution 👍 But some things should be fixed, check comments 👇

Also, move your form to the top of the page, cause now if I add many todos, I have to scroll down to add a new one. This is an example of bad UX.

src/App.tsx
Volodymyr1806: I've tried this code  again, but unfortunately it works wrong and I don't  still understand why
When I write anything , then I cannot remove first character in input , therefore I did one more 'else'

Olexandr6: I was wrong, I forgot to suggest `return` in the first condition, then it will work corectly

src/App.tsx
```diff
@@ -1,19 +1,132 @@
...
+      this.setState((state) => ({
+        ...state,
+        newTitle: value,
+        hasTitleError: false,
+      }));
+    } else {
+      this.setState({
+        newTitle: '',
+      });
+    }
```
Olexandr6, line 43:
```suggestion
    if (value && regexp.test(value)) {
      return this.setState((state) => ({
        ...state,
        newTitle: value,
        hasTitleError: false,
      }));
    }
    
    this.setState({
      newTitle: '',
    });
```

Olexandr6: Now, it will work correctly in such way)

Olexandr6 (approved):
Great job 🔥

PR #797, opened 2022-04-03
Add Todo Form

aspurgilus:
[DEMO LINK](https://aspurgilus.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -56,8 +56,6 @@ const App: React.FC = () => {
     }
   };
 
-  const clearInputs = () => resetForm();
-
   const changeTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
     const { value } = event.target;
     const pattern = /^[0-9enru ]*$/;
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -1,17 +1,148 @@
...
+    if (title && userId) {
+      addTodo(prevTodos => [...prevTodos, {
+        userId,
+        id: prevTodos.length + 1,
+        title,
+        completed: false,
+        user: usersFromServer.find(user => user.id === userId) || null,
+      }]);
+
+      resetForm();
```
krolich500grn, line 55: nice idea to move the reset into another function

krolich500grn (approved):
Nice idea to use Bulma here 👌

I got problems when tried to input the 'a-z' and 'а-я' symbols. 
And I'm still able to create empty Todo with only spaces input.

But the validation is a task with *, so you got approve for the main task.

aspurgilus:
> I got problems when tried to input the 'a-z' and 'а-я' symbols. And I'm still able to create empty Todo with only spaces input.
> But the validation is a task with *, so you got approve for the main task.
I don't know what regex I should use. So, I'll be waiting for mentor suggestion. Thank's for the review.

Ihor-Karpyn (approved):

PR #807, opened 2022-04-05
Solution

RomaBoreiko:
[DEMO LINK](https://RomaBoreiko.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,106 @@
-import React from 'react';
+import React, { useState } from 'react';
+import classnames from 'classnames';
 import './App.css';
+import todosFromServer from './api/todos';
+import { TodoList } from './Components/TodoList/TodoLIst';
 
 import users from './api/users';
 
+const preparedTodos = todosFromServer.map(todo => {
+  return {
+    ...todo,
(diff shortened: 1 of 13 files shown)


jursylian (approved):
Looks so nice!

src/App.tsx
```diff
@@ -1,17 +1,106 @@
...
+  const onFormSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    setHasTitleError(!title);
+    setHasUserIdError(!userId);
+
+    if (title && userId) {
+      const newTodo = {
+        userId,
+        id: preparedTodos.length + 1,
```
Ihor-Karpyn, line 35: should place in functional `setState`. And why you get length from `preparedTodos`.  You add new todos in `todos` but get new id from `preparedTodos` which don't change

src/App.tsx
```diff
@@ -1,17 +1,106 @@
...
+        title,
+        completed: false,
+        user: users.find(user => user.id === userId)
+          || null,
```
Ihor-Karpyn, line 39: Create function get user

Ihor-Karpyn (changes requested):

PR #809, opened 2022-04-05
Develop

jursylian:
[DEMO LINK](https://jursylian.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,42 +2,57 @@ import React, { useState } from 'react';
 import './App.css';
 import { TodoList } from './components/TodoList/TodoList';
 import users from './api/users';
-import todos from './api/todos';
+import todosFromServer from './api/todos';
+
+const preparedTodos = todosFromServer.map(todo => {
+  return {
+    ...todo,
+    user: users.find(user => user.id === todo.userId) || null,
+  };
+});
 
 const App: React.FC = () => {
+  const [todos, setTodos] = useState([...preparedTodos]);
   const [title, setTitle] = useState('');
-  const [hasTitleError, setTitleError] = useState(false);
+  const [hasTitleError, setHasTitleError] = useState(false);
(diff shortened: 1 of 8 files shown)


RomaBoreiko (approved):
Fire UI!

src/App.tsx
```diff
@@ -1,17 +1,119 @@
...
+    if (!userId) {
+      setHasUserIdError(true);
+    }
+
+    if (title && userId) {
+      const todo = {
+        id: unicId,
+        title,
+        completed: false,
+        userId: +userId,
```
Ihor-Karpyn, line 38:
```suggestion
        userId,
```

src/App.tsx
```diff
@@ -1,17 +1,119 @@
...
+      setHasTitleError(true);
+    }
+
+    if (!userId) {
+      setHasUserIdError(true);
+    }
+
+    if (title && userId) {
+      const todo = {
+        id: unicId,
```
Ihor-Karpyn, line 35: `unicId` this variable get value from `prevState`,  so it's should be place in functional `setState`

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,46 @@
...
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <div className="todo">
+      <ul className="todo__list">
+        {todos.map(({ id, title, userId }) => {
+          const person = users.find(user => userId === user.id);
```
Ihor-Karpyn, line 16: getUserById function and use it in prepared todo too

Ihor-Karpyn (changes requested):

src/App.tsx
jursylian: I chose using Date.now()

solaryasha (approved):

PR #811, opened 2022-04-06
solution

vitaliy-zhyvyliak:
[DEMO](https://vitaliy-zhyvyliak.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,16 +15,17 @@ const App: React.FC = () => {
   const [hasUserIdError, setHasUserIdError] = useState(false);
 
   const addNewTodo = () => {
-    const todoId = visibleTodos.length + 1;
+    setVisibleTodos((currTodos => {
+      const todoId = currTodos.length + 1;
+      const newTodo: Todo = {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,17 +1,99 @@
...
+    const todoId = visibleTodos.length + 1;
+
+    const newTodo: Todo = {
+      id: todoId,
+      userId,
+      title,
+      completed: false,
+    };
+
+    setVisibleTodos((currTodos => [...currTodos, newTodo]));
```
Ihor-Karpyn, line 27:
```suggestion
  
    setVisibleTodos((currTodos => {
    	const todoId = currTodos.length + 1; // also use Math.max to count new id

	      const newTodo: Todo = {
	        id: todoId,
	        userId,
	        title,
	        completed: false,
	      };
	      
	      return [...currTodos, newTodo]
    }));
```

Ihor-Karpyn (changes requested):

Ihor-Karpyn (approved):

PR #813, opened 2022-04-06
last saved

VitalyPasechnik:


src/App.tsx
```diff
@@ -1,19 +1,105 @@
...
+import todos from './api/todos';
 import users from './api/users';
 
-const App: React.FC = () => {
+import { ITodo } from './types/Interface';
+import { TodoList } from './components/TodoList/TodoList';
+
+const preparedTodos: Array<ITodo> = todos.map(todo => {
+  return { ...todo, user: users.find(user => user.id === todo.userId) || null };
+});
```
ArtemTopolskyi, line 13:
```suggestion
const preparedTodos: Array<ITodo> = todos.map(todo => ({
  ...todo,
  user: users.find(user => user.id === todo.userId) || null,
}));
```

src/App.tsx
```diff
@@ -1,19 +1,105 @@
...
+    setUserIdError(!currentUserId);
+
+    if (title && currentUserId) {
+      const todo = {
+        userId: currentUserId,
+        id: todos.length + 1,
+        title,
+        completed: false,
+        user: users.find(user => user.id === currentUserId) || null,
+      };
```
ArtemTopolskyi, line 44: maybe move it to separate method 'createTodo' or something like that

PR #868, opened 2022-05-18
solution

StasivPavlo:
[DEMO LINK](https://StasivPavlo.github.io/react_add-todo-form/)


viacheslav-petrovskyi (approved):
really good job, but if i choose a person, and then i can choose 'Select user', and send task without person, better disabled 'Select user'

PR #882, opened 2022-06-07
solution

AnastasiyaSopotnytska:
[DEMO LINK](https://AnastasiyaSopotnytska.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,100 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 
-import users from './api/users';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { FullTodo, Todo, User } from './react-app-env';
+import TodoList from './components/TodoList/TodoList';
 
 const App: React.FC = () => {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,17 +1,100 @@
...
+          className="level level-left m-2"
+          htmlFor="titleInput"
+        >
+          Title of Todo:
+        </label>
+        <input
+          type="text"
+          id="titleInput"
+          className="input is-link level level-left column is-2 ml-1"
+          required
```
SavMary, line 57: The task requires that after pressing Add we see `Please enter the title` or `Please choose a user`.  Attribute "required" is incorrect in this case.

src/App.tsx
```diff
@@ -1,17 +1,100 @@
...
+          value={newTitle}
+          onChange={(event) => setNewTitle(event.target.value)}
+        />
+
+        <select
+          name="user"
+          id="selectUser"
+          className="select is-link level level-left ml-1"
+          required
```
SavMary, line 66: The same here

SavMary (changes requested):

PR #888, opened 2022-06-08
React add todo form

dasha-chekmariova:
[DEMO LINK](https://dasha-chekmariova.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,127 @@
-import React from 'react';
-import './App.css';
+/* eslint-disable jsx-a11y/label-has-associated-control */
+import React, { useState } from 'react';
+import './App.scss';
 
+import todos from './api/todos';
 import users from './api/users';
+import { TodoList } from './components/TodoList/TodoList';
+import { TodoType } from './react-app-env';
+
+const preparedTodos: TodoType[] = todos.map(todo => (
+  {
+    ...todo,
+    user: users.find(user => user.id === todo.userId) || null,
+  }
+));
 
 const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [todosArr, setTodosArr] = useState(preparedTodos);
+  const [selectNameId, setSelectNameId] = useState(0);
+  const [completed, setCompleted] = useState(false);
+  const [errorInput, setErrorInput] = useState('');
+  const [errorSelect, setErrorSelect] = useState('');
+
+  const handle = (newTodo: TodoType) => {
+    setTodosArr(currentTodo => [...currentTodo, newTodo]);
+    setErrorInput('');
+    setErrorSelect('');
+    setQuery('');
+    setSelectNameId(0);
+  };
+
+  const submitEvent = (
+    event: React.FormEvent<HTMLFormElement>,
+  ) => {
+    event.preventDefault();
+
+    const newTodo = {
+      userId: Date.now(),
+      id: Date.now(),
+      title: query,
+      completed,
+      user: users.find(user => user.id === selectNameId) || null,
+    };
+
+    if (query && selectNameId) {
+      handle(newTodo);
+    }
+
+    if (!query) {
+      setErrorInput('Please enter the title');
+    }
+
+    if (!selectNameId) {
+      setErrorSelect('Please choose a user');
+    }
+  };
+
   return (
     <div className="App">
-      <h1>Add todo form</h1>
+      <h1 className="title is-1 is-spaced mainTitle">Static list of todos</h1>
+      <form
+        onSubmit={submitEvent}
+      >
+        <div>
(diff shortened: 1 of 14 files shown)


dellachaise (approved):
Good job! Looks great!

src/App.tsx
```diff
@@ -1,17 +1,139 @@
...
+
+          <label>Completed</label>
+
+        </div>
+
+        <button
+          className="button"
+          type="submit"
+        >
+          App
```
dellachaise, line 131: Please, rename Button to Add from App.

src/App.tsx
```diff
@@ -1,17 +1,139 @@
...
+    setCompleted(false);
+  };
+
+  const submitEvent = (
+    event: React.FormEvent<HTMLFormElement>,
+  ) => {
+    event.preventDefault();
+
+    const newTodo = {
+      userId: Date.now(),
```
Denchepornyuk, line 41: must be selectUserId

src/App.tsx
```diff
@@ -1,17 +1,139 @@
...
+    ...todo,
+    user: users.find(user => user.id === todo.userId) || null,
+  }
+));
 
 const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [todosArr, setTodosArr] = useState(preparedTodos);
+
+  const [selectNameId, setSelectNameId] = useState(0);
```
Denchepornyuk, line 22:
```suggestion
  const [selectUserId, setSelectUserId] = useState(0);
```

src/App.tsx
Denchepornyuk, line 131:
```suggestion
          Add
```

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { TodoType } from '../../react-app-env';
+import { TodoInfo } from '../TodoInfo/TodoInfo';
+import '../../App.scss';
+
+type Props = {
+  preparedTodos: TodoType[]
+};
+
+export const TodoList: React.FC<Props> = ({ preparedTodos }) => (
+  <>
```
Denchepornyuk, line 11: Do you need React.Fragment here?

src/App.tsx
```diff
@@ -1,17 +1,139 @@
...
+const preparedTodos: TodoType[] = todos.map(todo => (
+  {
+    ...todo,
+    user: users.find(user => user.id === todo.userId) || null,
+  }
+));
 
 const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [todosArr, setTodosArr] = useState(preparedTodos);
```
Anna-Tkachenko, line 20: Don't use data type names in variable names
```suggestion
  const [todos, setTodos] = useState(preparedTodos);
```

src/App.tsx
```diff
@@ -1,17 +1,139 @@
...
+
+const preparedTodos: TodoType[] = todos.map(todo => (
+  {
+    ...todo,
+    user: users.find(user => user.id === todo.userId) || null,
+  }
+));
 
 const App: React.FC = () => {
+  const [query, setQuery] = useState('');
```
Anna-Tkachenko, line 19: Will be better to name it as todoTitle, or something like that, for better understanding

src/App.tsx
```diff
@@ -1,17 +1,139 @@
...
   return (
     <div className="App">
-      <h1>Add todo form</h1>
+      <h1 className="title is-1 is-spaced mainTitle">Static list of todos</h1>
+      <form
+        onSubmit={submitEvent}
+      >
+
+        <div>
+          <p className="error">{errorInput}</p>
```
Anna-Tkachenko, line 69: You shouldn't render tag p if there is no input error

src/App.tsx
```diff
@@ -1,17 +1,139 @@
...
+            onChange={(event) => {
+              setQuery(event.target.value);
+              setErrorInput('');
+            }}
+          />
+
+        </div>
+
+        <div>
+          <p className="error">{errorSelect}</p>
```
Anna-Tkachenko, line 87: Same here, you shouldn't render this tag if there is no errorSelect

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,22 @@
...
+};
+
+export const TodoList: React.FC<Props> = ({ preparedTodos }) => (
+  <ul>
+    {preparedTodos.map(item => (
+      <li className="box box-color box--hover" key={item.id}>
+        {item.user
+          && (
+            <TodoInfo todo={item} />
+          )}
```
Anna-Tkachenko, line 17: This check is redundant here
You should render TodoInfo component even if todo has no user (it will be displayed without UserInfo component)

PR #891, opened 2022-06-09
Solution

AndriiRut:
[DEMO LINK](https://andriirut.github.io/react_add-todo-form/)


vlasiuk-anatolii (approved):
Checkbox does not exist.

Denchepornyuk (commented):
Screenshot 2022-06-10 at 12 55 03

the error should disappear when the input isn't empty the same for select

Denchepornyuk (approved):

PR #893, opened 2022-06-09
Develop

vlasiuk-anatolii:
[DEMO LINK](https://vlasiuk-anatolii.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,6 +3,7 @@ import './App.css';
 import { PrepArray } from './react-app-env';
 import users from './api/users';
 import todos from './api/todos';
+import { TodoList } from './components';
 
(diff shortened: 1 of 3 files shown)


Andriyk0 (approved):

src/react-app-env.d.ts
```diff
@@ -1 +1,41 @@
...
+  username: string,
+  email: string,
+  address: Address,
+  phone: string,
+  website: string,
+  company: Company,
+}
+
+export interface PrepArray extends Todo {
+  user: User | undefined,
```
Denchepornyuk, line 40:
```suggestion
  user: User | null,
```
use null instead if undefined

Denchepornyuk (approved):

PR #900, opened 2022-06-09
task is solved

enrikadej:
[DEMO LINK](https://enrikadej.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,128 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 
 import users from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+
+import { User, PrepTodo } from './react-app-env';
+
+const getUserById = (userId: number): User | null => {
+  return users.find(user => user.id === userId) || null;
+};
+
+const todosWithUser: PrepTodo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
 
 const App: React.FC = () => {
+  const [todos, setTodos] = useState<PrepTodo[]>(todosWithUser);
+  const [newTodoName, setNewTodoName] = useState('');
+  const [hasNameError, setHasNameError] = useState(false);
+
+  const [selectUserId, setSelectUserId] = useState(0);
+  const [hasUserIdError, setHasUserIdError] = useState(false);
+
+  const addTodo = (name: string, userId: number) => {
+    const newTodo: PrepTodo = {
+      id: Date.now(),
+      name,
+      userId,
+      completed: false,
+      user: getUserById(selectUserId),
+    };
+
+    setTodos((prevTodos) => [...prevTodos, newTodo]);
+  };
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    setHasNameError(!newTodoName);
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,17 +1,128 @@
...
+      <form
+        onSubmit={handleSubmit}
+      >
+        <label className="">
+          <p className="App__title">Title</p>
+          <input
+            type="text"
+            value={newTodoName}
+            placeholder="write to do"
+            onChange={(event) => {
```
iwanttosmile4u, line 70: It'll be better to transfer callback in onChange event before return and pass only function link.

src/App.tsx
```diff
@@ -1,17 +1,128 @@
...
+            {hasNameError && (
+              <span className="error">Write name of &quot;to do&quot;</span>
+            )}
+          </p>
+        </label>
+
+        <div className="">
+          <select
+            value={selectUserId}
+            onChange={(event) => {
```
iwanttosmile4u, line 86: The same recomendation as before)

iwanttosmile4u (approved):
👍

src/App.tsx
```diff
@@ -1,17 +1,128 @@
...
+
+  const [selectUserId, setSelectUserId] = useState(0);
+  const [hasUserIdError, setHasUserIdError] = useState(false);
+
+  const addTodo = (name: string, userId: number) => {
+    const newTodo: PrepTodo = {
+      id: Date.now(),
+      name,
+      userId,
+      completed: false,
```
Anna-Tkachenko, line 32: You should add checkbox for completed / non completed value that user can choose in the form

src/App.tsx
```diff
@@ -1,17 +1,128 @@
...
+      addTodo(newTodoName, selectUserId);
+      setNewTodoName('');
+      setSelectUserId(0);
+    }
+  };
+
+  // const todosWithUser: PrepTodo[] = todos.map(todo => ({
+  //   ...todo,
+  //   user: getUserById(todo.userId),
+  // }));
```
Anna-Tkachenko, line 55: Delete redundant comments

src/App.tsx
```diff
@@ -1,17 +1,128 @@
...
+              setNewTodoName(event.target.value);
+              setHasNameError(false); // set false in state
+            }}
+            className="App__input"
+          />
+          <p>
+            {hasNameError && (
+              <span className="error">Write name of &quot;to do&quot;</span>
+            )}
+          </p>
```
Anna-Tkachenko, line 80: You shouldn't render tag p if there is no hasNameError

src/App.tsx
```diff
@@ -1,17 +1,128 @@
...
+              <option
+                key={user.id}
+                value={user.id}
+              >
+                {user.name}
+              </option>
+            ))}
+          </select>
+
+          <p>
```
Anna-Tkachenko, line 109: Same here with tag p

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -1,17 +1,146 @@
...
+          </label>
+          <label>
+            should complete
+            <input
+              type="radio"
+              name="complete"
+              value="false"
+              onChange={() => {
+                setWasCompleted(false);
+              }}
```
Anna-Tkachenko, line 129: One of this inputs should be checked by default

Anna-Tkachenko (approved):

PR #902, opened 2022-06-10
add todo form

KostiaO:


enrikadej (changes requested):
it seems that some files are empty
if it helps, try using the command 'git add .' before commit

src/App.tsx
```diff
@@ -1,17 +1,166 @@
...
+
+  const [isCompl, setIsCompl] = useState<boolean>(false);
+
+  const [title, setTitle] = useState<string>('');
+
+  const [selectedUserId, setUserId] = useState<number>(-1);
+
+  const [titleErr, setTitleErr] = useState(false);
+
+  const [userErr, setUserErr] = useState(false);
```
Anna-Tkachenko, line 29: Use full variable names
```suggestion
  const [isCompleted, setIsCompl] = useState<boolean>(false);

  const [title, setTitle] = useState<string>('');

  const [selectedUserId, setUserId] = useState<number>(-1);

  const [titleError, setTitleErr] = useState(false);

  const [userError, setUserErr] = useState(false);
```

PR #903, opened 2022-06-10
solution

vlad-demchuk:
[DEMO LINK](https://vlad-demchuk.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,128 @@
-import React from 'react';
+import React, { useState } from 'react';
+import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
 import './App.css';
 
-import users from './api/users';
+import { TodoList } from './components/TodoList/TodoList';
+import { Todo } from './react-app-env';
 
 const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([...todosFromServer]);
+
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [completed, setCompleted] = useState(false);
+
+  const [titleIsEmpty, setTitleIsEmpty] = useState(false);
+  const [userIdIsEmpty, setUserIdIsEmpty] = useState(false);
+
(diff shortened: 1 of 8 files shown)


ZUBOV-ILLIA (approved):
Wow Your button is magic !!! 

image

src/App.tsx
```diff
@@ -1,17 +1,128 @@
...
+      setUserIdIsEmpty(true);
+    }
+
+    if (title && userId) {
+      setTodos([...todos, {
+        userId,
+        id: todos.length + 1,
+        title,
+        completed,
+      }]);
```
Anna-Tkachenko, line 36: Will be better to use function inside setTodos, if you want to change the state depending on the previous value

src/App.tsx
```diff
@@ -1,17 +1,128 @@
...
+      setTitle('');
+      setUserId(0);
+      setCompleted(false);
+    }
+  };
+
+  const preparedTodos = todos.map(todo => ({
+    ...todo,
+    user: usersFromServer.find(user => todo.userId === user.id) || null,
+  }));
```
Anna-Tkachenko, line 47: You can do in once before the defining react component
And set prepared todos to the state

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,26 @@
...
+  preparedTodos: PreparedTodos[]
+};
+
+export const TodoList: React.FC<Props> = ({ preparedTodos }) => {
+  return (
+    <ul className="list">
+      {preparedTodos.map(todo => (
+        <li key={todo.id} className="item">
+          <p>{todo.user?.name}</p>
+          <p>{todo.user?.email}</p>
```
Anna-Tkachenko, line 15: Will be better to create UserInfo component for displaying this info

Anna-Tkachenko: And you shouldn't render this tags if there is no user

Anna-Tkachenko (approved):

PR #905, opened 2022-06-11
User todo Form

Oleksandr-Ivanchenko:
[DEMO LINK](https://Oleksandr-Ivanchenko.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,145 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 
+import todos from './api/todos';
 import users from './api/users';
+import { PreparedTodo } from './react-app-env';
+import { TodoList } from './componets/TodoList';
+
+const getUserById = (userId: number) => {
+  return users.find(person => person.id === userId) || null;
+};
+
+const usersPrepareTodos: PreparedTodo[] = todos.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
 
 const App: React.FC = () => {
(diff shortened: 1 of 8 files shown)


EnjSadman (approved):

src/App.tsx
```diff
@@ -1,17 +1,145 @@
...
+
+    const newTodo = {
+      id: Date.now(),
+      title: newTodosName,
+      userId: selectUserId,
+      user: getUserById(selectUserId),
+      completed: isCompleted,
+    };
+
+    if (!newTodo.userId || newTodo.title === '') {
```
Anna-Tkachenko, line 69: Will be better to check state variables here, and create newTodo object only if there is no errors

src/componets/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,31 @@
...
+}) => (
+  <ul className="list">
+    {preparedTodos.map((todo: PreparedTodo) => (
+      <li
+        key={todo.id}
+        className="list_item"
+      >
+        <h5>
+          {todo.user?.name}
+        </h5>
```
Anna-Tkachenko, line 20: You shouldn't render this tag if there is no user

src/componets/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,31 @@
...
+        <div>
+          <input
+            type="checkbox"
+            checked={todo.completed}
+          />
```
Anna-Tkachenko, line 25: Will be better to display some text status instead of empty or non empty checkbox

Anna-Tkachenko (approved):
Will be great to add some styles to your project

PR #906, opened 2022-06-12
finished hw with optional task

iwanttosmile4u:
[DEMO LINK](https://iwanttosmile4u.github.io/react_add-todo-form/)


khozhainov-aleksandr (approved):
Good job 🔥

src/App.tsx
```diff
@@ -1,17 +1,149 @@
...
+  const [title, setTitle] = useState('');
+  const [selectUserId, setSelectUserId] = useState(0);
+  const [todos, setTodos] = useState(fullTodo);
+  const [hasTitleError, setHasTitleError] = useState(false);
+  const [hasUserIdError, setUserIdError] = useState(false);
+  const [correctString, setCorrectString] = useState(true);
+
+  const changeTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
+    if (!/^[a-zA-Zа-яА-Я\d\s]+$/.test(event.target.value)) {
+      event.preventDefault();
```
Anna-Tkachenko, line 31: Why did you call preventDefault here?

src/App.tsx
```diff
@@ -1,17 +1,149 @@
...
+      setCorrectString(false);
```
Anna-Tkachenko, line 32: Will be better to have only one variable in state that should be responsible on the title errors

Anna-Tkachenko (approved):

PR #913, opened 2022-06-15
Develop

Danylo-Dziuban:
[DEMO LINK](https://Danylo-Dziuban.github.io/react_add-todo-form/)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+
+import { User } from '../../types';
+
+type Props = {
+  user: User | null,
```
Anna-Tkachenko, line 6: null is redundant here, you made user check before rendering UserInfo component

Anna-Tkachenko (approved):

PR #914, opened 2022-06-16
solution

ShapovalDenys:
http://shapovaldenys.github.io/react_add-todo-form/

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "license": "GPL-3.0",
   "dependencies": {
     "@cypress/react": "^5.12.4",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -7,7 +7,7 @@
   "license": "GPL-3.0",
   "dependencies": {
     "@cypress/react": "^5.12.4",
-    "bulma": "^0.9.3",
+    "bulma": "^0.9.4",
```
vkryvytskyy, line 10: image
Use `.trim()` before adding new to do to the list, to check it's length

vkryvytskyy (approved):

PR #917, opened 2022-06-16
the task is complete except error handling and styling

efdalyalcin:
- [DEMO LINK](https://efdalyalcin.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,7 +19,7 @@ const App: React.FC = () => {
   const [initialTodos, setInitialTodos] = useState(preparedTodos);
   const [todoTitle, setTodoTitle] = useState('');
   const [selectedUser, setSelectedUser] = useState(0);
-  const [todoId, setTodoId] = useState(preparedTodos[lastIndexOfTodos].id)
+  const [todoId, setTodoId] = useState(preparedTodos[lastIndexOfTodos].id);
 
   const addTodo = (title: string, userId: number) => {
     setTodoId((prev) => prev + 1);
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,17 +1,113 @@
...
+        <input
+          type="text"
+          placeholder="Create todo"
+          className={cn({ error: hasValidTitle })}
+          value={todoTitle}
+          onChange={(event: React.ChangeEvent<HTMLInputElement>) => {
+            setHasValidTitle(false);
+            setTodoTitle(event.target.value);
+          }}
+        />
```
amitpatiljc, line 78: Following point is missing from your solution-
if the title is empty, show an error message next to the title field (Please enter the title)

src/App.tsx
```diff
@@ -1,17 +1,113 @@
...
 
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
+      <form
+        onSubmit={hasValidInput}
+      >
+        <input
+          type="text"
```
amitpatiljc, line 70: Using label is good practice , let's use it!

amitpatiljc (changes requested):
Good Job, very clean code. Just have a glace at suggestions

amitpatiljc (approved):
Good Job 👍

src/App.tsx
```diff
@@ -1,17 +1,132 @@
...
+          })}
+        </select>
+
+        <div className="error">
+          <span>
+            {hasUserError && (
+              'Please choose a User'
+            )}
+          </span>
+        </div>
```
meforoma, line 116: you're rendering an empty div still if there's no error
better conditionally render the whole element, not just <span> content

meforoma (approved):
see the 2 non-blocking comments

PR #919, opened 2022-06-16
solution with minimal styling

LukaszGryziewicz:
[DEMO LINK](https://LukaszGryziewicz.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,94 @@
-import React from 'react';
-import './App.css';
-
+import React, { ChangeEvent, useState } from 'react';
+import './App.scss';
 import users from './api/users';
+import todos from './api/todos';
+import { PreparedTodo } from './types/app.typedefs';
+import { TodoList } from './components/TodoList/TodoList';
+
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,17 +1,94 @@
...
+    setShowUserError(!todoUser);
+  };
+
   return (
     <div className="App">
-      <h1>Add todo form</h1>
+      <form onSubmit={handleSubmit}>
+        <input
+          type="text"
+          data-cy="titleInput"
```
amitpatiljc, line 56: Using labels is good practice lets use it!

amitpatiljc (approved):

src/App.tsx
meforoma: yes, labels should serve their purpose - namely if we need to expand the clickable area,
yet in @LukaszGryziewicz case, following the minimal styling there's no real need for a label - it's only an input that you focus at.
So not an actual blocking matter.

meforoma (approved):

PR #922, opened 2022-06-19
Solution

PRO-GRAM-MER:
[DEMO](https://pro-gram-mer.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,17 +4,12 @@ import todos from './api/todos';
 import { TodoList } from './components/TodoList/TodoList';
 import { PreparedTodo } from './react-app-env';
 
-const preparedTodos: PreparedTodo[] = todos.map((todo) => {
-  const user = users.find(person => person.id === todo.userId) || undefined;
-
-  return (
-    {
-      ...todo,
-      user,
-    }
-  );
-});
-
+const preparedTodos: PreparedTodo[] = todos.map((todo) => (
+  {
+    ...todo,
+    user: users.find(person => person.id === todo.userId),
+  }
+));
 const App: React.FC = () => {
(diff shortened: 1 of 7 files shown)


efdalyalcin (approved):
Code is great, just missing some styling...

src/App.tsx
```diff
@@ -1,17 +1,98 @@
...
+const preparedTodos: PreparedTodo[] = todos.map((todo) => {
+  const user = users.find(person => person.id === todo.userId) || undefined;
+
+  return (
+    {
+      ...todo,
+      user,
+    }
+  );
+});
```
meforoma, line 16: fine, but could be shorter
```suggestion
const preparedTodos: PreparedTodo[] = todos.map((todo) => (
    {
      ...todo,
      user: users.find(person => person.id === todo.userId),
    }
  );
);
```

src/App.tsx
```diff
@@ -1,17 +1,98 @@
...
+  const [todoList, setTodoList] = useState(preparedTodos);
+  const [todoTitle, setTodoTitle] = useState('');
+  const [todoUser, setTodoUser] = useState(0);
+  const [showTitleError, setShowTitleError] = useState(false);
+  const [showUserError, setShowUserError] = useState(false);
+
+  const addTodo = (title: string, userId: number):void => {
+    const todo: PreparedTodo = {
+      userId,
+      id: todoList.length + 1,
```
meforoma, line 28: > - `id` is a maximal `id` in the array + 1

not `list.length`
something like this:

meforoma (changes requested):
good, except few comments
please fix

meforoma (approved):

PR #923, opened 2022-06-20
Solution

iLebo-stack:
[DEMO LINK](https://iLebo-stack.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,0 +1,126 @@
+import React, { useState } from 'react';
+import users from '../../api/users';
+import { ToDos } from '../../react-app-env';
+import TodoInfo from '../TodoInfo/TodoInfo';
+import UserInfo from '../UserInfo/UserInfo';
+import './TodoList.css';
+
+interface Props {
+  initialTodos: ToDos[];
+}
+
+const TodoList: React.FC<Props> = ({ initialTodos }) => {
+  const [todos, setTodos] = useState(initialTodos);
+  const [newTitle, setNewTitle] = useState('');
+  const [selectedUser, setSelectedUSer] = useState(0);
+  const [showTitleError, setShowTitleError] = useState(false);
+  const [showSelectErrorMessage, setShowSelectErrorMessage] = useState(false);
+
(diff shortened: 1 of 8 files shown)


efdalyalcin (approved):
Coding is great, styling is not that much 😃

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,125 @@
...
+  const [selectedUser, setSelectedUSer] = useState(0);
+  const [showTitleError, setShowTitleError] = useState(false);
+  const [showSelectErrorMessage, setShowSelectErrorMessage] = useState(false);
+
+  function addTodo() {
+    const newTodo = {
+      title: newTitle,
+      userId: selectedUser,
+      completed: false,
+      id: todos.length + 1,
```
meforoma, line 24: > - `id` is a maximal `id` in the array + 1

not `list.length`
something like this:

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,125 @@
...
+              const title = event
+                .target
+                .value
+                .split('').filter(character => (
+                  /[a-z0-9]/i.test(character) || /\s/.test(character)
+                ))
+                .join('');
+
+              setNewTitle(title);
+            }}
```
meforoma, line 65: let's move it outside JSX - before `return` and just pass it here as a one-line prop
```suggestion
            onChange={handleInputChange}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,125 @@
...
+            onChange={(event) => {
+              setShowTitleError(false);
+
+              const title = event
+                .target
+                .value
+                .split('').filter(character => (
+                  /[a-z0-9]/i.test(character) || /\s/.test(character)
+                ))
+                .join('');
```
meforoma, line 62: also wondering why do you actually need this?
what if user's language has non-iso characters? 🤔

meforoma (changes requested):
good job 👍 
please check few comments left

meforoma (approved):

PR #924, opened 2022-06-20
task completed

andreisucman:


efdalyalcin (changes requested):
Your code looks good but I can't check if it works since the page is not deployed

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+interface Props {
+  preparedTodos: PreparedTodos[];
+}
+
+export const TodoList: React.FC<Props> = ({ preparedTodos }) => {
+  return (
+    <ul>
+      {preparedTodos.map(todo => {
+        return (
```
meforoma, line 13: not a big deal, but implicit returns are suitable

meforoma (approved):
just a few missed implicit return,
but otherwise good code 👍

PR #927, opened 2022-06-21
solution

huberciak:
[DEMO LINK](https://huberciak.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,115 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
-
+import { TodoList } from './components/TodoList';
+import { PreparedTodo } from './app.typedefs';
+import todos from './api/todos';
 import users from './api/users';
 
+const preparedTodos: PreparedTodo[] = todos.map(todo => ({
+  ...todo,
+  user: users.find(user => user.id === todo.userId) || undefined,
+}));
+
 const App: React.FC = () => {
+  const [todosList, setTodos] = useState(preparedTodos);
+  const [todosTitle, setTodosTitle] = useState('');
+  const [selectedUser, setSelectedUser] = useState(0);
+  const [titleError, setTitleError] = useState(false);
+  const [userError, setUserError] = useState(false);
+
+  const addTodo = (title: string, userId: number) => {
+    const todo: PreparedTodo = {
+      userId,
+      id: todosList.length + 1,
+      title,
+      completed: false,
+      user: users.find(user => user.id === userId),
+    };
+
+    setTodos([...todosList, todo]);
+  };
+
+  const handleSubmit = (event: React.FormEvent):void => {
+    event.preventDefault();
+
+    if (!todosTitle && !selectedUser) {
+      setTitleError(true);
+      setUserError(true);
+    } else if (!selectedUser) {
+      setUserError(true);
+    } else if (!todosTitle) {
+      setTitleError(true);
+    } else {
+      addTodo(todosTitle, selectedUser);
+      setTodosTitle('');
+      setSelectedUser(0);
+    }
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
+      <form
+        onSubmit={handleSubmit}
+      >
 
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
+        <div>
+          <input
+            type="text"
+            name="title"
+            data-cy="titleInput"
+            placeholder="Add task title"
+            value={todosTitle}
+            onChange={(event: React.ChangeEvent<HTMLInputElement>) => {
+              setTitleError(false);
+              setTodosTitle(event.target.value);
+            }}
+          />
+
+          {titleError && (
+            <span> Enter the title</span>
+          )}
+        </div>
+
+
+
+        <div>
+          <select
+            value={selectedUser}
+            onChange={(event) => {
+              setUserError(false);
+              setSelectedUser(Number(event.target.value));
+            }}
+          >
+            <option value="0" disabled>
+              Choose a user
+            </option>
+
+            {users.map(user => (
+              <option
+                value={user.id}
+                key={user.id}
+              >
+                {user.name}
+              </option>
+            ))}
+          </select>
+
+          {userError && (
+            <span> Choose a user</span>
+          )}
+        </div>
+
+        <div>
+          <button
+            type="submit"
+          >
+            ADD
+          </button>
+        </div>
+
+      </form>
+      <TodoList preparedTodos={todosList} />
     </div>
   );
 };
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,17 +1,115 @@
...
+      setUserError(true);
+    } else if (!selectedUser) {
+      setUserError(true);
+    } else if (!todosTitle) {
+      setTitleError(true);
+    } else {
+      addTodo(todosTitle, selectedUser);
+      setTodosTitle('');
+      setSelectedUser(0);
+    }
```
amitpatiljc, line 46: You can replace this long chain of else if statement with a function to validate the input

src/App.tsx
```diff
@@ -1,17 +1,113 @@
...
+  const handleSubmit = (event: React.FormEvent):void => {
+    event.preventDefault();
+
+    if (!todosTitle && !selectedUser) {
+      setTitleError(true);
+      setUserError(true);
+    } else if (!selectedUser) {
+      setUserError(true);
+    } else if (!todosTitle) {
+      setTitleError(true);
```
meforoma, line 41: this whole `if ... else if ... else` chain is not graceful
think of if from this perspective - how do you return a boolean whether variable is not falsy? (`!undefined = true` ; `!'string' = false`) so you could just `setTitleError(!title)` - same for the user
and you won't need the entire `if ... else if ... else` chain

src/App.tsx
```diff
@@ -1,17 +1,113 @@
...
+  const [todosList, setTodos] = useState(preparedTodos);
+  const [todosTitle, setTodosTitle] = useState('');
+  const [selectedUser, setSelectedUser] = useState(0);
+  const [titleError, setTitleError] = useState(false);
+  const [userError, setUserError] = useState(false);
+
+  const addTodo = (title: string, userId: number) => {
+    const todo: PreparedTodo = {
+      userId,
+      id: todosList.length + 1,
```
meforoma, line 23: > - `id` is a maximal `id` in the array + 1

not `list.length`
something like this:

meforoma (changes requested):

src/App.tsx
```diff
@@ -1,17 +1,112 @@
...
 import './App.css';
-
+import { TodoList } from './components/TodoList';
+import { PreparedTodo } from './app.typedefs';
+import todos from './api/todos';
 import users from './api/users';
 
+const preparedTodos: PreparedTodo[] = todos.map(todo => ({
+  ...todo,
+  user: users.find(user => user.id === todo.userId) || undefined,
```
ihor-jpeg, line 10: find method return undefined anyway if it does not find anything

src/App.tsx
```diff
@@ -1,17 +1,112 @@
...
+  const [todosList, setTodos] = useState(preparedTodos);
+  const [todosTitle, setTodosTitle] = useState('');
+  const [selectedUser, setSelectedUser] = useState(0);
+  const [titleError, setTitleError] = useState(false);
+  const [userError, setUserError] = useState(false);
+
+  const addTodo = (title: string, userId: number) => {
+    const todo: PreparedTodo = {
+      userId,
+      id: Math.max(...todosList.map((element) => element.id)) + 1,
```
ihor-jpeg, line 23: why not `todosList.length + 1`?

src/App.tsx
```diff
@@ -1,17 +1,112 @@
...
+      title,
+      completed: false,
+      user: users.find(user => user.id === userId),
```
ihor-jpeg, line 26: why do we need to store entire user's info in each todo?

src/components/UserInfo/UserInfo.tsx
```diff
@@ -0,0 +1,15 @@
...
+interface Props {
+  user: User | null,
+}
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  return (
+    <ul>
+      <li data-cy="username">{`Name: ${user?.name}`}</li>
+      <li data-cy="email">{`Email: ${user?.email}`}</li>
+    </ul>
```
ihor-jpeg, line 13: it's not really a list

src/App.tsx
```diff
@@ -1,17 +1,112 @@
...
+
+          {userError && (
+            <span> Choose a user</span>
+          )}
+        </div>
+
+        <div>
+          <button
+            type="submit"
+          >
```
ihor-jpeg, line 103: dont do it for just one attribute

ihor-jpeg (changes requested):

src/App.tsx
```diff
@@ -1,17 +1,121 @@
...
+import todos from './api/todos';
 import users from './api/users';
 
+const findUser = (userId: number):User => {
+  const selectedUser = users.find(user => user.id === userId);
+
+  const name = selectedUser?.name;
+  const email = selectedUser?.email;
+
+  return { name, email } as User;
```
ihor-jpeg, line 14: nope. what if the user actually does not exist? your app will crash. try to never use `as`

src/App.tsx
ihor-jpeg: agree. nice choice

src/components/UserInfo/UserInfo.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+interface Props {
+  user: User | null,
+}
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  return (
+    <>
+      <div data-cy="username">{`Name: ${user?.name}`}</div>
+      <div data-cy="email">{`Email: ${user?.email}`}</div>
```
ihor-jpeg, line 12: this is not the way to go. you cannot simply use ? everywhere
maybe it would be correct to not display the user at all if there is none

ihor-jpeg (changes requested):

src/App.tsx
VinnieJ-2k20: Still need to fix that. Don't use type assertion unless completely necessary. In this case it is not.

src/App.tsx
VinnieJ-2k20: +

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,24 @@
...
+  return (
+    <div className="todo" data-cy="title">
+      <h3>{`Title: ${title}`}</h3>
+      <p data-cy="status">
+        {completed
+          ? 'Completed'
+          : 'Not completed'}
+      </p>
+      User:
+      {user && <UserInfo user={user} />}
```
VinnieJ-2k20, line 21: Add some backup option, smth like "user not found"

VinnieJ-2k20 (changes requested):
Still need to fix couple of things.

src/App.tsx
huberciak: I dont get it, where is this long chain of else if?

VinnieJ-2k20 (approved):

PR #929, opened 2022-06-21
progress

Uktam19980416:


Uktam19980416:
[DEMO LINK](https://Uktam19980416.github.io/react_add-todo-form/)

amitpatiljc (approved):
Good Job!

src/App.tsx
```diff
@@ -1,17 +1,150 @@
-import React from 'react';
-import './App.css';
-
+import React, { useState } from 'react';
+// import './App.scss';
```
meforoma, line 2: no commented code please

src/App.tsx
```diff
@@ -1,17 +1,150 @@
...
+  const [todoTitle, setTitle] = useState('');
+  const [selectedUser, setSelectedUser] = useState(0);
+
+  const [hasTitleInputError, setHasTitleInputError] = useState(false);
+  const [hasUserInputError, setHasUserInputError] = useState(false);
+
+  const addtodo = (title: string, userId: number) => {
+    const newTodo: Todo = {
+      userId,
+      id: todosList.length + 1,
```
meforoma, line 29: > - `id` is a maximal `id` in the array + 1

not `list.length`
something like this:

src/App.tsx
```diff
@@ -1,17 +1,150 @@
...
+          <div className="notification">
+            <select
+              value={selectedUser}
+              onChange={(event) => {
+                setHasUserInputError(false);
+                setSelectedUser(Number(event.target.value));
+              }}
+              className={cn({ error: hasUserInputError })}
+            >
+              <option value="0" disabled>
```
meforoma, line 111: try adding `hidden` - so it's not even on the list
```suggestion
              <option value="0" hidden>
```

meforoma (changes requested):
good job 👍 
consider few comments

meforoma (approved):

PR #930, opened 2022-06-22
Add todo form task

proznerka:
[DEMO LINK](https://proznerka.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,25 @@
 import React from 'react';
 import './App.css';
 
+import todos from './api/todos';
 import users from './api/users';
+import { TodoList } from './components/TodoList';
+import { PreparedTodo } from './typedefs';
 
-const App: React.FC = () => {
-  return (
+const preparedTodos: PreparedTodo[] = todos.map(todo => {
+  return {
+    ...todo,
+    user: users.find(user => user.id === todo.userId) || null,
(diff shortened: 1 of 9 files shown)


VinnieJ-2k20 (changes requested):
Almost done, very small thing to fix with changing state.

src/App.tsx
```diff
@@ -1,17 +1,123 @@
...
+        onSubmit={addTodo}
+        className="form"
+      >
+        <div>
+          <select
+            className="form__select"
+            name=""
+            id=""
+            data-cy="userSelect"
+            placeholder="Choose a user"
```
meforoma, line 60: this does not impact `select` element
```suggestion
```

src/App.tsx
```diff
@@ -1,17 +1,123 @@
...
+            value={selectUser}
+            onChange={(changeEvent) => {
+              setSelectUser(+changeEvent.target.value);
+              setErrorUser(false);
+            }}
+          >
+            <option
+              value={0}
+              disabled
```
meforoma, line 69: you could also add `hidden` so this option is not even visible on the list
```suggestion
              disabled
              hidden
```

meforoma (approved):
👍 @proznerka 
left two minor comments that are not blocking

PR #931, opened 2022-06-23
Task complete

rHashTag:
[DEMO LINK](https://rHashTag.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,126 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
+import cn from 'classnames';
 
 import users from './api/users';
+import todos from './api/todos';
+import { TodoList } from './components/TodoList/TodoList';
+import { PreparedTodo } from './app.typedef';
+
+const preparedTodos: PreparedTodo[] = todos.map(todo => ({
+  ...todo,
+  user: users.find(user => user.id === todo.userId) || null,
+}));
 
 const App: React.FC = () => {
+  const [todosList, setTodosList] = useState([...preparedTodos]);
+  const [todoTitle, setTodoTitle] = useState('');
+  const [selectedUser, setSelectedUser] = useState(0);
+  const [titleInputError, setTitleInputError] = useState(false);
+  const [userInputError, setUserInputError] = useState(false);
+
+  const addtodo = (title: string, userId: number) => {
+    const newTodo: PreparedTodo = {
+      userId,
+      id: todosList.length + 1,
+      title,
+      completed: false,
+      user: users.find(user => userId === user.id) || null,
+    };
+
+    setTodosList([...todosList, newTodo]);
+  };
+
+  const clearForm = () => {
+    setSelectedUser(0);
+    setTodoTitle('');
+  };
+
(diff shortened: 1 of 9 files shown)


amitpatiljc (approved):
Good Job!

src/App.tsx
```diff
@@ -1,17 +1,126 @@
...
+  const [todosList, setTodosList] = useState([...preparedTodos]);
+  const [todoTitle, setTodoTitle] = useState('');
+  const [selectedUser, setSelectedUser] = useState(0);
+  const [titleInputError, setTitleInputError] = useState(false);
+  const [userInputError, setUserInputError] = useState(false);
+
+  const addtodo = (title: string, userId: number) => {
+    const newTodo: PreparedTodo = {
+      userId,
+      id: todosList.length + 1,
```
meforoma, line 25: almost everyone has this one

> - `id` is a maximal `id` in the array + 1

not `list.length`
something like this:

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+type Props = {
+  preparedTodos: PreparedTodo[];
+};
+
+export const TodoList: React.FC<Props> = ({ preparedTodos }) => {
+  return (
+    <ul className="todoList">
+      {preparedTodos.map((todo: PreparedTodo) => {
+        return (
```
meforoma, line 13: implicit returns

meforoma (changes requested):
95% 👍 
just a few tiny comments

src/App.tsx
```diff
@@ -1,17 +1,133 @@
...
+import { TodoList } from './components/TodoList/TodoList';
+import { PreparedTodo } from './app.typedef';
+
+const preparedTodos: PreparedTodo[] = todos.map(todo => ({
+  ...todo,
+  user: users.find(user => user.id === todo.userId) || null,
+}));
 
 const App: React.FC = () => {
+  const [todosList, setTodosList] = useState([...preparedTodos]);
```
ihor-jpeg, line 16: why destructure?

src/App.tsx
```diff
@@ -1,17 +1,133 @@
...
+  const [todoTitle, setTodoTitle] = useState('');
+  const [selectedUser, setSelectedUser] = useState(0);
```
ihor-jpeg, line 18: this is more like `selectedUserId`

src/App.tsx
```diff
@@ -1,17 +1,133 @@
...
+  const [titleInputError, setTitleInputError] = useState(false);
+  const [userInputError, setUserInputError] = useState(false);
+
+  const addtodo = (title: string, userId: number) => {
+    const newTodo: PreparedTodo = {
+      userId,
+      id: Math.max(...todosList.map((todo) => todo.id)) + 1,
+      title,
+      completed: false,
+      user: users.find(user => userId === user.id) || null,
```
ihor-jpeg, line 28: maybe if the user does not exist we should not add a todo? otherwise who will do it?

ihor-jpeg (approved):

PR #940, opened 2022-07-14
Develop

nastia-sydorchuk:
[DEMO LINK](https://nastia-sydorchuk.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/UserInfo/UserInfo.tsx
+++ b/src/components/UserInfo/UserInfo.tsx
@@ -1,0 +1,16 @@
+import React from 'react';
+import User from '../../types/User';
+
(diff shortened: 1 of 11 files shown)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import User from '../../types/User';
+
+export const UserInfo: React.FC<User> = ({ name, email, username }) => (
+  <>
+    <div>
```
alexandra-protyanova, line 6: use more semantic tags

PR #944, opened 2022-07-16
my develop

Pavlo000:
[DEMO LINK](https://Pavlo000.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -24,17 +24,13 @@ const App: React.FC = () => {
   };
 
   const addTodo = () => {
-    const valid = () => (
-      userSelect !== 0 && title !== ''
-    );
+    setClick(true);
 
-    setClick(!click);
-
-    if (valid()) {
+    if (title && userSelect > 0) {
       setTodos(() => {
         const newTodo = {
           userId: userSelect,
-          id: currToddos.length + 1,
+          id: currToddos[currToddos.length - 1].id + 1,
           title,
           completed: false,
           user: users.find(user => user.id === userSelect) || null,
@@ -42,6 +38,7 @@ const App: React.FC = () => {
 
         setTitle('');
         setUserSelect(0);
+        setClick(false);
 
         return [...currToddos, newTodo];
       });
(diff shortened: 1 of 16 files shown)


src/App.tsx
```diff
@@ -1,17 +1,107 @@
...
 
 const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [userSelect, setUserSelect] = useState(0);
+  const [currToddos, setTodos] = useState(preparedTodos);
+  const [click, setClick] = useState(false);
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+  };
```
AndriiNychaiuk, line 20: Nice job! We can add new todos and it's definetly a win. But there're few thing I guess you chould fix.
Firs of all I'm not realy sure you need differnt function for handling changes. You can make logic with one fucntion for both cases

src/App.tsx
```diff
@@ -1,17 +1,107 @@
...
+    setUserSelect(Number(event.target.value));
+  };
+
+  const addTodo = () => {
+    const valid = () => (
+      userSelect !== 0 && title !== ''
+    );
+
+    setClick(!click);
+
```
AndriiNychaiuk, line 32: This logic won't work the way we need. Every time when I press the button the message appears or disappears whether there is anything in the field or not. But you need the error to not appear under a non-empty field

src/App.tsx
```diff
@@ -1,17 +1,107 @@
...
+    if (valid()) {
+      setTodos(() => {
+        const newTodo = {
+          userId: userSelect,
+          id: currToddos.length + 1,
+          title,
```
AndriiNychaiuk, line 38: Yes, I had the same idea, but after thinking twice, I decided to change the way I provide the id, since it will work almost the same way as an index. The task says that you should use the highest id from the already created todos

AndriiNychaiuk (changes requested):

src/App.tsx
Pavlo000: Hi! thank for review, i corrected some mistakes, please look again

src/App.tsx
AndriiNychaiuk: It's mush better way to provide id for this task, but in general it still there's an issue. For example if your app had the reverse button it wouldn't work correct because you take id from the last object in the list, not from the object with the highes id.

AndriiNychaiuk (approved):

src/App.tsx
```diff
@@ -1,17 +1,112 @@
...
+
   return (
     <div className="App">
-      <h1>Add todo form</h1>
+      <h1 className="App__title">Static list of todos</h1>
+      <div className="App__form-field">
+        <form onSubmit={(event) => {
+          event.preventDefault();
+          addTodo();
+        }}
```
jstmpelowycz, line 55: you'd rather create a handler

jstmpelowycz (approved):

PR #979, opened 2022-07-20
Solution

yanaost:
[DEMO LINK](https://yanaost.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,1 +1,32 @@
-export const TodoInfo = () => {};
+import { User, UserInfo } from '../UserInfo/UserInfo';
+
+export interface Todo {
+  userId: number,
+  id: number,
+  title: string,
+  completed: boolean,
+  user?: User | null,
+}
+
+type Props = {
+  todo: Todo,
+};
+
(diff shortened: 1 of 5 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,32 @@
...
+
+type Props = {
+  todo: Todo,
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const { user, title, completed } = todo;
+
+  return (
+
```
jstmpelowycz, line 19:
```suggestion
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,32 @@
...
+    <article
+      data-id={todo.id}
+      className={`TodoInfo${completed ? ' TodoInfo--completed' : ''}`}
```
jstmpelowycz, line 22: use the following tool here, using ternary with empty string is not a solution
https://www.npmjs.com/package/classnames

jstmpelowycz (changes requested):

mykola-koval (approved):

PR #1003, opened 2022-08-01
Develop

IvanG6:
https://ivang6.github.io/react_add-todo-form/


valerii-derkach (changes requested):
add separate error handling for if a user was selected; if not, show an error message next to the select (Please choose a user);

do not display title error for user select
image

Bohdan-Maksymiuk (approved):

PR #1006, opened 2022-08-05
develop

fledith:
[DEMO LINK](https://fledith.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,175 @@
+import React, { useState } from 'react';
+import classNames from 'classnames';
+import { TodoList } from './components/TodoList/TodoList';
+
+import { User } from './type/User';
+
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-export const App = () => {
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,61 +1,175 @@
...
+  for (let i = 1; i < maxId[0].id; i += 1) {
+    if (todos.every(todo => todo.id !== i)) {
+      return i;
+    }
+  }
+
+  console.log(todos); // eslint-disable-line
+
+  return maxId[0].id + 1;
+}
```
BudnikOleksii, line 33: I'm not sure, but I think you don't need to check the last id every time. You can check it only once and save it into a variable, and when you create a new todo just increase this id number. I did that😅

BudnikOleksii (changes requested):
Great design, I like it, good job!
For tests, you need to show all errors at the same time if the form is empty
6) should show all errors if the form is empty
image

src/App.tsx
fledith: > I'm not sure, but I think you don't need to check the last id every time. You can check it only once and save it into a variable, and when you create a new todo just increase this id number. I did that😅

not last id, it is the id that is not in the array

ivan-kadykalo (approved):
👍

PR #1011, opened 2022-08-09
Solution

Ksenia-Didukh:
[DEMO LINK](https://Ksenia-Didukh.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,155 @@
 import './App.scss';
+import React, { Component } from 'react';
+import { TodoList } from './components/TodoList/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-export const App = () => {
-  return (
-    <div className="App">
-      <h1>Add todo form</h1>
-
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,61 +1,155 @@
...
+  emptyTitle: boolean,
+  emptyName: boolean,
 };
+
+const visibleTodos = [...todosFromServer];
+
+export class App extends Component <{}, State> {
+  state = {
+    title: '',
+    name: 'Choose a user',
```
Bohdan-Maksymiuk, line 21: it should be a placeholder

src/App.tsx
```diff
@@ -1,61 +1,155 @@
...
+    id: visibleTodos.length - 1,
```
Bohdan-Maksymiuk, line 22: > id is the largest id in the array + 1 (add data-id={todo.id} attribute to each .TodoInfo).

what if we will add the ability to delete todos?

Bohdan-Maksymiuk (changes requested):
https://user-images.githubusercontent.com/94289038/184349674-904e9915-545f-496b-a92b-5d1d698593a8.mov

form should become empty after submit

PR #1032, opened 2022-08-15
Solution

Lentyaeev:
[DEMO LINK](https://lentyaeev.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -51,8 +51,12 @@ export const App = () => {
   };
 
   return (
-    <div className="App">
-      <h1>Add todo form</h1>
+    <div className="App panel">
+      <h1
+        className="heading is-size-3"
+      >
+        Add todo form
+      </h1>
 
       <form
         action="/api/users"
@@ -62,6 +66,7 @@ export const App = () => {
         <div className="field">
           <input
             type="text"
+            className="input"
             data-cy="titleInput"
             placeholder="Enter a title"
             value={todoTitle}
@@ -73,30 +78,36 @@ export const App = () => {
           />
           {titleError && <span className="error">Please enter a title</span>}
         </div>
-
-        <div className="field">
-          <label>
-            User:
-
-            {' '}
-            <select
-              data-cy="userSelect"
-              onChange={event => {
-                setSelectedUser(event.target.value);
-                setUserError(false);
-              }}
-            >
-              <option value="-1" disabled>Choose a user</option>
-              {usersFromServer.map(user => (
-                <option value={user.id} key={user.id}>{user.name}</option>
-              ))}
-            </select>
-          </label>
+        <label
+          className="label"
+          htmlFor="select"
+        >
+          User:
+        </label>
+        <div className="select is-rounded">
+          <select
+            id="select"
+            data-cy="userSelect"
+            value={selectedUser}
+            onChange={event => {
+              setSelectedUser(event.target.value);
+              setUserError(false);
+            }}
+          >
+            <option value="-1" disabled selected>Choose a user</option>
+            {usersFromServer.map(user => (
+              <option value={user.id} key={user.id}>{user.name}</option>
+            ))}
+          </select>
 
           {userError && <span className="error">Please choose a user</span>}
         </div>
 
-        <button type="submit" data-cy="submitButton">
+        <button
+          type="submit"
+          data-cy="submitButton"
+          className="button"
+        >
           Add
         </button>
       </form>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,99 @@
 import './App.scss';
-
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { FormEvent, useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,128 @@
...
+            data-cy="titleInput"
+            placeholder="Enter a title"
+            value={todoTitle}
+            onChange={event => {
+              setTitleError(false);
+              if (event.target.value.match(/[А-яа-я]/g)) {
+                setTitleLanguageError(true);
+              } else {
+                setTitleLanguageError(false);
+              }
```
Oleksandr-Pustovit, line 80: _comment_ 
Usually when you want to use if else construction to determine one value, it can be done in a more short form.
If I understand this code right you can simplify it to something like this.
```suggestion
              setTitleLanguageError(event.target.value.match(/[А-яа-я]/g));
```

src/App.tsx
```diff
@@ -1,61 +1,128 @@
...
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
-        </div>
+          <input
+            type="text"
+            className="input"
+            data-cy="titleInput"
+            placeholder="Enter a title"
+            value={todoTitle}
+            onChange={event => {
```
Oleksandr-Pustovit, line 74: I am not sure if that was on any checklist but mentors usually mention that it is better to make event handler functions instead of writing funtion inline.
```suggestion
            onChange={event => handleInputChange(event)}
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,25 @@
...
+import users from '../../api/users';
+
+type Props = {
+  todo: Todo
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const currentUser = users.find(user => user.id === todo.userId);
+
+  return (
```
Oleksandr-Pustovit, line 11: [checklist 5](https://github.com/mate-academy/react_add-todo-form/blob/master/newChecklist.md)
There is no need to download users from server twice in one component. Better prepare all the information about todos in one component and send it to al the others as parameters. This correction is only for visualisation of my thoughts. You can create other ideas of how to send it.
```suggestion
import users from '../../api/users';

type Props = {
  todo: Todo,
  currentUser: User,
};

export const TodoInfo: React.FC<Props> = ({ todo, currentUser }) => {

  return (
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,25 @@
...
+      data-id={todo.id}
+      className={`TodoInfo${todo.completed ? ' TodoInfo--completed' : ''}`}
+    >
+      <h2 className="TodoInfo__title">
+        {todo.title}
+      </h2>
+
+      <a className="UserInfo has-text-grey" href={`mailto:${currentUser?.email}`}>
+        {currentUser?.name}
+      </a>
```
Oleksandr-Pustovit, line 22: [checklist 4](https://github.com/mate-academy/react_add-todo-form/blob/master/newChecklist.md)
You can make a check for user and prevent rendering this part at all if there is no one
```suggestion
      {currentUser &&  (
        <a className="UserInfo has-text-grey" href={`mailto:${currentUser.email}`}>
          {currentUser.name}
        </a>
      )
```

Oleksandr-Pustovit (changes requested):
Well done!
There are some comments to fix.

VETALEK (changes requested):
Found some bug 😉 
image

src/App.tsx
```diff
@@ -1,61 +1,135 @@
...
 
 export const App = () => {
+  const [selectedUser, setSelectedUser] = useState('-1');
+  const [userError, setUserError] = useState(false);
+  const [todoTitle, setTodoTitle] = useState('');
+  const [titleError, setTitleError] = useState(false);
+  const [todos, setTodos] = useState(todosFromServer);
+  const [titleLanguageError, setTitleLanguageError] = useState(false);
+
+  const titleHandler = (event: FormEvent<HTMLInputElement>) => {
```
solaryasha, line 15: Function's name should start with a verb (so you could easily tell what your method actually do)

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,28 @@
...
+  todos: Todo[],
+  users: User[],
+};
+
+export const TodoList:React.FC<Props> = ({ todos, users }) => {
+  const currentUser = (todo: Todo) => {
+    const currUser = users.find(user => user.id === todo.userId);
+
+    return currUser || null;
+  };
```
solaryasha, line 15: 5 пункт чекліста - https://github.com/mate-academy/react_add-todo-form/blob/master/newChecklist.md

solaryasha (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,148 @@
...
+    }
+
+    if (todoTitle === '') {
+      setTitleError(true);
+    }
+
+    return false;
+  };
+
+  const submitHandler = (event: FormEvent) => {
```
solaryasha, line 55:
```suggestion
  const handleSubmit = (event: FormEvent) => {
```

solaryasha (approved):

PR #1033, opened 2022-08-15
solution

DmitriyTeteryn:
[DEMO LINK](https://DmitriyTeteryn.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,182 @@
+import React, { useState } from 'react';
+import classNames from 'classnames';
+import { TodoList } from './components/TodoList/TodoList';
+
+import { User } from './type/User';
+
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-export const App = () => {
+function getUser(userId: number): User | null {
+  return usersFromServer.find(user => user.id === userId) || null;
+}
+
+const todos = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+function checkKey() {
+  const maxId = [...todos].sort((a, b) => b.id - a.id);
+
+  for (let i = 1; i < maxId[0].id; i += 1) {
+    if (todos.every(todo => todo.id !== i)) {
+      return i;
+    }
+  }
+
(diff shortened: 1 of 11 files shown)


opanasiukvlad (approved):
Looks fantastic! Good job man!

src/App.tsx
```diff
@@ -1,61 +1,182 @@
...
+function checkKey() {
+  const maxId = [...todos].sort((a, b) => b.id - a.id);
+
+  for (let i = 1; i < maxId[0].id; i += 1) {
+    if (todos.every(todo => todo.id !== i)) {
+      return i;
+    }
+  }
+
+  console.log(todos); // eslint-disable-line
```
VETALEK, line 30: you don't need it in the final code

src/App.tsx
```diff
@@ -1,61 +1,182 @@
...
+  return usersFromServer.find(user => user.id === userId) || null;
+}
+
+const todos = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+function checkKey() {
+  const maxId = [...todos].sort((a, b) => b.id - a.id);
```
VETALEK, line 22:
```suggestion
  const maxId = [...todos].sort((a, b) => b.id - a.id)[0];
```
you mate take the max id, to not keep all the array

VETALEK (changes requested):
GJ 👍 
commented

src/App.tsx
```diff
@@ -1,61 +1,180 @@
...
+const addTodo = (title: string, completed: boolean, userId: number): void => {
+  const newTodo = {
+    id: checkKey(),
+    title,
+    completed,
+    userId,
+    user: getUser(userId),
+  };
+
+  todos.push(newTodo);
```
solaryasha, line 42: Avoid storing data and having state outside component, use `useState` function

src/App.tsx
```diff
@@ -1,61 +1,180 @@
...
+      return;
+    }
+
+    if (userId === '0') {
+      setErrorUser(true);
+
+      return;
+    }
+
+    addTodo(title, completed, +userId);
```
solaryasha, line 100: It's better to pass an object, so you don't make mistake with params order

solaryasha (changes requested):

PR #1041, opened 2022-08-16
task is ready

opanasiukvlad:
[DEMO](https://opanasiukvlad.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,125 @@
 import './App.scss';
+import { FormEvent, useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList/TodoList';
+
+const copyOfUsersFromServer: User[] = [...usersFromServer];
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+const biggestTodoId = () => {
+  const ids = [...todos].map(todo => {
+    return todo.id;
+  });
+
+  return Math.max(...ids);
+};
+
+const createNewToDo = (inputTitle: string, selectedUser: string): Todo => {
+  return {
+    id: biggestTodoId() + 1,
+    title: inputTitle,
+    completed: false,
+    userId: +selectedUser,
+    user: getUser(+selectedUser),
+  };
+};
 
 export const App = () => {
+  const [inputTitle, setInputTitle] = useState('');
+  const [selectedUser, setSelectedUser] = useState('');
+  const [titleIsEmpty, setTitleIsEmpty] = useState(false);
+  const [userIsEmpty, setUserIsEmpty] = useState(false);
+
+  const handleSubmit = (event: FormEvent) => {
+    event.preventDefault();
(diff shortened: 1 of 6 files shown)


DmitriyTeteryn:
Хорошая робота, все ок!

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList/TodoList';
+
+const copyOfUsersFromServer: User[] = [...usersFromServer];
+
+function getUser(userId: number): User | null {
```
solaryasha, line 12:
```suggestion
function getUserById(userId: number): User | null {
```

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+  return foundUser || null;
+}
+
+const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+const biggestTodoId = () => {
+  const ids = [...todos].map(todo => {
```
solaryasha, line 24: If you are using a non-mutating array method, you don't need to create a copy of the array

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+    if (!selectedUser) {
+      setUserIsEmpty(true);
+    }
+
+    if (inputTitle.length <= 0 || !selectedUser) {
+      return;
+    }
+
+    if (!titleIsEmpty && !userIsEmpty) {
+      todos.push(createNewToDo(inputTitle, selectedUser));
```
solaryasha, line 62: No, let's don't mutate array and create specific state in the Component

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
+        action="/api/users"
+        method="POST"
+        onSubmit={(event) => handleSubmit(event)}
```
solaryasha, line 78:
```suggestion
        onSubmit={handleSubmit}
```

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+          <label htmlFor="titleInput">Title: </label>
+          <input
+            type="text"
+            data-cy="titleInput"
+            id="titleInput"
+            placeholder="Enter a title"
+            value={inputTitle}
+            onChange={event => {
+              setInputTitle(event.target.value);
+              setTitleIsEmpty(false);
```
solaryasha, line 90: Let's call this function only when `titleIsEmpty` equals true

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+            && <span className="error">Please enter a title</span>}
         </div>
 
         <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>Choose a user</option>
+          <select
+            onChange={(event) => {
+              setSelectedUser(event.target.value);
+              setUserIsEmpty(false);
```
solaryasha, line 102: Let's call this function only when userIsEmpty equals true

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+            }}
+            data-cy="userSelect"
+            value={selectedUser}
+          >
+            <option value="" disabled>Choose a user</option>
+
+            {copyOfUsersFromServer.map((user) => (
```
solaryasha, line 109: If you are using a non-mutating array method, you don't need to create a copy of the array

solaryasha (changes requested):

opanasiukvlad:
Thank you, all problems are fixed.

solaryasha (approved):

PR #1053, opened 2022-08-17
Develop

oleg-zolkiewskyi:


src/App.tsx
```diff
@@ -1,60 +1,128 @@
...
+
+  const onSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    const user = usersFromServer
+      .find(curUser => curUser.name === users) || null;
+
+    const newTodo = {
+      userId: user ? user.id : 0,
+      id: todos.length + 1,
```
Bohdan-Maksymiuk, line 36: > id is the largest id in the array + 1

src/App.tsx
```diff
@@ -1,60 +1,128 @@
...
+  const [userError, setUserError] = useState(false);
+
+  const onSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    const user = usersFromServer
+      .find(curUser => curUser.name === users) || null;
+
+    const newTodo = {
+      userId: user ? user.id : 0,
```
Bohdan-Maksymiuk, line 35: why default userId is 0?
do we need to create Todo id we don't have user?

Bohdan-Maksymiuk (changes requested):
check your demo link - it doesn't work

PR #1054, opened 2022-08-20
add-todo-info

ghost:
[DEMO LINK](https://Vitaliy-Anochin.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,71 +1,66 @@
-import React, { useState } from 'react';
+import { FormEvent, useState } from 'react';
 import './App.scss';
 
 import { TodoList } from './components/TodoList/TodoList';
-import { PreparedTodo } from './components/types/preparedTodo';
 
 import usersFromServer from './api/users';
 import todosFromServer from './api/todos';
+import { Todo } from './types/Todo';
 
-const preparedTodos: PreparedTodo[] = todosFromServer.map(todo => {
-  return {
-    ...todo,
-    user: usersFromServer.find(user => user.id === todo.userId),
-  };
-});
+function getUser(usrId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === usrId);
 
-let maxTodoId = [...todosFromServer].sort((el1, el2) => (
-  el2.id - el1.id))[0].id;
+  return foundUser || null;
+}
 
-const addTodo = (title: string, userName: string) => {
-  maxTodoId += 1;
-
-  const newTodo = {
-    id: maxTodoId,
-    title,
-    completed: false,
-    userId: 1,
-    user: usersFromServer.find(user => user.name === userName),
-  };
-
-  preparedTodos.push(newTodo);
-};
+const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+
+const preparedTodos: PreparedTodo[] = todosFromServer.map(todo => {
+  return {
+    ...todo,
+    user: usersFromServer.find(user => user.id === todo.userId),
+  };
+});
+
+let maxTodoId = [...todosFromServer].sort((el1, el2) => (
+  el2.id - el1.id))[0].id;
```
Bohdan-Maksymiuk, line 18: move this to addTodo callback
and better to map => getArray of all id's => use `Math.max()`

src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+  const handleChange = (
+    event: React.ChangeEvent<HTMLInputElement | HTMLSelectElement>,
+  ) => {
+    const { name, value } = event.target;
+
+    if (name === 'title') {
+      setTitle(value);
+    } else {
+      setUser(value);
+    }
```
Bohdan-Maksymiuk, line 49: what if we will add more inputs?
it will be more `if else`?
better to use switch

Bohdan-Maksymiuk (changes requested):
https://user-images.githubusercontent.com/94289038/186249077-c66edc8b-b673-4b4c-b899-bef9676d1a48.mov

* error message should disappear after putting valid input and adding todo

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [todoList, setTodoList] = useState(todos);
+  const [userId, setUserId] = useState(0);
+  const [title, setTitle] = useState('');
+  const [isTitle, setIsTitle] = useState(false);
+  const [isUser, setIsUser] = useState(false);
```
SanyaBratashchuk, line 26: isTitle, isUser???
If it is about errors use hasTitleError

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+
+  const maxId = () => {
+    return Math.max(...todoList.map(todo => todo.id)) + 1;
+  };
```
SanyaBratashchuk, line 30:
```suggestion
  const getMaxId = () => {
    return Math.max(...todoList.map(todo => todo.id)) + 1;
  };
```

src/react-app-env.d.ts
```diff
@@ -1 +1,14 @@
-/// <reference types="react-scripts" />
+interface Todo {
+  id: number,
+  title: string,
+  completed: boolean,
+  userId: number,
+  user?: User,
+}
```
SanyaBratashchuk, line 7:
```suggestion
interface Todo {
  id: number,
  title: string,
  completed: boolean,
  userId: number,
}
interface ExtendedTodo extednds Todo {
 user: User
}
```

SanyaBratashchuk (changes requested):

alexandra-protyanova (approved):
GJ!

PR #1068, opened 2022-09-08
solution

kostya-konev:
(https://kostya-konev.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,6 +15,7 @@ const App: React.FC = () => {
   const [newTask, setNewTask] = useState('');
   const [userFromForm, setUserFromForm] = useState('Leanne Graham');
   const [allTodos, setAllTodos] = useState(preparedTodos);
+  const [isEmpty, setIsEmpty] = useState(false);
 
   return (
     <div className="App">
@@ -35,8 +36,13 @@ const App: React.FC = () => {
             completed: false,
             user: neededUser,
           };
-          
-          setAllTodos(prev => [...prev, todoToAdd]);
+
+          if (todoToAdd.title) {
+            setAllTodos(prev => [...prev, todoToAdd]);
+            setIsEmpty(false);
+          } else {
+            setIsEmpty(true);
+          }
         }}
       >
         <input
@@ -66,6 +72,10 @@ const App: React.FC = () => {
           Add
         </button>
       </form>
+
+      {isEmpty && (
+        <div className='App__message'>Title is empty. Please, add some task</div>
+      )}
 
       <TodoList todos={allTodos}/>
     </div>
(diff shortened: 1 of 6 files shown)


alexandra-protyanova (changes requested):
Todo shouldn't be added if is empty and error message should be displayed

src/App.tsx
```diff
@@ -1,17 +1,83 @@
...
+import todos from './api/todos';
+
+const preparedTodos = todos.map(todo => ({
+  ...todo,
+  user: users.find(user => user.id === todo.userId) || null,
+}));
 
 const App: React.FC = () => {
+  const [newTask, setNewTask] = useState('');
+  const [userFromForm, setUserFromForm] = useState('Leanne Graham');
```
SanyaBratashchuk, line 16: hold userId here

SanyaBratashchuk, line 16:
```suggestion
  const [selectedUser, setSelectedUser] = useState('Leanne Graham');
```

src/App.tsx
```diff
@@ -1,17 +1,83 @@
...
 import users from './api/users';
+import todos from './api/todos';
+
+const preparedTodos = todos.map(todo => ({
+  ...todo,
+  user: users.find(user => user.id === todo.userId) || null,
+}));
 
 const App: React.FC = () => {
+  const [newTask, setNewTask] = useState('');
```
SanyaBratashchuk, line 15:
```suggestion
  const [title, setTitle] = useState('');
```

src/App.tsx
```diff
@@ -1,17 +1,83 @@
...
+          const todoToAdd = {
+            userId: neededUser.id,
+            id: allTodos[allTodos.length - 1].id + 1,
+            title: newTask,
+            completed: false,
+            user: neededUser,
+          };
+
+          if (todoToAdd.title) {
+            setAllTodos(prev => [...prev, todoToAdd]);
```
SanyaBratashchuk, line 41: move it to the separate function

src/App.tsx
```diff
@@ -1,17 +1,83 @@
...
+const preparedTodos = todos.map(todo => ({
+  ...todo,
+  user: users.find(user => user.id === todo.userId) || null,
+}));
 
 const App: React.FC = () => {
+  const [newTask, setNewTask] = useState('');
+  const [userFromForm, setUserFromForm] = useState('Leanne Graham');
+  const [allTodos, setAllTodos] = useState(preparedTodos);
+  const [isEmpty, setIsEmpty] = useState(false);
```
SanyaBratashchuk, line 18: You don't need this state, you already have state with title so it can be just a variable, `const isEmpty = title.length === 0`

SanyaBratashchuk (changes requested):

PR #1070, opened 2022-09-09
Solution

Dmytro-Matsehora:
[DEMO LINK](https://Dmytro-Matsehora.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,80 @@
+import { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { Todo, User } from './types';
+import { TodoList } from './components/TodoList/TodoList';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+const visibleTodos = [...todos];
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [option, setOption] = useState('0');
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
+        action="/api/users"
+        method="POST"
+        onSubmit={(event) => event.preventDefault()}
+      >
         <div className="field">
-          <input type="text" data-cy="titleInput" />
+          <label>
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,117 @@
...
+      user: getUser(userId),
+    };
+
+    visibleTodos.push(newTodo);
+
+    setTitle('');
+    setUserId(0);
+
+    setErrorUser(false);
+    setErrorTittle(false);
```
tiserett, line 55:
```suggestion
    setErrorUser(false);
    setErrorTittle(false);
```
I feel like your code gonna work even without these 2 lines of code, because you already handled cases if one of them is invalid

tiserett (approved):
good job ;)

src/App.tsx
```diff
@@ -1,61 +1,114 @@
...
+
+            <input
+              type="text"
+              data-cy="titleInput"
+              placeholder="Enter a title"
+              value={title}
+              onChange={(event) => {
+                setTitle(event.target.value);
+                setErrorTittle(false);
+              }}
```
dellachaise, line 76: better create a function and use it here.

src/App.tsx
```diff
@@ -1,61 +1,114 @@
...
-            <option value="0" disabled>Choose a user</option>
-          </select>
+          <label>
+            <span className="inputTitle">User: </span>
+            <select
+              data-cy="userSelect"
+              value={userId}
+              onChange={(event) => {
+                setUserId(+event.target.value);
+                setErrorUser(false);
```
dellachaise, line 90: here the same.

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,32 @@
...
+  } = todo;
+
+  return (
+    <article
+      data-id={id}
+      className={
+        completed
+          ? 'TodoInfo TodoInfo--completed'
+          : 'TodoInfo'
+      }
```
dellachaise, line 23: Better install [classnames](https://www.npmjs.com/package/classnames) and use it instead of ternary operator.
The ternary operator is very limited in its functionality to be used in classNames. (this way usually did in real projects and would be great to work with it now and would be needed to do in test tasks).

dellachaise (changes requested):
Great job 👍 

Just add changes according to the comments.

Optional: 
The react tasks you're currently working on are great to try out some css framework and see how it works.
for example, to use [bulma](https://bulma.io/), just run the command
`npm install bulma`
import bulma in the main index.js file `index.tsx`
`import 'bulma/css/bulma.min.css';`
and all :heavy_check_mark:
you can apply the necessary classes to the components and see how it looks :tada:

dellachaise (approved):
Well done 👍

PR #1075, opened 2022-09-15
develop

Blankevich:
https://Blankevich.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,121 @@
 import './App.scss';
+import React, { useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  // if there is no user with a given userId
+  return foundUser || null;
+}
+
+const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+let timeId = Math.max(...todos.map(todo => todo.id));
+
+const newTodo = (title: string, userId: number): Todo => ({
+  id: timeId,
+  userId,
+  title,
+  completed: false,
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,61 +1,121 @@
...
+};
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [user, setUser] = useState('');
+  const [errosUser, setErrorUser] = useState(false);
+  const [errorTitle, setErrorTitle] = useState(false);
+
+  const changeUser = (event:
+  { target: { value: React.SetStateAction<string>; }; }) => {
```
SanyaBratashchuk, line 45: Add correct types

src/App.tsx
```diff
@@ -1,61 +1,121 @@
...
+    setErrorUser(false);
+    setUser(event.target.value);
+  };
+
+  const changeTitle = (event:
+  { target: { value: React.SetStateAction<string>; }; }) => {
```
SanyaBratashchuk, line 51: here too

SanyaBratashchuk (changes requested):
Improve validation, because now I can create a todo with title which contain only spaces
image

src/App.tsx
```diff
@@ -1,61 +1,121 @@
...
+
+const newTodo = (title: string, userId: number): Todo => ({
+  id: timeId,
+  userId,
+  title,
+  completed: false,
+  user: getUser(userId),
+});
+
+const addTodo = (title: string, id: number) => {
```
TarasHoliuk, line 31:
```suggestion
const addTodo = (title: string, userId: number) => {
```

src/App.tsx
```diff
@@ -1,61 +1,121 @@
...
+  timeId += 1;
+  const todo = newTodo(title, id);
+
+  todos.push(todo);
+};
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [user, setUser] = useState('');
```
TarasHoliuk, line 40: since you set only id, better name it `userId, setUserId`

TarasHoliuk (changes requested):
GJ! But I've added a few names suggestions

src/App.tsx
Blankevich: changed

SanyaBratashchuk (approved):

PR #1082, opened 2022-09-18
react_add-todo-form

yurii-kozlov:
Demo link - https://yurii-kozlov.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,128 @@
+import { useState } from 'react';
 import './App.scss';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
 
-export const App = () => {
+  return foundUser || null;
+}
+
+export const todos: Todo[] = todosFromServer.map((todo) => ({
+  ...todo,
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,128 @@
...
+}
+
+export const todos: Todo[] = todosFromServer.map((todo) => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [titleError, setTitleError] = useState(false);
```
SanyaBratashchuk, line 22:
```suggestion
  const [hasTitleError, setHasTitleError] = useState(false);
```

src/App.tsx
```diff
@@ -1,61 +1,128 @@
...
+  const [userId, setUserId] = useState(0);
+  const [userError, setUserError] = useState(false);
```
SanyaBratashchuk, line 24:
```suggestion
  const [hasUserError, setHasUserError] = useState(false);
```

src/App.tsx
```diff
@@ -1,61 +1,128 @@
...
+
+  const newTodos = () => {
+    const newTodo = {
+      id: [...todos.sort((a, b) => b.id - a.id)][0].id + 1,
+      userId: +userId,
```
SanyaBratashchuk, line 29:
```suggestion
      userId: userId,
```

SanyaBratashchuk (changes requested):

src/App.tsx
yurii-kozlov: Typescript doesn't allow to use just userId as it has to be number

SanyaBratashchuk (approved):

PR #1085, opened 2022-09-19
First commit

dm-sheremetiev:
[DEMO LINK](https://dm-sheremetiev.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,133 @@
 import './App.scss';
+import { ChangeEvent, useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { UsersList } from './components/UsersList/UsersList';
+import { TodoList } from './components/TodoList';
+import { User } from './types/user';
+import { Todo } from './types/todo';
+
+function getUser(userId: number): User | undefined {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || undefined;
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [selectedUser, setSelectedUser] = useState('0');
+  const [title, setTitle] = useState('');
+  const [isClickedUser, setIsClickedUser] = useState(false);
+  const [isEmptyTitle, setIsEmptyTitle] = useState(false);
+  const isSomeFieldEmpty = !title || selectedUser === '0';
+
+  // eslint-disable-next-line max-len
+  const pattern = 'абвгдеёжзийклмнопрстуфхцчшщъыьэюяabcdefghijklmnopqrstuvwxyz1234567890 ';
+
+  const handleSelectChange = (event: ChangeEvent<HTMLSelectElement>) => {
+    setSelectedUser(event.target.value);
+  };
+
+  const handleTitleChange = (event: ChangeEvent<HTMLInputElement>) => {
+    const targetValueArr = event.target.value.split('');
+    const lastSymbol = targetValueArr[targetValueArr.length - 1];
+
+    if (pattern.includes(lastSymbol.toLowerCase())) {
+      setTitle(event.target.value);
+    } else {
+      setTitle(targetValueArr.slice(0, targetValueArr.length - 1).join(''));
+    }
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    const sortCopyTodo = [...todos].sort((todo1, todo2) => todo1.id - todo2.id);
+    const maxId = sortCopyTodo[sortCopyTodo.length - 1].id + 1;
+    const targetUser = usersFromServer.find(user => user.name === selectedUser);
+
+    event.preventDefault();
+
+    if (isSomeFieldEmpty) {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+import { ChangeEvent, useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { UsersList } from './components/UsersList/UsersList';
+import { TodoList } from './components/TodoList';
+import { User } from './types/user';
```
TarasHoliuk, line 9: from where you're doing this import? you didn't even create this file
please, fix it

TarasHoliuk: and all the same issues

src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
 
 export const App = () => {
+  const [selectedUser, setSelectedUser] = useState('0');
+  const [title, setTitle] = useState('');
+  const [isClickedUser, setIsClickedUser] = useState(false);
+  const [isEmptyTitle, setIsEmptyTitle] = useState(false);
+  const isSomeFieldEmpty = !title || selectedUser === '0';
+
+  // eslint-disable-next-line max-len
+  const pattern = 'абвгдеёжзийклмнопрстуфхцчшщъыьэюяabcdefghijklmnopqrstuvwxyz1234567890 ';
```
TarasHoliuk, line 31: better use regexp here:
[doc](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/RegExp)
[usage](https://www.freecodecamp.org/news/javascript-string-replace-example-with-regex/)

src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+
+import { UsersList } from './components/UsersList/UsersList';
+import { TodoList } from './components/TodoList';
+import { User } from './types/user';
+import { Todo } from './types/todo';
+
+function getUser(userId: number): User | undefined {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || undefined;
```
TarasHoliuk, line 15: `|| undefined` is redundant here. found user already is user or `undefined`

src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+
+  const handleSelectChange = (event: ChangeEvent<HTMLSelectElement>) => {
+    setSelectedUser(event.target.value);
+  };
+
+  const handleTitleChange = (event: ChangeEvent<HTMLInputElement>) => {
+    const targetValueArr = event.target.value.split('');
+    const lastSymbol = targetValueArr[targetValueArr.length - 1];
+
+    if (pattern.includes(lastSymbol.toLowerCase())) {
```
TarasHoliuk, line 41: why do you check only the last symbol? + you don't need to create an array to get it. `targetValueArr[targetValueArr.length - 1]` - the same logic works with string

TarasHoliuk: please, do validation in another way

TarasHoliuk (changes requested):
Changes are needed (check comments)

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+      }
+
+      if (title === '') {
+        setIsEmptyTitle(true);
+      }
+
+      return;
+    }
+
+    if (targetUser !== undefined) {
```
SanyaBratashchuk, line 66:
```suggestion
    if (targetUser) {
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,28 @@
...
+type Props = {
+  todo: Todo,
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  return (
+    <article className={classNames(
+      'TodoInfo',
+      {
+        'TodoInfo--completed': todo.completed === true,
```
SanyaBratashchuk, line 17:
```suggestion
        'TodoInfo--completed': todo.completed,
```

SanyaBratashchuk (changes requested):
Improve validation because now I can create todo with title which contains only spaces
image

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+  const isSomeFieldEmpty = !title || selectedUser === '0';
+
+  const pattern = new RegExp('[а-яА-ЯёЁA-Za-z0-9 ]');
+
+  const handleSelectChange = (event: ChangeEvent<HTMLSelectElement>) => {
+    setSelectedUser(event.target.value);
+  };
+
+  const handleTitleChange = (event: ChangeEvent<HTMLInputElement>) => {
+    const targetValue = event.target.value;
```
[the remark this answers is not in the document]
dm-sheremetiev: i can't do this because it wants to be destructured. Maybe it is a reserved name for variable. I'm using another one.

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,16 @@
...
+  todosList: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todosList }) => (
+  <section className="TodoList">
+    {todosList.map(todo => (
+      <TodoInfo todo={todo} key={todo.id} />
+    ))}
+  </section>
+);
```
marina-tilniak, line 16: use semantic tag for list

marina-tilniak (approved):

PR #1095, opened 2022-09-21
AddTodoForm solution

NataliyaAn:
[DEMO LINK](https://NataliyaAn.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,114 @@
+import React, { useState } from 'react';
 import './App.scss';
+import { TodoList } from './components/TodoList';
+import { TodoWithPerson, Person } from './type';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-export const App = () => {
+function getUser(userId: number): Person | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+const todos: TodoWithPerson[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+export const App:React.FC = () => {
+  const [todosList] = useState([...todos]);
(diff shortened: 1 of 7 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,30 @@
...
+// import { stringify } from "querystring";
+
+type Props = {
+  todo: TodoWithPerson;
+};
+
+export const TodoInfo:React.FC< Props > = ({
+  todo,
+}) => {
+  return (
```
max-kravchenko, line 15: I guess props destructuring would be a good idea

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,17 @@
-export const UserInfo = () => {};
+import React from 'react';
+import { Person } from '../../type';
+
+type Props = {
+  person: Person,
+};
+
+export const UserInfo: React.FC<Props> = ({ person }) => {
+  return (
```
max-kravchenko, line 9: I guess props destructuring would be a good idea here as well
p.s. And I suppose some tests are failing because of naming - person instead of user.

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+export const App:React.FC = () => {
+  const [todosList] = useState([...todos]);
+  const [addTodoTitle, setTodoTitle] = useState('');
+  const [personID, setPerson] = useState(0);
+  const [setNotValidPerson, setErrorUser] = useState(false);
+  const [errorTittle, setErrorTittle] = useState(false);
```
max-kravchenko, line 25: addTodoTitle and setNotValidPerson were a bit confusing for me because they sounded like action verbs and not the names of variables. I would rename them for smth like NewTodoTitle and invalidUser... smth like that

max-kravchenko (approved):
Great job!!

p.s. Only minor comments below

AlinaMatuschak (approved):
just small issues 👍

PR #1104, opened 2022-09-22
Make React Todo Form with ability to add todos

max-kravchenko:
[DEMO LINK](https://max-kravchenko.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -29,8 +29,9 @@ export const App = () => {
   const [titleError, setTitleError] = useState(false);
   const [userError, setUserIdError] = useState(false);
 
-  const handleTitleChange
-  = (event: { target: { value: SetStateAction<string>; }; }) => {
+  const handleTitleChange = (
+    event: { target: { value: SetStateAction<string>; }; },
+  ) => {
     setTitle(event.target.value);
     if (event.target.value.toString().match(/^[\sа-яіїєa-z0-9]+$/i)) {
       setTitleError(false);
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,61 +1,147 @@
...
+                setUserIdError(false);
+                setselectedUserId(Number(event.target.value));
+              }}
+            >
+              <option value="0" disabled>
+                Choose a user
+              </option>
+              {usersFromServer.map((user) => (
+                <option
+                  value={user.id}
```
NataliyaAn, line 123: you forgot to add "key"

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,29 @@
...
+      className={classNames('TodoInfo',
+        { 'TodoInfo--completed': completed })}
+    >
+      <h2 className="TodoInfo__title">
+        {title}
+      </h2>
+
+      {user
+        ? <UserInfo user={user} />
+        : 'Userus Anonymus'}
```
NataliyaAn, line 26: Good idea. But we will never see the message 'Userus Anonymus', we will see the error message

NataliyaAn (approved):
Good job!
you can make small fixes

src/App.tsx
```diff
@@ -1,61 +1,147 @@
...
+          }
+
+          if (!selectedUserId) {
+            setUserIdError(true);
+          }
+
+          if (!newTitle || !newTitle.match(/^[\sа-яіїєa-z0-9]+$/i)
+            || onlySpaces(newTitle) || !selectedUserId) {
+            return;
+          }
```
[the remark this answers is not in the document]
max-kravchenko: Thanks for the feedback!!

src/App.tsx
max-kravchenko: thanks! I thought that value as user.id will work as key. But I will add key)

AlinaMatuschak (approved):

PR #1143, opened 2022-10-02
Done

Oleksii-Blyzniuk:
[](https://Oleksii-Blyzniuk.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,104 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import React, { useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { User } from './components/types/User';
+import { Todo } from './components/types/Todo';
+import { TodoList } from './components/TodoList';
+
+function getUser(userId: number): User | null {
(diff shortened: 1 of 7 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,33 @@
...
+      className={classNames(
+        'TodoInfo', { 'TodoInfo--completed': completed },
+      )}
+      data-id={id}
+    >
+      <h2 className="TodoInfo__title">
+        {title}
+      </h2>
+
+      { user && <UserInfo user={user} />}
```
liliia-livandovskaya, line 29:
```suggestion
      {user && <UserInfo user={user} />}
```

src/App.tsx
```diff
@@ -1,25 +1,104 @@
...
+      setTitleError(true);
+    }
+
+    if (!userId) {
+      setUserError(true);
+    }
+
+    if (titleError || userError) {
+      return;
+    }
```
liliia-livandovskaya, line 54:
```suggestion
    if (!title.length) {
      setTitleError(true);
      return;
    }

    if (!userId) {
      setUserError(true);
      return;
    }
```

liliia-livandovskaya (approved):
Good job 🔥

PR #1152, opened 2022-10-10
Dev

fatihsamur:
[DEMO LINK](https://fatihsamur.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -41,7 +41,7 @@ export const App = () => {
     return true;
   };
 
-  const onsubmitHandler = (event: FormEvent<HTMLFormElement>) => {
+  const onSubmitHandler = (event: FormEvent<HTMLFormElement>) => {
     event.preventDefault();
     const userId = selectedUser;
     const title = processTitle(todoTitle);


kamal-tufekcic (changes requested):
Everything looks okay, but checklist # 3 is missing

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { User } from './types/User';
+
+function getUser(userId: number): User | null {
```
VinnieJ-2k20, line 11: in basic naming convention `getUser` means that error should be thrown if user is not found.
In this case it's better to use `findUser`.
```suggestion
function findUser(userId: number): User | null {
```

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+  const processTitle = (str: string) => {
+    return str.replace(/[^a-zA-Za-åa-ö-w-я 0-9]/gi, '');
+  };
+
+  const isValidInput = (title: string, userId: number):boolean => {
+    if (title.trim() === '') {
+      setIsTitleEmpty(true);
+    }
+
+    if (userId === 0) {
```
VinnieJ-2k20, line 33: Same.

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+    }
+
+    if (title.trim() === '' || userId === 0) {
+      return false;
+    }
+
+    return true;
+  };
+
+  const onsubmitHandler = (event: FormEvent<HTMLFormElement>) => {
```
VinnieJ-2k20, line 44:
```suggestion
  const onSubmitHandler = (event: FormEvent<HTMLFormElement>) => {
```

VinnieJ-2k20 (changes requested):
Looks good. But code can be improved.

PR #1154, opened 2022-10-12
solution

ihorandrianov:
[DEMO LINK](https://ihorandrianov.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/UserInfo/UserInfo.tsx
+++ b/src/components/UserInfo/UserInfo.tsx
@@ -1,1 +1,18 @@
-export const UserInfo = () => {};
+import { FC } from 'react';
+import users from '../../api/users';
+import { User } from '../../react-app-env';
+
+interface Props {
+  userId: number;
+}
+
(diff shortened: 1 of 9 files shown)


Denchepornyuk:
Screenshot 2022-10-25 at 17 09 21
the error should disappear when I write some text, the same for user selector

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,18 @@
...
+}
+
+export const UserInfo: FC<Props> = ({ userId }) => {
+  const user: User | undefined = users
+    .find((oneUser: User) => oneUser.id === userId);
+
+  return (
+    <a className="UserInfo link" href={`mailto:${user?.email}`}>
+      {user?.name}
+    </a>
```
Denchepornyuk, line 16: It's better not to render `a` if user doesn't exist

Denchepornyuk (approved):

PR #1157, opened 2022-10-19
add todo form is done

illia-sosnytskyi:
[DEMO LINK](https://illia-sosnytskyi.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,95 @@
 import './App.scss';
+import { useState } from 'react';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { Todo } from './react-app-env';
 
 export const App = () => {
+  const [todos, setTodos] = useState(todosFromServer);
+  const [title, setTitle] = useState('');
+  const [selectUser, setSelectUser] = useState(0);
+  const [submit, setSubmit] = useState(false);
+
+  const largestId = Math.max(...todos.map(todo => todo.id));
+
+  const addTodo = () => {
+    const newTodo: Todo = {
+      id: largestId + 1,
+      title,
+      completed: false,
+      userId: +selectUser,
+    };
+
+    setTodos([...todos, newTodo]);
+  };
+
+  const handleSubmit = (e: React.MouseEvent<HTMLFormElement>) => {
+    e.preventDefault();
+    setSubmit(true);
+
+    if (title && selectUser) {
+      addTodo();
+      setSubmit(false);
+      setTitle('');
+      setSelectUser(0);
+    }
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
+        action="/api/users"
+        method="POST"
+        onSubmit={handleSubmit}
+      >
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <label htmlFor="title">
+            Title:
+          </label>
+
+          <input
+            id="title"
+            type="text"
+            data-cy="titleInput"
+            placeholder="Enter a title"
+            value={title}
+            onChange={(e) => setTitle(e.currentTarget.value)}
+          />
+          {
+            submit && !title
+            && (<span className="error">Please enter a title</span>)
+          }
         </div>
 
         <div className="field">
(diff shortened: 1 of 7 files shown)


src/react-app-env.d.ts
```diff
@@ -1 +1,19 @@
...
+export interface User {
+  id: number;
+  name: string;
+  username: string;
+  email: string;
+}
+
+export interface FullTodoInfo extends Todo {
+  user?: User;
+}
```
sergii-nosachenko, line 19:
```suggestion
```
Move types to separate files

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,37 @@
...
+    return todosArr.map(todo => ({
+      ...todo,
+      user: users.find(user => user.id === todo.userId),
+    }));
+  };
+
+  const preparedTodos = prepareTodos(
+    usersFromServer,
+    todos,
+  );
```
sergii-nosachenko, line 26: You can do it once in parent component. It's not good idea to call api  twice every time you render the list (do it only in parent)

src/App.tsx
```diff
@@ -1,61 +1,103 @@
...
+          <label htmlFor="selectUser">
+            User:
+          </label>
+
+          <select
+            id="selectUser"
+            data-cy="userSelect"
+            value={selectUser}
+            onChange={(e) => setSelectUser(+e.currentTarget.value)}
+          >
```
sergii-nosachenko, line 79:
```suggestion
            onChange={handleUserSelect}
          >
```
Create handler

src/App.tsx
```diff
@@ -1,61 +1,103 @@
...
+            Title:
+          </label>
+
+          <input
+            id="title"
+            type="text"
+            data-cy="titleInput"
+            placeholder="Enter a title"
+            value={title}
+            onChange={(e) => setTitle(e.currentTarget.value)}
```
sergii-nosachenko, line 61: Create handler

src/App.tsx
```diff
@@ -1,61 +1,103 @@
...
+
+import { Todo } from './react-app-env';
 
 export const App = () => {
+  const [todos, setTodos] = useState(todosFromServer);
+  const [title, setTitle] = useState('');
+  const [selectUser, setSelectUser] = useState(0);
+  const [submit, setSubmit] = useState(false);
+
+  const largestId = Math.max(...todos.map(todo => todo.id));
```
sergii-nosachenko, line 16: Calculate it only when you need it (when submitting the form before create new todo)

src/App.tsx
```diff
@@ -1,61 +1,103 @@
...
+  const handleSubmit = (e: React.MouseEvent<HTMLFormElement>) => {
+    e.preventDefault();
+    setSubmit(true);
+
+    if (title && selectUser) {
+      addTodo();
+      setSubmit(false);
+      setTitle('');
+      setSelectUser(0);
+    }
```
sergii-nosachenko, line 38: 1. Don't call change of same state twice in one function call.
2. Revert logic - if something wrong - return. In other case proceed with adding new todo.
3. You need to check if title is string totally with spaces (blank string, not empty)

sergii-nosachenko (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,128 @@
...
+          <input
+            id="title"
+            type="text"
+            data-cy="titleInput"
+            placeholder="Enter a title"
+            value={title}
+            onChange={setTitleHandle}
+          />
+          {
+            hasTitleError && !title
```
sergii-nosachenko, line 89:
```suggestion
            hasTitleError
```
Don't double check the same

src/App.tsx
```diff
@@ -1,61 +1,128 @@
...
+            id="selectUser"
+            data-cy="userSelect"
+            value={selectUser}
+            onChange={setSelectHandle}
+          >
             <option value="0" disabled>Choose a user</option>
-          </select>
 
-          <span className="error">Please choose a user</span>
+            {usersFromServer.map(user => (
```
sergii-nosachenko, line 107: Can be destructurized

src/App.tsx
```diff
@@ -1,61 +1,128 @@
...
+              <option key={user.id} value={user.id}>
+                {user.name}
+              </option>
+            ))}
+
+          </select>
+          {
+            hasUserError && !selectUser
```
sergii-nosachenko, line 115: Don't double check the same

sergii-nosachenko (approved):
Good work 👍
Keep an eye on my comments

PR #1161, opened 2022-10-25
solution was added

Maksym845:
- [DEMO LINK](https://Maksym845.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -30,21 +30,10 @@ export const App = () => {
 
   const handleSubmit = (event: FormEvent<HTMLFormElement>) => {
     event.preventDefault();
-    if (inputText.trim() === '' && selectedUser === '0') {
-      setSubmittedTitle(false);
-      setSubmittedUser(false);
 
-      return;
-    }
-
-    if (inputText.trim() === '') {
-      setSubmittedTitle(false);
-
-      return;
-    }
-
-    if (selectedUser === '0') {
-      setSubmittedUser(false);
+    if (inputText.trim() === '' || selectedUser === '0') {
+      setSubmittedTitle(inputText.trim() !== '');
+      setSubmittedUser(selectedUser !== '0');
 
       return;
     }
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,136 @@
 import './App.scss';
+import { useState } from 'react';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { TodoWithUser } from './types';
+
+function todoWithUsers(): TodoWithUser[] {
+  return todosFromServer.map(todo => {
+    const foundUser = usersFromServer.find(user => user.id === todo.userId);
+
+    return {
+      ...todo,
+      user: foundUser,
+    };
+  });
+}
 
 export const App = () => {
+  const [users] = useState(usersFromServer);
+  const [todos, setTodos] = useState(todoWithUsers());
+
+  const [inputText, setInputText] = useState('');
+  const [submittedTitle, setSubmittedTitle] = useState(true);
+
+  const [selectedUser, setSelectedUser] = useState('0');
+  const [submittedUser, setSubmittedUser] = useState(true);
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
       <form action="/api/users" method="POST">
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <label>
+            {'Title: '}
+            <input
+              type="text"
+              data-cy="titleInput"
+              placeholder="Enter a title"
+              value={inputText}
+              onChange={(event) => {
+                setSubmittedTitle(true);
+                const char = event.target.value;
+
+                setInputText(char.replace(/[^\p{L} ' ']/gu, ''));
+              }}
+            />
+          </label>
+          {!submittedTitle
+            && (<span className="error">Please enter a title</span>)}
         </div>
 
         <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>Choose a user</option>
-          </select>
+          <label>
+            {'User: '}
+            <select
+              data-cy="userSelect"
+              value={selectedUser}
+              onChange={(event) => {
+                setSubmittedUser(true);
+                setSelectedUser(event.target.value);
+              }}
+            >
+              <option value="0">Choose a user</option>
+              {users.map(user => {
+                const { id, name } = user;
 
-          <span className="error">Please choose a user</span>
+                return (
+                  <option key={id} value={id}>
+                    {name}
+                  </option>
+                );
+              })}
+            </select>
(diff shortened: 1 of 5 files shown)


Yeliseiev (changes requested):
Good job. But you need to add a check if the title is not filled with spaces

Yeliseiev (approved):
Good job!

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,28 @@
...
+
+import { TodoWithUser } from '../../types';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: TodoWithUser;
+};
+
+export const TodoInfo: FC<Props> = ({ todo }) => {
+  const { id, title, user } = todo;
```
sergii-nosachenko, line 12: Add `completed` key also.

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,28 @@
...
+
+  return (
+    <article
+      data-id={id}
+      className={cn({
+        TodoInfo: true,
+        'TodoInfo--completed': todo.completed,
+      })}
```
sergii-nosachenko, line 20:
```suggestion
      className={cn(
        'TodoInfo',
        {
          'TodoInfo--completed': completed,
        },
      )}
```
Static classes can be passes as strings

src/App.tsx
```diff
@@ -1,61 +1,141 @@
...
-          </select>
+          <label>
+            {'User: '}
+            <select
+              data-cy="userSelect"
+              value={selectedUser}
+              onChange={(event) => {
+                setSubmittedUser(true);
+                setSelectedUser(event.target.value);
+              }}
```
sergii-nosachenko, line 110: Create separate handler for this event

src/App.tsx
```diff
@@ -1,61 +1,141 @@
...
+              type="text"
+              data-cy="titleInput"
+              placeholder="Enter a title"
+              value={inputText}
+              onChange={(event) => {
+                setSubmittedTitle(true);
+                const char = event.target.value;
+
+                setInputText(char.replace(/[^\p{L} ' ']/gu, ''));
+              }}
```
sergii-nosachenko, line 94: Create separate handler for this event

src/App.tsx
```diff
@@ -1,61 +1,141 @@
...
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
+        action="/api/users"
+        method="POST"
+        onSubmit={(event) => submitHandle(event)}
```
sergii-nosachenko, line 79:
```suggestion
        onSubmit={handleSubmit}
```
Follow [these](https://medium.com/javascript-in-plain-english/handy-naming-conventions-for-event-handler-functions-props-in-react-fc1cbb791364) naming conventions for methods

src/App.tsx
```diff
@@ -1,61 +1,141 @@
...
+  const [selectedUser, setSelectedUser] = useState('0');
+  const [submittedUser, setSubmittedUser] = useState(true);
+
+  const submitHandle = (event: React.FormEvent) => {
+    event.preventDefault();
+    if (inputText.trim() !== '') {
+      setSubmittedTitle(true);
+    } else {
+      setSubmittedTitle(false);
+    }
```
sergii-nosachenko, line 37:
```suggestion
    if (inputText.trim() === '') {
      setSubmittedTitle(false);
    
      return;
    }
```
Better avoid unnecessary code execution.

src/App.tsx
```diff
@@ -1,61 +1,141 @@
...
+
+    if (selectedUser !== '0') {
+      setSubmittedUser(true);
+    } else {
+      setSubmittedUser(false);
+    }
```
sergii-nosachenko, line 43: The same

src/App.tsx
```diff
@@ -1,61 +1,141 @@
...
+
+    if (inputText.trim() !== '' && selectedUser !== '0') {
```
sergii-nosachenko, line 45:
```suggestion
```
Can be skipped if you rewrite your upper code

src/App.tsx
```diff
@@ -1,61 +1,141 @@
...
+      setTodos(listOfTodos => {
+        const sorted = [...todos].sort((a, b) => (a.id - b.id));
+        const newId = sorted[sorted.length - 1].id + 1;
+        const user = users.find(searchUser => {
+          return searchUser.id === Number(selectedUser);
+        });
+
+        setInputText('');
+        setSelectedUser('0');
```
sergii-nosachenko, line 54:
```suggestion
```
Don't call state hooks from inside another state hook. It's a bad practice that can lead to bugs.

sergii-nosachenko (changes requested):

src/App.tsx
```diff
@@ -28,45 +28,62 @@ export const App = () => {
   const [selectedUser, setSelectedUser] = useState('0');
   const [submittedUser, setSubmittedUser] = useState(true);
 
-  const submitHandle = (event: React.FormEvent) => {
+  const handleSubmit = (event: React.FormEvent) => {
     event.preventDefault();
-    if (inputText.trim() !== '') {
-      setSubmittedTitle(true);
-    } else {
+    if (inputText.trim() === '' && selectedUser === '0') {
```
sergii-nosachenko, line 33: Optimize code:
```js
if (condition1 || condition2) {
  setState1(condition1);
  setState2(condition2);

  return;
}
```
You can skip a lot of duplicate code.

src/App.tsx
```diff
@@ -86,12 +103,7 @@ export const App = () => {
               data-cy="titleInput"
               placeholder="Enter a title"
               value={inputText}
-              onChange={(event) => {
-                setSubmittedTitle(true);
-                const char = event.target.value;
-
-                setInputText(char.replace(/[^\p{L} ' ']/gu, ''));
-              }}
+              onChange={(event) => handleInputOnChange(event.target.value)}
```
sergii-nosachenko, line 106:
```suggestion
              onChange={handleInputOnChange}
```
Process event as argument of your handlers. Here and in other cases. Even if it's one line of code.

sergii-nosachenko (changes requested):

src/App.tsx
Maksym845: Sorry, I can't understand your tip. In this task, I must throw an error (near the field) when the field is empty or Select is selected. So without these statements, it will not work. Because I need to set the Error in different situations:

- when the first field is empty;
- when the second field is empty;
- when both fields are empty.

sergii-nosachenko: Your error depends on condition. If one if your conditions is true you can set errors and exit the function. But you did it in 3 ifs. Combine it in one.
Example:
```js
const a = '';
const b = 1;

if (a === '' || b === 0) { // you have any combination of errors
  setErrorA(a === ''); // true
  set ErrorB(b === 0) // false

  return;
}
```

Maksym845: OMG, I got it, thank you very much for the explanation, it was so easy...

sergii-nosachenko (approved):
Perfect 👍👍👍

PR #1166, opened 2022-10-26
Implemented the todo form

anastasiia-nahaiska:
[DEMO LINK](https://anastasiia-nahaiska.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,144 @@
 import './App.scss';
+import React, { useState } from 'react';
+import { TodoList } from './components/TodoList';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoWithUser, User } from './react-app-env';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+const getUser = (userId: number): User | null => {
+  return usersFromServer.find(user => user.id === userId) || null;
+};
+
+const todoWithUser: TodoWithUser[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [isSubmit, setIsSubmit] = useState(false);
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,61 +1,159 @@
...
+    setIsButtonClicked(false);
+  };
+
+  const clearTitle = () => {
+    setTitle('');
+  };
+
+  const clearSelectedUserId = () => {
+    setSelectedUserId(0);
+  };
```
Apollonia5, line 46: You can combine the last two methods into a single reset method and move the button click logic right into the handle method.

Apollonia5 (approved):
GJ!

src/App.tsx
anastasiia-nahaiska: Yes, but I tried to follow this recommendation^_^
image

src/App.tsx
```diff
@@ -1,61 +1,159 @@
...
+  const [title, setTitle] = useState('');
+
+  const addTodo = () => {
+    const maxId = Math.max(...todos.map(todo => todo.id));
+    const newTodo = {
+      id: maxId + 1,
+      title,
+      completed: false,
+      userId: selectedUserId,
+      user: usersFromServer.find(user => user.id === selectedUserId) || null,
```
Anna-Tkachenko, line 30: you're already has function with this logic, use it here

src/App.tsx
```diff
@@ -1,61 +1,159 @@
...
+            id="createTitle"
+            type="text"
+            value={title}
+            placeholder="Enter a title"
+            data-cy="titleInput"
+            onChange={handleInputChange}
+          />
+
+          {isButtonClicked
+          && !title
```
Anna-Tkachenko, line 95: you should create separate boolean variables in state for handling errors for each input

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

PR #1168, opened 2022-10-26
Develop

ArturNagaycev:
[DEMO LINK](https://ArturNagaycev.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,118 @@
 import './App.scss';
+import React, { useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { Todo, TodoWithUser } from './react-app-env';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+const getUserById = (id: number) => {
+  return usersFromServer.find(user => user.id === id) || null;
+};
 
-export const App = () => {
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [todos, setTodos] = useState<Todo[]>(todosFromServer);
+  const [isTitleEmpty, setIsTitleEmpty] = useState(false);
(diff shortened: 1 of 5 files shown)


Vladosik1337 (commented):
Good Job 🔥

Vladosik1337 (approved):

src/App.tsx
```diff
@@ -1,61 +1,126 @@
...
-// import todosFromServer from './api/todos';
+const getUserById = (id: number) => {
+  return usersFromServer.find(user => user.id === id) || null;
+};
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [todos, setTodos] = useState<Todo[]>(todosFromServer);
+  const [isTitleEmpty, setIsTitleEmpty] = useState(false);
```
naziamoff, line 16: to me it seems that this case does not deserve to be a separate state, we can do it just by title.length

src/App.tsx
```diff
@@ -1,61 +1,126 @@
...
+  const [isUserNotSelected, setIsUserNotSelected] = useState(false);
```
naziamoff, line 17: same here

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,15 @@
...
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: TodoWithUser[],
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <section className="TodoList">
+    {todos.map(todo => (
+      <TodoInfo todo={todo} key={todo.id} data-id={todo.id} />
```
naziamoff, line 12: you should move each prop/attribute to a new line if you have more than 2 of them

naziamoff (approved):
Nice job! Still I suggest you to fix the comments

src/App.tsx
ArturNagaycev: if we use title.length the error message will be shown immediately after launch. But we need to see this message after button click

naziamoff: kk, agree with you

PR #1177, opened 2022-10-27
dynamic list of todo with form and error checks on fill

TheVovchik:
**[DEMO LINK](https://TheVovchik.github.io/react_add-todo-form/)**


nikitaovch99 (commented):
In the title, I'm setting blank spaces and choosing a user. But when I'm adding this todo. After a submit error doesn't go away. I don't think that this is the right behavior. https://prnt.sc/aqYZ2nuYnZcg

nikitaovch99 (approved):
GJ

naziamoff (approved):
well done 🔥

PR #1178, opened 2022-10-27
Implement todo list with adding new todos

imaksweb:
- [DEMO LINK](https://imaksweb.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,104 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import React, { useState } from 'react';
+import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import { TodoList } from './components/TodoList';
+import { Todo } from './react-app-env';
 
 export const App = () => {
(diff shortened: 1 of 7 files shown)


Vladosik1337 (approved):
Nice work 🔥

src/App.tsx
```diff
@@ -1,61 +1,121 @@
...
+              const inputValue = event.target.value;
+
+              setNewTodoTitle(inputValue);
+
+              if (inputValue.trim().length <= 0) {
+                setHasTitleError(true);
+              } else {
+                setHasTitleError(false);
+              }
+            }}
```
naziamoff, line 78: looks like the logic inside the click handler is too complicated to be inline. Can we make it a separate method?

src/App.tsx
```diff
@@ -1,61 +1,121 @@
...
+              const selectedValue = +event.target.value;
+
+              setSelectedUser(selectedValue);
+
+              if (selectedValue === 0) {
+                setHasUserIdError(true);
+              } else {
+                setHasUserIdError(false);
+              }
+            }}
```
naziamoff, line 99: same here

naziamoff (approved):
Well done, but better to separate handler's logic

PR #1180, opened 2022-10-27
add todo form created

Yeliseiev:
[DEMO LINK](https://Yeliseiev.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,1 +1,21 @@
-export const TodoList = () => {};
+import React from 'react';
+import { TodoInfo } from '../TodoInfo';
+import { TodoWithUser } from '../../react-app-env';
+
+type Props = {
+  todos: TodoWithUser[]
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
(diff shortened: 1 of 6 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,21 @@
...
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="TodoList">
+      {todos.map(todo => (
+        <React.Fragment key={todo.id}>
+          <TodoInfo
+            todo={todo}
+          />
```
Maksym845, line 16: Why did you add React.Fragment. I think it will be better to add a key in the TodoInfo component.

Maksym845 (approved):
Good solution

src/components/TodoList/TodoList.tsx
Yeliseiev: Just forgot that we can do like this

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Huge improvements, awesome 🚀🚀🚀

PR #1185, opened 2022-10-28
Implemented Todo list with dynamic adding of new todo

Dikey945:
[DEMO LINK](https://Dikey945.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,146 @@
 import './App.scss';
+import React, { FunctionComponent, useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { Todo } from './react-app-env';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+const getUserById = (userId: number) => {
+  const foundedUser = usersFromServer.find(user => user.id === userId);
 
-export const App = () => {
+  return foundedUser || null;
+};
+
+const todosWithUser: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.id),
+}));
+
+export const App: FunctionComponent = () => {
+  const [title, setTitle] = useState('');
+  const [authorId, setAuthorId] = useState(0);
+  const [todoList, setTodoList] = useState(todosWithUser);
+  const [titleError, setTitleError] = useState(false);
+  const [authorError, setAuthorError] = useState(false);
+  const id = Math.max(...todosFromServer.map(todo => todo.id));
+  const getNewTodoId = () => (
+    id + 1
+  );
+  const clearForm = () => {
+    setTitle('');
+    setAuthorId(0);
+    setTitleError(false);
+    setAuthorError(false);
+  };
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,146 @@
...
+export const App: FunctionComponent = () => {
+  const [title, setTitle] = useState('');
+  const [authorId, setAuthorId] = useState(0);
+  const [todoList, setTodoList] = useState(todosWithUser);
+  const [titleError, setTitleError] = useState(false);
+  const [authorError, setAuthorError] = useState(false);
+  const id = Math.max(...todoList.map(todo => todo.id));
+  const getNewTodoId = () => (
+    id + 1
+  );
```
xiimiik, line 28: maybe it will be better to create separate func and do all this logic there

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,32 @@
...
+      data-id={id}
+      className={classNames(
+        'TodoInfo',
+        { 'TodoInfo--completed': completed },
+      )}
+    >
+      <h2 className="TodoInfo__title">
+        {title}
+      </h2>
+      {user && <UserInfo user={user} />}
```
xiimiik, line 29:
```suggestion
      {user && (
        <UserInfo user={user} />
      )}
```

xiimiik (approved):
Nice!

src/components/TodoInfo/TodoInfo.tsx
Dikey945: I don`t think in this case it required to use of parentheses) BTW mentor will tell for sure 😄

src/App.tsx
```diff
@@ -1,61 +1,146 @@
...
+  user: getUserById(todo.id),
+}));
+
+export const App: FunctionComponent = () => {
+  const [title, setTitle] = useState('');
+  const [authorId, setAuthorId] = useState(0);
+  const [todoList, setTodoList] = useState(todosWithUser);
+  const [titleError, setTitleError] = useState(false);
+  const [authorError, setAuthorError] = useState(false);
+  const id = Math.max(...todoList.map(todo => todo.id));
```
Anna-Tkachenko, line 25: will be better to create separate function for defining id for new todo

src/App.tsx
```diff
@@ -1,61 +1,146 @@
...
+    setTitleError(false);
+    setAuthorError(false);
+  };
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    const trimmedTitle = title.trim();
+
+    if (!trimmedTitle) {
+      setTitleError(() => true);
```
Anna-Tkachenko, line 41: if you don't use current state for defining new value, functional setState is redundant, just pass needed value as an argument to the setTitleError function

src/App.tsx
```diff
@@ -1,61 +1,146 @@
...
+      setTodoList((prevTodoList) => [
+        ...prevTodoList,
+        newTodo,
+      ]);
+
+      clearForm();
+    }
+  };
+
+  const handleInput = (event: React.ChangeEvent<HTMLInputElement>) => {
```
Anna-Tkachenko, line 66: it's not clear for what input you created this handler, check function name

Anna-Tkachenko (approved):

PR #1188, opened 2022-10-28
implemented add todo form

nikitaovch99:
[DEMO LINK](https://nikitaovch99.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,110 @@
 import './App.scss';
+import React, { useState } from 'react';
+import { Todo, User } from './react-app-env';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [todosWithUser, setTodosWithUser] = useState(todos);
+  const [todoTitle, setTodoTitle] = useState('');
+  const [selectedUserById, setSelectedUserById] = useState(0);
+  const [hasTitleError, setHasTitleError] = useState(false);
+  const [hasUserIdError, setHasUserIdError] = useState(false);
+
+  const resetForm = () => {
+    setTodoTitle('');
+    setSelectedUserById(0);
+  };
+
+  const handleFormSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    setHasTitleError(!todoTitle);
+    setHasUserIdError(!selectedUserById);
+
+    const user = getUser(selectedUserById);
+
+    const newId = Math.max(...todosWithUser.map(todo => todo.id)) + 1;
+
+    const newTodo: Todo = {
+      id: newId,
+      title: todoTitle,
+      completed: false,
+      userId: selectedUserById,
+      user,
+    };
+
+    if (newTodo && selectedUserById && todoTitle) {
+      setTodosWithUser((currentTodos) => [...currentTodos, newTodo]);
+      resetForm();
+    }
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
+  const resetForm = () => {
+    setTodoTitle('');
+    setSelectedUserById(0);
+  };
+
+  const handleFormSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    setHasTitleError(!todoTitle);
+    setHasUserIdError(!selectedUserById);
```
TheVovchik, line 36: 👍

src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
+  return foundUser || null;
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [todosWithUser, setTodosWithUser] = useState(todos);
```
TheVovchik, line 21: need to define type of useState (it's not a primitive)

src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
+
+    const newTodo: Todo = {
+      id: newId,
+      title: todoTitle,
+      completed: false,
+      userId: selectedUserById,
+      user,
+    };
+
+    if (newTodo && selectedUserById && todoTitle) {
```
TheVovchik, line 50: selectedUserById && todoTitle we can create newTodo only if we have appropriate data

in this block we create it and transfer to state

TheVovchik (approved):
GJ

src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [todosWithUser, setTodosWithUser] = useState<Todo[]>(todos);
+  const [todoTitle, setTodoTitle] = useState('');
+  const [selectedUserById, setSelectedUserById] = useState(0);
+  const [hasTitleError, setHasTitleError] = useState(false);
+  const [hasUserIdError, setHasUserIdError] = useState(false);
```
naziamoff, line 25: I like your approach here, gj 👍 
It can be improved though. It is a good practice to start with positive - that helps to read the code easier.
```suggestion
  const [isTitleValid, setIsTitleValid] = useState(true);
  const [isUserIdValid, setIsUserIdValid] = useState(true);
```

naziamoff: just a suggestion here

src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
+
+  const handleFormSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    setHasTitleError(!todoTitle);
+    setHasUserIdError(!selectedUserById);
+
+    const user = getUser(selectedUserById);
+
+    const newId = Math.max(...todosWithUser.map(todo => todo.id)) + 1;
```
naziamoff, line 40: What you do now:
1. Iterating each todo into its id
2. Iterating each id to find the max value
3. Adding a one to it.

Why didn't you just use .length + 1?

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,28 @@
...
+    completed,
+    user,
+  } = todo;
+
+  return (
+    <article
+      data-id={todo.id}
+      className={completed
+        ? 'TodoInfo TodoInfo--completed'
+        : 'TodoInfo'}
```
naziamoff, line 20: you should use classnames library here

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,12 @@
...
+import { TodoInfo } from '../TodoInfo';
+import { Todo } from '../../react-app-env';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <section className="TodoList">
+    {todos.map((todo) => <TodoInfo todo={todo} key={todo.id} />)}
```
naziamoff, line 10: let's make a component from the new line

naziamoff (changes requested):

src/App.tsx
nikitaovch99: But the id's, are kinda random there can't use length. Or did I misunderstand you?

naziamoff: I agree 👍

naziamoff (approved):
🔥

PR #1191, opened 2022-10-28
added todo form

igor-1987:


src/react-app-env.d.ts
```diff
@@ -1 +1,20 @@
...
+  id: number,
+  title: string,
+  completed: boolean,
+  userId: number,
+  user: User,
+}
+
+export interface TodoWithUser extends Todo {
+  user: User | null,
+}
```
Yuriy-Komolov, line 20: Sergii Nosachenko told that it's bad practice to make types in react-acpp-env
It's better to make it in separate file it `types` folder

Yuriy-Komolov (approved):

Yuriy-Komolov:
Well done! But i guess it's too much separated files, hard to read such structure.

silchyn (changes requested):
Looks good, but I don't see demo link

igor-1987:
> Looks good, but I don't see demo link

Yeah, it's my bad. Fixed.

silchyn (approved):
Cool!

PR #1196, opened 2022-10-28
Develop

Proph7000:
[DEMO LINK](https://proph7000.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -26,19 +26,18 @@ export const App: React.FC = () => {
     const { value } = event.target;
     const lastIndex = value.length - 1;
 
-    if (value.length && pattern.test(value[lastIndex])) {
+    if (!value || pattern.test(value[lastIndex])) {
       setTitle(value);
     }
 
-    if (!event.target.value) {
-      setTitle(value);
-    }
+    setShowErrorTitle(false);
   }
 
   function handleUserId(event: React.ChangeEvent<HTMLSelectElement>): void {
     const { value } = event.target;
 
     setSelectedUserId(Number(value));
+    setShowErrorUser(false);
   }
 
   function handleSubmit(event: React.FormEvent<HTMLFormElement>) {
@@ -48,11 +47,11 @@ export const App: React.FC = () => {
       setShowErrorTitle(true);
     }
 
-    if (selectedUserId === 0) {
+    if (!selectedUserId) {
       setShowErrorUser(true);
     }
 
-    if (!title || selectedUserId === 0 || !title.trim()) {
+    if (!title || !selectedUserId || !title.trim()) {
       return;
     }
 
@@ -87,10 +86,7 @@ export const App: React.FC = () => {
               data-cy="titleInput"
               placeholder="Enter a title"
               value={title}
-              onChange={(event) => {
-                handleTitle(event);
-                setShowErrorTitle(false);
-              }}
+              onChange={handleTitle}
             />
           </label>
           {showErrorTitle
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,61 +1,148 @@
...
+import { TodoFromServer } from './types/TodoFromServer';
+import { TodoList } from './components/TodoList';
+import { getTodos } from './utils/getTodos';
+import { addTodo } from './utils/addTodo';
+
+const pattern = /[А-ЯёїіA-Z\d\s]/i;
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [selectedUserId, setSelectedUserId] = useState('0');
```
vlad-demchuk, line 16: use number  type instead of string

src/App.tsx
```diff
@@ -1,61 +1,148 @@
...
+    const lastIndex = value.length - 1;
+
+    if (value.length && pattern.test(value[lastIndex])) {
+      func(value);
+    }
+
+    if (!event.target.value) {
+      func(value);
+    }
+  }
```
vlad-demchuk, line 36: don't add a common function for onChange. 
just use separate functions: 
handleTitle/input/select = (event: React.ChangeEvent<HTMLInputElement>) {
setTitle/input/select..
}

src/App.tsx
```diff
@@ -1,61 +1,148 @@
...
+    }
+
+    setTodos(getTodos(
+      addTodo(
+        Number(selectedUserId),
+        maxCurrTodoId + 1,
+        title,
+        currTodos,
+      ),
+    ));
```
vlad-demchuk, line 69: it's not readable. create a variable with new object and then use it. and why do you use `getTodos`? use current state instead of it

src/App.tsx
```diff
@@ -1,61 +1,148 @@
...
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
+        action="/api/users"
+        method="POST"
+        onSubmit={(event) => handleSubmit(
+          event,
+          todos,
```
vlad-demchuk, line 84:
```suggestion
        onSubmit={handleSubmit}
```
use the current state instead of passing todos again. also you don't need to call another function inside a callback. just call your handleSubmit.

vlad-demchuk (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
+    const { value } = event.target;
+    const lastIndex = value.length - 1;
+
+    if (value.length && pattern.test(value[lastIndex])) {
+      setTitle(value);
+    }
+
+    if (!event.target.value) {
+      setTitle(value);
+    }
```
AlisaVlasova, line 35:
```suggestion
    if (!value || pattern.test(value[lastIndex])) {
      setTitle(value);
    }
```

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
+  }
+
+  function handleSubmit(event: React.FormEvent<HTMLFormElement>) {
+    event.preventDefault();
+
+    if (!title || !title.trim()) {
+      setShowErrorTitle(true);
+    }
+
+    if (selectedUserId === 0) {
```
AlisaVlasova, line 51:
```suggestion
    if (!selectedUserId) {
```

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
+            {'Title: '}
+
+            <input
+              type="text"
+              data-cy="titleInput"
+              placeholder="Enter a title"
+              value={title}
+              onChange={(event) => {
+                handleTitle(event);
+                setShowErrorTitle(false);
```
AlisaVlasova, line 92: роби це в хендлері

AlisaVlasova (changes requested):

AlisaVlasova (approved):

PR #1209, opened 2022-10-30
Solution

rstarovoit:
[DEMO LINK](https://rstarovoit.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,104 @@
+import React, { useState } from 'react';
 import './App.scss';
+import { TodoList } from './components/TodoList';
+import { Todo } from './react-app-env';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+const getTodoId = (todos: Todo[]) => {
+  const id = Math.max(...todos.map(todo => todo.id));
+
+  return id + 1;
+};
+
+const getUser = (userId: number) => (
+  usersFromServer.find(user => user.id === userId) || null
+);
+
+const todosAndUsers: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [todos, setTodos] = useState(todosAndUsers);
+  const [title, setTitle] = useState('');
+  const [todoUserId, setTodoUserId] = useState(0);
+  const [hasError, setHasError] = useState(false);
+
+  const reset = () => {
+    setTitle('');
+    setTodoUserId(0);
+    setHasError(false);
+  };
+
+  const onFormSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (!todoUserId || !title) {
+      setHasError(true);
+    } else {
+      setTodos(
+        [
+          ...todos,
+          {
+            id: getTodoId(todos),
+            title,
+            completed: false,
+            userId: todoUserId,
+            user: getUser(todoUserId),
+          },
+        ],
+      );
+
+      reset();
+    }
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
+        action="/api/users"
+        method="POST"
+        onSubmit={event => onFormSubmit(event)}
+      >
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <input
+            type="text"
+            data-cy="titleInput"
+            placeholder="Enter a title"
+            value={title}
+            onChange={event => setTitle(event.target.value)}
+          />
+          {(!title && hasError) && (
+            <span className="error">Please enter a title</span>
+          )}
         </div>
 
         <div className="field">
-          <select data-cy="userSelect">
+          <select
+            data-cy="userSelect"
+            value={todoUserId}
+            onChange={event => setTodoUserId(+event.target.value)}
+          >
             <option value="0" disabled>Choose a user</option>
+            {usersFromServer.map(user => {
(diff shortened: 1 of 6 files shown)


grafit-off (approved):
LGTM 🆗

src/App.tsx
```diff
@@ -1,61 +1,112 @@
...
+          <input
+            type="text"
+            data-cy="titleInput"
+            placeholder="Enter a title"
+            value={title}
+            onChange={event => setTitle(event.target.value)}
+          />
+          {(!title && hasError) && (
+            <span className="error">Please enter a title</span>
+          )}
```
sergii-nosachenko, line 78: You need to create 2 errors - one for Title, one for User. Use only one source of truth for every field

src/App.tsx
```diff
@@ -1,61 +1,112 @@
...
+              return (
+                <option value={id} key={id}>
+                  {name}
+                </option>
+              );
+            })}
           </select>
 
-          <span className="error">Please choose a user</span>
+          {(hasError && !todoUserId) && (
```
sergii-nosachenko, line 99: Create userHasError f.e.

src/App.tsx
```diff
@@ -1,61 +1,112 @@
...
+    setTodoUserId(0);
+    setHasError(false);
+  };
+
+  const onFormSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (!todoUserId || !title) {
+      setHasError(true);
+    } else {
```
sergii-nosachenko, line 41: You don't need else branch. If you have an error - set errors and return immediately from function.

src/App.tsx
```diff
@@ -1,61 +1,112 @@
...
+  const reset = () => {
+    setTitle('');
+    setTodoUserId(0);
+    setHasError(false);
+  };
+
+  const onFormSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (!todoUserId || !title) {
```
sergii-nosachenko, line 39: Check if title is a blank string with spaces only. Use trim().

src/App.tsx
```diff
@@ -1,61 +1,112 @@
...
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
+        action="/api/users"
+        method="POST"
+        onSubmit={event => onFormSubmit(event)}
```
sergii-nosachenko, line 66:
```suggestion
        onSubmit={handleSubmit}
```
Use proper naming `handleSubmit` and move logic there.

sergii-nosachenko (changes requested):

src/App.tsx
```diff
@@ -25,19 +25,22 @@ export const App = () => {
...
-  const onFormSubmit = (event: React.FormEvent) => {
+  const handleSubmit = (event: React.FormEvent) => {
     event.preventDefault();
 
-    if (!todoUserId || !title) {
-      setHasError(true);
+    if (!todoUserId || !title.trim()) {
+      setHasTitleError(true);
+      setHasUserError(true);
     } else {
```
sergii-nosachenko, line 44: Better immediately return from function if you have errors. Using else is bad practice.

src/App.tsx
```diff
@@ -73,7 +76,7 @@ export const App = () => {
             value={title}
             onChange={event => setTitle(event.target.value)}
           />
-          {(!title && hasError) && (
+          {(!title && hasTitleError) && (
```
sergii-nosachenko, line 79: Don't check same condition twice. Use only error state here.

sergii-nosachenko (changes requested):

src/App.tsx
rstarovoit: without else i can't pass tests
image

src/App.tsx
rstarovoit: without "!title" code doesn't work correctly(

src/App.tsx
```diff
@@ -96,7 +99,7 @@ export const App = () => {
             })}
           </select>
 
-          {(hasError && !todoUserId) && (
+          {(hasUserError && !todoUserId) && (
```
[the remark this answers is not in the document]
rstarovoit: without "!todoUserId" code doesn't work correctly(

src/App.tsx
sergii-nosachenko: @rstarovoit The reason of this that you need to improve your logic here. Look at the example below:
```js
if (someErrorCondition1 || someErrorCondition2) {
  setError1(someErrorCondition1);
  setError2(someErrorCondition2);
  
  return;
}

makeSomethingIfOk();
```

src/App.tsx
sergii-nosachenko: @rstarovoit Problem is in your submit handler

sergii-nosachenko (changes requested):
Read comments

src/App.tsx
rstarovoit: Many thanks for the example and explanations!

src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+    setTodoUserId(0);
+    setHasTitleError(false);
+    setHasUserError(false);
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (!todoUserId || !title.trim()) {
+      if (!title.trim()) {
```
Andr1yK, line 42: u can save `title.trim` to const

src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+        setHasTitleError(true);
+      }
+
+      if (todoUserId === 0) {
+        setHasUserError(true);
+      }
```
Andr1yK, line 48: also, u case write this condition inside a setState `setHasUserError(todoUserId === 0);`

Andr1yK (approved):

PR #1211, opened 2022-10-30
add-todo-form

oleksandr-korobko:
[DEMO LINK](https://oleksandr-korobko.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,148 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
 
-export const App = () => {
+import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+
+const getUser = (userId: number): User | null => {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+};
+
+const todosWithUsers: Todo[] = todosFromServer.map(todo => ({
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,148 @@
...
+  return foundUser || null;
+};
+
+const todosWithUsers: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+const getIdForTodo = (todos: Todo[]) => {
+  let biggest = 0;
```
avkovlit, line 23: u better use more understanding naming, biggestId will be better

avkovlit (approved):
GJ!!

src/App.tsx
Anna-Tkachenko: agree, fix @avkovlit comment

src/App.tsx
```diff
@@ -1,61 +1,148 @@
...
+
+  todos.forEach(todo => {
+    if (todo.id > biggest) {
+      biggest = todo.id;
+    }
+  });
```
Anna-Tkachenko, line 29: you can also use Math.max for defining max value

src/App.tsx
```diff
@@ -1,61 +1,148 @@
...
+
+    const todo: Todo = {
+      id: getIdForTodo(todos),
+      title: titleInput,
+      completed: false,
+      userId: userSelect,
+      user: getUser(userSelect),
+    };
+
+    setTodos(prevTodos => [...prevTodos, todo]);
```
Anna-Tkachenko, line 71: it's not prevTodos, it's current todos

Anna-Tkachenko (approved):

PR #1212, opened 2022-10-30
add form implemented - v.1

korniienkokarina:
[DEMO LINK](https://korniienkokarina.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,118 @@
 import './App.scss';
+import React, { useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-export const App = () => {
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,118 @@
...
+      setTitleIsCorrect(false);
+    }
+
+    if (!user) {
+      setUserIsCorrect(false);
+    }
+
+    if (title.trim() && user) {
+      setList(
+        [...list, {
```
Arsen-Shpak, line 45: mb better move this to separate function

src/App.tsx
```diff
@@ -1,61 +1,118 @@
...
+          id: getNewId(list),
+          userId: +user,
+          title,
+          completed: false,
+          user: getUser(+user),
+        }],
+      );
+
+      setTitle('');
```
Arsen-Shpak, line 54: create reset function for this

Arsen-Shpak (approved):

src/App.tsx
Anna-Tkachenko: agree, will be better to create separate function for reset form functionality

Anna-Tkachenko (approved):

PR #1217, opened 2022-10-31
solution

Veremchukk:
- [DEMO LINK](https://Veremchukk.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,102 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { TodoWithUser } from './types/TodoWithUser';
 
-export const App = () => {
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+const findUserById = (userId: number) => (
+  usersFromServer.find(user => user.id === userId) || null
+);
+const todosWithUsers: TodoWithUser[] = todosFromServer.map(todo => {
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
+      >
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <input
+            type="text"
+            data-cy="titleInput"
+            placeholder="Enter a title"
+            value={titleValue}
+            onChange={(event) => setTitleValue(event.target.value)}
```
sergii-nosachenko, line 72: Add handler function

src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
+    setSelectedUser(0);
+    setSubmitTouched(false);
+  };
+
+  const onFormSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (!selectedUser || !titleValue) {
+      setSubmitTouched(true);
+    } else {
```
sergii-nosachenko, line 39: Return from function immediately if you have errors. Don't use else in this case.
Also you need to check if your title is blank string only with spaces. Use trim()

src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
+      setTodosList(
+        [
+          ...todosList,
+          {
+            id: Math.max(...todosList.map(todo => todo.id)) + 1,
```
sergii-nosachenko, line 44: Create helper function for this calculation

sergii-nosachenko (changes requested):

PR #1222, opened 2022-11-02
implemented add todo form

ArtemHulkoDev:
[DEMO LINK](https://ArtemHulkoDev.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,122 @@
+import { useState } from 'react';
 import './App.scss';
+import { TodoList } from './components/TodoList';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
 
-export const App = () => {
+  return foundUser || null;
+}
+
+export const initialTodos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+const maxTodoId = initialTodos
+  .map(todo => todo.id)
+  .sort((a, b) => a - b)[initialTodos.length - 1];
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [todos, setTodos] = useState(initialTodos);
+  const [todoId, setTodoId] = useState(maxTodoId);
+  const [isEmptyTitle, setIsEmptyTitle] = useState(false);
+  const [isEmptyUser, setIsEmptyUser] = useState(false);
+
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,130 @@
...
+        setIsEmptyTitle(true);
+      }
+
+      if (userId === 0) {
+        setIsEmptyUser(true);
+      }
+
+      return;
+    }
+
```
vallerii, line 59: For what did you wrap if in if?
why you cant use only 2 'if'?
like:
      if (title.trim() === '') {
        setIsEmptyTitle(true);
      }

      if (userId === 0) {
        setIsEmptyUser(true);
      }

ArtemHulkoDev: here is the reason on the screenshot, it is needed in order to show the error for both cases if the user is not chosen and the title is not entered and stop the function by return
Screenshot_101

vallerii:
GJ 👍

src/App.tsx
```diff
@@ -1,61 +1,130 @@
...
+}));
+
+const maxTodoId = initialTodos
+  .map(todo => todo.id)
+  .sort((a, b) => a - b)[initialTodos.length - 1];
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [todos, setTodos] = useState(initialTodos);
```
Denchepornyuk, line 27: add type for state `todos`

src/App.tsx
```diff
@@ -1,61 +1,130 @@
...
+
+  const changeId = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setUserId(+event.target.value);
+    if (+event.target.value !== 0) {
+      setIsEmptyUser(false);
+    }
+  };
+
+  const handleSubmit = (event: React.ChangeEvent<HTMLFormElement>) => {
+    event?.preventDefault();
```
Denchepornyuk, line 47:
```suggestion
    event.preventDefault();
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,25 @@
...
+import classNames from 'classnames';
+import { UserInfo } from '../UserInfo';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todo: Todo
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  return (
```
Denchepornyuk, line 10: you don't need return here

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,16 @@
-export const TodoList = () => {};
+import { TodoInfo } from '../TodoInfo';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todos: Todo[]
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
```
Denchepornyuk, line 9: the same, avoid return if you can

Denchepornyuk (approved):

PR #1229, opened 2022-11-06
implemented ADD Todo Form

VovaDzhuryk:
[DEMO LINK](https://VovaDzhuryk.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,1 +1,31 @@
-export const TodoList = () => {};
+import { FC } from 'react';
+import classNames from 'classnames';
+import { FullTodoInfo } from '../../types/Type';
+import { TodoInfo } from '../TodoInfo';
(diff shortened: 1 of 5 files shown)


glugovsky (approved):

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,31 @@
...
+        return (
+          <article
+            data-id={id}
+            className={classNames('TodoInfo',
+              { 'TodoInfo--completed': completed })}
+            key={id}
+          >
+            <TodoInfo todo={todo} />
+            {user && <UserInfo user={user} />}
+          </article>
```
OleksiiBlezniuk, line 26: You have created component for this purpose, why to render all info in list?

OleksiiBlezniuk: Btw, would be better to use semantic tags for list here

OleksiiBlezniuk (approved):

PR #1241, opened 2022-11-13
solution

bradleymatjie:
[DEMO LINK](https://bradleymatjie.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,126 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
 
-export const App = () => {
+import { User } from './types/User';
+import { TodoWithUser } from './types/TodoWithUser';
+import { TodoList } from './components/TodoList';
+
+function getUserById(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+const getTodoId = (todos: TodoWithUser[]) => {
+  const id = Math.max(...todos.map(todo => todo.id));
+
+  return id + 1;
+};
+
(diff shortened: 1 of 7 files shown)


mayuces (approved):
Approved

src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+
+import { User } from './types/User';
+import { TodoWithUser } from './types/TodoWithUser';
+import { TodoList } from './components/TodoList';
+
+function getUserById(userId: number): User | null {
```
VinnieJ-2k20, line 11: In naming convention function that `gets` user should throw error if it has not been found.
Here it's better to use `find`

```suggestion
function findUserById(userId: number): User | null {
```

src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+const getTodoId = (todos: TodoWithUser[]) => {
```
VinnieJ-2k20, line 17:
```suggestion
const getNextTodoId = (todos: TodoWithUser[]) => {
```

src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+  const [errorUser, setErrorUser] = useState(false);
+  const [errorTitle, setErrorTitle] = useState(false);
+  const [availableTodos, setAvailableTodos] = useState(todos);
+
+  const clearState = () => {
+    setTitle('');
+    setUserId(0);
+  };
+
+  const handleTitle = (value: string) => {
```
VinnieJ-2k20, line 40:
```suggestion
  const handleTitleChange = (value: string) => {
```

VinnieJ-2k20 (approved):
Looks and works fine. Pay attention to variable names.

PR #1247, opened 2022-11-16
React add todo form

MeshackDev:
[DEMO LINK](https://MeshackDev.github.io/react_add-todo-form/)


mayuces (changes requested):
There is no checkmark.

PR #1248, opened 2022-11-17
Solution

oles-partyka:
[DEMO LINK](https://oles-partyka.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,126 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
 
-export const App = () => {
+import { User } from './types/User';
+import { TodoWithUser } from './types/TodoWithUser';
+import { TodoList } from './components/TodoList';
+
+function getUserById(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+const getTodoId = (todos: TodoWithUser[]) => {
+  const id = Math.max(...todos.map(todo => todo.id));
+
+  return id + 1;
+};
+
+const todos: TodoWithUser[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [errorUser, setErrorUser] = useState(false);
+  const [errorTitle, setErrorTitle] = useState(false);
+  const [availableTodos, setAvailableTodos] = useState(todos);
+
+  const clearState = () => {
+    setTitle('');
+    setUserId(0);
+  };
+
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+const getTodoId = (todos: TodoWithUser[]) => {
+  const id = Math.max(...todos.map(todo => todo.id));
+
+  return id + 1;
+};
+
+const todos: TodoWithUser[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
```
qoa11a, line 26: Why do you use function declaration and arrow functions? Better keep one style.

src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [errorUser, setErrorUser] = useState(false);
+  const [errorTitle, setErrorTitle] = useState(false);
```
qoa11a, line 32: The naming is a little bit confusing. I'd assume there's an error text inside this state, but there's boolean. Either store here the error text and than use it in JSX, or name it starting with 'is'.

src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+
+  const clearState = () => {
+    setTitle('');
+    setUserId(0);
+  };
+
+  const handleTitle = (value: string) => {
+    setErrorTitle(false);
+    setTitle(value);
+  };
```
qoa11a, line 43: After word 'handle' should be a verb. It's better to name it handleTitleChange.

src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+            <span>Title: </span>
+
+            <input
+              type="text"
+              data-cy="titleInput"
+              placeholder="Enter a title"
+              value={title}
+              onChange={(event) => (
+                handleTitle(event.target.value)
+              )}
```
qoa11a, line 90: I recommend to pass just an event to handleTitle function. Than in this function you can get event.target.value. So the syntax will be so after these changes.
```suggestion
              onChange={handleTitle}
```

src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+            />
+          </label>
+          {errorTitle
+            && (
+              <span className="error">
+                Please enter a title
+              </span>
+            )}
```
qoa11a, line 98:
```suggestion
          {errorTitle && (
            <span className="error">
              Please enter a title
            </span>
          )}
```

src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+          <label htmlFor="userSelect">User: </label>
+
+          <select
+            id="userSelect"
+            data-cy="userSelect"
+            value={userId}
+            onChange={(event) => {
+              setErrorUser(false);
+              setUserId(+event.target.value);
+            }}
```
qoa11a, line 111: Why did you create separate function for title change, but you didn't for user change? Let's create it as well.

qoa11a (changes requested):
Great job! 👍 
Have some improvement recommendations.

PR #1253, opened 2022-11-21
add_todo

boriskerekesha:
https://boriskerekesha.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,140 @@
+import React, { useState, useEffect } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
 
-export const App = () => {
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [userName, setUserName] = useState('');
+  const [todosItems, setTodosItems] = useState(todos);
+  const [isSubmit, setIsSubmit] = useState(false);
+  const [isVisibleTittleError, setIsVisibleTittleError] = useState(false);
+
+  useEffect(() => {
+    setIsVisibleTittleError(false);
+  }, [title]);
+
+  const handlerSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+    setIsSubmit(true);
+
+    if (!title || !userName) {
+      setIsVisibleTittleError(true);
+
+      return;
+    }
+
+    const correctTitle = title.trim().replace(/[^\d\sA-Za-zA-ЯА-я]/gi, '');
+
+    if (!correctTitle.length) {
+      // eslint-disable-next-line no-alert
+      alert('Title can contains letters (ru and en), digits, and spaces');
+
+      return;
+    }
+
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,140 @@
...
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+
+function getUser(userId: number): User | null {
```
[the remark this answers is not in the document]
vlad-demchuk: it will be clearer

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,7 @@
-export const UserInfo = () => {};
+import { UserInfoType } from '../../types/UserInfoType';
+
+export const UserInfo: React.FC<UserInfoType> = ({ user }) => (
+  <a className="UserInfo" href={`mailto:${user?.email}`}>
+    {user?.name}
+  </a>
+);
```
tetiana-ivanchenko, line 7: you can only call UserInfo if your user is not null, then you don't need ? after user
and also use destructuring, for the user

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
+              disabled
+            >
+              Choose a user
+            </option>
+            {usersFromServer.map((user) => (
+              <option key={user.id} value={user.name}>
+                {user.name}
+              </option>
+
+            ))}
```
tetiana-ivanchenko, line 125:
```suggestion
              </option>
            ))}
```

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
+      user: selectedUser || null,
+    };
+
+    setTodosItems((currentTodos) => [
+      ...currentTodos,
+      newTodo,
+    ]);
+    setTitle('');
+    setUserName('');
+    setIsSubmit(false);
```
tetiana-ivanchenko, line 69: move it to the separate method, like "clearForm" or "resetForm"

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
+    event.preventDefault();
+    setIsSubmit(true);
+
+    if (!title || !userName) {
+      setIsVisibleTittleError(true);
+
+      return;
+    }
+
+    const correctTitle = title.trim().replace(/[^\d\sA-Za-zA-ЯА-я]/gi, '');
```
tetiana-ivanchenko, line 41: not so good idea to replace some data that user enters without some warns

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
+      newTodo,
+    ]);
+    setTitle('');
+    setUserName('');
+    setIsSubmit(false);
+  };
+
+  const handlerInput = (
+    event: React.ChangeEvent<HTMLInputElement>,
+  ) => setTitle(event.currentTarget.value);
```
tetiana-ivanchenko, line 74:
```suggestion
  const handleInput = (
    event: React.ChangeEvent<HTMLInputElement>,
  ) => setTitle(event.currentTarget.value);
```

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
+  const [userName, setUserName] = useState('');
+  const [todosItems, setTodosItems] = useState(todos);
+  const [isSubmit, setIsSubmit] = useState(false);
+  const [isVisibleTittleError, setIsVisibleTittleError] = useState(false);
+
+  useEffect(() => {
+    setIsVisibleTittleError(false);
+  }, [title]);
+
+  const handlerSubmit = (event: React.FormEvent) => {
```
tetiana-ivanchenko, line 31:
```suggestion
  const handleSubmit = (event: React.FormEvent) => {
```

tetiana-ivanchenko (changes requested):

vlad-demchuk (approved):

PR #1264, opened 2022-11-28
Develop

andriy-shymkiv:
[DEMO LINK](https://andriy-shymkiv.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 import './App.scss';
-import { useState } from 'react';
+import React, { useState } from 'react';
 
 import todosFromServer from './api/todos';
 import usersFromServer from './api/users';
@@ -52,6 +52,21 @@ export const App = () => {
     }
   };
 
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+    setTittleError(false);
+  };
+
+  const handleUserSelection = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setSelectedUser(+event.target.value);
+    setUserError(false);
+  };
+
+  const handleFormSubmit = (event: React.ChangeEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    addTodo(selectedUser);
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
@@ -59,10 +74,7 @@ export const App = () => {
       <form
         action="/api/users"
         method="POST"
-        onSubmit={(event) => {
-          event.preventDefault();
-          addTodo(selectedUser);
-        }}
+        onSubmit={handleFormSubmit}
       >
         <div className="field">
           <label>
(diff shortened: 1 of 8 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,33 @@
...
+export const TodoInfo: React.FC<Props> = ({ todos }) => {
+  const {
+    title,
+    completed,
+    user,
+  } = todos;
+
+  return (
+    <article
+      data-id={todos.id}
```
andrii-bassarab, line 18:
```suggestion
  const {
    id,
    title,
    completed,
    user,
  } = todos;

  return (
    <article
      data-id={id}
```
it is better to also destruct the id and use it as follows

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,14 @@
...
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: Array<Todo>;
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <section className="TodoList">
+    {todos.map(item => (
+      <TodoInfo key={item.id} todos={item} />
```
andrii-bassarab, line 11:
```suggestion
      <TodoInfo key={item.id} todo={item} />
```
it is better to call 'todo', because you pass one element of todo to the props, not the whole array, which is probably why not all tests pass

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,33 @@
-export const TodoInfo = () => {};
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todos: Todo
+};
+
+export const TodoInfo: React.FC<Props> = ({ todos }) => {
```
andrii-bassarab, line 9:
```suggestion
export const TodoInfo: React.FC<Props> = ({ todo }) => {
```
same here

andrii-bassarab (approved):
GJ

src/App.tsx
```diff
@@ -1,60 +1,139 @@
...
+  const [selectedUser, setSelectedUser] = useState(0);
+  const [title, setTitle] = useState('');
+  const [titleError, setTittleError] = useState(false);
+  const [userError, setUserError] = useState(false);
+
+  const addTodo = (id: number) => {
+    const foundedUser = usersFromServer.find(user => user.id === id);
+    const uniqueId = Math.max(...todos.map(todo => todo.id)) + 1;
+
+    if (title.trim().length === 0) {
```
mykola-koval, line 33: should work like this too
```suggestion
    if (title.trim().length) {
```

mykola-koval (approved):

src/App.tsx
Andriy-Shymkiv: no it doesn't 😢

PR #1276, opened 2022-11-29
v01

Anna-Shapovalova:
[DEMO LINK](https://Anna-Shapovalova.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,121 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { User } from './types/User';
+import { Todo } from './types/Todo';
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,129 @@
...
+        return [
+          ...prevTodos,
+          foundedUser,
+        ];
+      });
+
+      setUserId(0);
+      setTitle('');
+      setEnteredTitle(true);
+      setChoosedUser(true);
```
Svitlana-Yudina, line 49: I'm not sure, but maybe you don't need these setters here, if you set "true" inside the form in "onChange".

Svitlana-Yudina (approved):
Great Job!))

src/App.tsx
Anna-Shapovalova: Here I reset the values to default) But U`r right, everything is OK without these actions)

Svitlana-Yudina: I did exactly the same myself. And then I found that without it, it also works))

mykola-koval (approved):

PR #1292, opened 2022-12-01
add todo form is ready

shtht24:
[DEMO LINK](https://shtht24.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,116 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
 
-export const App = () => {
+function getUserById(userId: number): User | null {
+  return usersFromServer.find(user => user.id === userId) || null;
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [userId, setUserId] = useState(0);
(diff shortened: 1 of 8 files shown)


MikhailKucherina (approved):
Looks good!

src/App.tsx
```diff
@@ -1,61 +1,124 @@
...
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [userId, setUserId] = useState(0);
+  const [title, setTitle] = useState('');
+  const [isFailedToSubmit, setIsFailedToSubmit] = useState(false);
```
illiarerun, line 22: I don't see why you need this, you can create 2 different states for `titleError` and `userError` and update them onSubmit or onChange
```suggestion
  const [title, setTitle] = useState('');
```

src/App.tsx
```diff
@@ -1,61 +1,124 @@
...
+  const [currentTodos, setCurrentTodos] = useState(todos);
+  const hasNoTitle = title.trim() === '' && isFailedToSubmit;
```
illiarerun, line 24: so you wouldn't need this either

src/App.tsx
```diff
@@ -1,61 +1,124 @@
...
+            <option value="0">Choose a user</option>
+            {usersFromServer.map(user => (
+              <option value={user.id} key={user.id}>
+                {user.name}
+              </option>
+            ))}
           </select>
 
-          <span className="error">Please choose a user</span>
+          {!userId && isFailedToSubmit && (
```
illiarerun, line 109: then you only need to check `userError` here

illiarerun (changes requested):

src/App.tsx
shtht24: ty, i've already fixed form)

illiarerun (approved):

PR #1309, opened 2022-12-15
Added the task solution

oleksusov:
[DEMO LINK](https://oleksusov.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,157 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import React, { useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
 
-export const App = () => {
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [newTodos, setNewTodos] = useState(todos);
+  const [invalidInput, setInvalidInput] = useState(false);
+  const [invalidSelect, setInvalidSelect] = useState(false);
+  const [title, setTitle] = useState('');
+  const [selectedUser, setSelectedUser] = useState('0');
+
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,157 @@
...
+
+      default:
+        break;
+    }
+  };
+
+  const submitChanges = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if ((title === '') && (selectedUser === '0')) {
```
loralevitska, line 54:
```suggestion
    if (!title && !Number(selectedUser))) {
```

src/App.tsx
```diff
@@ -1,61 +1,157 @@
...
+      return;
+    }
+
+    if (title === '') {
+      setInvalidInput(true);
+
+      return;
+    }
+
+    if (selectedUser === '0') {
```
loralevitska, line 67:
```suggestion
    if (!Number(selectedUser)) {
```

src/App.tsx
```diff
@@ -1,61 +1,157 @@
...
+      setInvalidSelect(true);
+
+      return;
+    }
+
+    setNewTodos([
+      ...newTodos,
+      {
+        id: (Math.max(...newTodos.map(todo => todo.id)) + 1),
```
loralevitska, line 76:
```suggestion
        id: newTodos.length + 1),
```

src/App.tsx
```diff
@@ -1,61 +1,157 @@
...
+                key={user.id}
+              >
+                {user.name}
+              </option>
+            ))}
           </select>
 
-          <span className="error">Please choose a user</span>
+          {invalidSelect
+            && (<span className="error">Please choose a user</span>)}
```
loralevitska, line 143:
```suggestion
          {invalidSelect && (
            <span className="error">Please choose a user</span>
          )}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,19 @@
...
+  return (
+    <section className="TodoList">
+      {todos.map(todo => (
+        <TodoInfo
+          key={todo.id}
+          todo={todo}
+        />
+      ))}
+    </section>
+  );
```
loralevitska, line 18: use semantic tags for list `ul` `li`

loralevitska (changes requested):
almost done)

src/App.tsx
oleksusov: I agree this variant is better, but in this case I can't use it because in 'todosFromServer' we have had todo with id = 15 and length of todos array is 3

PR #1312, opened 2022-12-19
all done

Yevhen-Pavlovskyi:
https://yevhen-pavlovskyi.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,123 @@
 import './App.scss';
+import React, { useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-export const App = () => {
+import { TodoList } from './components/TodoList';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+
+function findUserId(userId: number): User | null {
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,25 +1,123 @@
...
+  return foundUser || null;
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: findUserId(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [selectedUser, setSelectedUser] = useState(0);
```
loralevitska, line 23: Why is the default value an object? A user is an object

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,14 @@
...
+
+type Props = {
+  todos: Todo[];
+};
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <section className="TodoList">
+    {todos.map(todo => (
+      <TodoInfo key={todo.id} todo={todo} />
+    ))}
+  </section>
```
loralevitska, line 13: use semantic tags for list `ul` `li`

loralevitska (changes requested):
almost perfect)

qoa11a (approved):

PR #1313, opened 2022-12-19
Solution

UmizDemud:
[DEMO LINK](https://UmizDemud.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,165 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList/TodoList';
 
-export const App = () => {
+enum ChangeType {
+  TITLE,
+  USER,
+  COMPLETED,
+}
+
+export const App: React.FC<{}> = () => {
+  const [title, setTitle] = useState('');
+  const [user, setUser] = useState(0);
+  const [completed, setCompleted] = useState(false);
+  const [todos, setTodos] = useState(
+    todosFromServer.sort((a, b) => a.id - b.id),
+  );
+
+  const [errors, setErrors] = useState({
+    userNotSelected: false,
+    titleIsEmpty: false,
+  });
+
+  const clearForm = () => {
+    setTitle('');
+    setUser(0);
+    setCompleted(false);
+  };
+
+  const handleSubmit = (
+    e: React.SyntheticEvent | React.MouseEvent<HTMLButtonElement>,
+  ) => {
+    e.preventDefault();
+    setErrors({
+      userNotSelected: user === 0,
+      titleIsEmpty: title.length === 0,
+    });
+
+    setTodos(prev => {
+      if (user === 0 || title.length === 0) {
+        return prev;
+      }
+
+      const newTodo = {
+        id: prev[prev.length - 1].id + 1,
+        title,
+        userId: user,
+        completed,
+      };
+
+      clearForm();
+
+      return ([
+        ...prev,
+        newTodo,
+      ]);
+    });
+  };
+
+  const handleChange = (
+    e: React.ChangeEvent<HTMLInputElement | HTMLSelectElement>,
+    field: ChangeType,
+  ) => {
+    switch (field) {
+      case ChangeType.TITLE:
+        setTitle(e.target.value);
+        setErrors(prev => ({
+          userNotSelected: prev.userNotSelected,
+          titleIsEmpty: false,
+        }));
+        break;
+      case ChangeType.USER:
+        setUser(Number.parseInt(e.target.value, 10));
+        setErrors(prev => ({
+          userNotSelected: false,
+          titleIsEmpty: prev.titleIsEmpty,
+        }));
+        break;
+      default: // ChangeType.COMPLETED
+        setCompleted(prev => !prev);
+        break;
+    }
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
+        action="/api/users"
+        method="POST"
+        onSubmit={handleSubmit}
+      >
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,165 @@
...
+  USER,
+  COMPLETED,
+}
+
+export const App: React.FC<{}> = () => {
+  const [title, setTitle] = useState('');
+  const [user, setUser] = useState(0);
+  const [completed, setCompleted] = useState(false);
+  const [todos, setTodos] = useState(
+    todosFromServer.sort((a, b) => a.id - b.id),
```
naziamoff, line 19: let's not caclulate the state inside the useState, better to make it a separate variable

src/App.tsx
```diff
@@ -1,61 +1,165 @@
...
+  );
+
+  const [errors, setErrors] = useState({
+    userNotSelected: false,
+    titleIsEmpty: false,
```
naziamoff, line 24: booleans should start with `is`, `should`, `can` etc.

src/App.tsx
```diff
@@ -1,61 +1,165 @@
...
+            id="title"
+            name="title"
+            type="text"
+            value={title}
+            data-cy="titleInput"
+            onChange={e => handleChange(e, ChangeType.TITLE)}
+            placeholder="Enter title..."
+          />
+          {errors.titleIsEmpty
+          && <span className="error">Please enter a title</span>}
```
naziamoff, line 109: let's polish it a bit
```suggestion
          {errors.titleIsEmpty && (
            <span className="error">
              Please enter a title
            </span>
           )}
```

naziamoff (changes requested):
let's fix those + checkbox looks totally unclear to me

naziamoff (commented):
also, would be nice to add some styling

src/components/AddTodoForm/AddTodoForm.tsx
```diff
@@ -0,0 +1,182 @@
...
+  ) => void;
+  users: User[],
+};
+
+export const AddTodoForm = ({ handleSubmit, users }: Props) => {
+  const [title, setTitle] = useState('');
+  const [user, setUser] = useState(0);
+  const [completed, setCompleted] = useState(false);
+  const [errors, setErrors] = useState({
+    isUserNotSelected: false,
```
naziamoff, line 25: not really good idea to start with negative in booleans, better to avoid it. It is a bit difficult to compile in my head, like `isNotSelected = false`. It means `isUserSelected = true`?

src/components/AddTodoForm/AddTodoForm.tsx
```diff
@@ -0,0 +1,182 @@
...
+  const onSubmit = (
+    e: React.SyntheticEvent | React.MouseEvent<HTMLButtonElement>,
+  ) => {
+    e.preventDefault();
+    setErrors({
+      isUserNotSelected: user === 0,
+      isTitleEmpty: title.length === 0,
+    });
+
+    if (user === 0 || title.length === 0) {
```
naziamoff, line 74: user === 0 seems really illogical. I expect user to be an object - we get number instead

src/components/AddTodoForm/AddTodoForm.tsx
```diff
@@ -0,0 +1,182 @@
...
+            <option
+              key={item.username}
+              value={item.id}
+            >
+              {item.name}
+            </option>
+          ))}
+        </select>
+        {errors.isUserNotSelected
+        && <span className="error">Please choose a user</span>}
```
naziamoff, line 147: looks more clear this way I guess
```suggestion
        {errors.isUserNotSelected && (
          <span className="error">
            Please choose a user
          </span>
        )}
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,22 @@
...
+    data-id={todo.id}
+    className={`TodoInfo${todo.completed ? ' TodoInfo--completed' : ''}`}
+  >
+    <h2 className="TodoInfo__title">
+      {todo.title}
+    </h2>
+
+    {todo.user
+      ? <UserInfo user={todo.user} />
+      : <></>}
```
naziamoff, line 20: isn't this what you're trying to do here?
```suggestion
    {todo.user && (
      <UserInfo user={todo.user} />
    )}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,13 @@
...
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }: Props) => (
+  <section className="TodoList">
+    {todos.map(todo => <TodoInfo key={todo.id} todo={todo} />)}
```
naziamoff, line 11: better to make it multilined

naziamoff (changes requested):
bunch of code style suggestions. Also, the checkbox is still not clear to me. Let's do this in a more user-friendly way

src/components/AddTodoForm/AddTodoForm.tsx
UmizDemud: But I'm trying to list down all possible errors. Then to maybe turn it into its own folder for errors, for reusability. Just to say what I had in mind

naziamoff: 🤔

naziamoff: in this context you're right I guess.
Then I guess would be better to create separate validator for that and check all the fields on submit

naziamoff: or validate fields separately after they're touched

src/components/AddTodoForm/AddTodoForm.tsx
```diff
@@ -22,8 +22,8 @@ export const AddTodoForm = ({ handleSubmit, users }: Props) => {
   const [user, setUser] = useState(0);
   const [completed, setCompleted] = useState(false);
   const [errors, setErrors] = useState({
-    isUserNotSelected: false,
-    isTitleEmpty: false,
+    isUserSelected: true,
+    isTitle: true,
```
naziamoff, line 26: isTitleEmpty sounded good to me actually 😄

naziamoff (approved):

PR #1314, opened 2022-12-21
solution

oleksii-bieliaiev:
[DEMO LINK](https://oleksii-bieliaiev.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,1 +1,30 @@
-export const TodoInfo = () => {};
+import classNames from 'classnames';
+import { Todo } from '../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todo
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    user,
+    title,
+    completed,
+  } = todo;
+
+  return (
+    <article
+      data-id={todo.id}
+      className={classNames(
+        'TodoInfo',
+        {
+          'TodoInfo--completed': completed === true,
+        },
+      )}
+    >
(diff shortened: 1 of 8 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,30 @@
...
+type Props = {
+  todo: Todo
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    user,
+    title,
+    completed,
+  } = todo;
```
Anastasiia-Svintsova, line 14:
```suggestion
  const {
    user,
    title,
    completed,
    id
  } = todo;
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,30 @@
...
+
+  return (
+    <article
+      data-id={todo.id}
```
Anastasiia-Svintsova, line 18:
```suggestion
      data-id={id}
```

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,15 @@
...
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  return (
+    user && (
+      <a className="UserInfo" href={`mailto:${user.email}`}>
+        {user.name}
+      </a>
+    )
+  );
```
Anastasiia-Svintsova, line 14: You don't need to render this component if you don't have a user at all. So the user can't be null

Anastasiia-Svintsova (changes requested):
- Check if the title is not empty before submitting

https://user-images.githubusercontent.com/99403198/209119311-8a9d6f3f-15ce-4070-b979-d3930d69510c.mov

qoa11a (approved):

PR #1325, opened 2022-12-28
Solution

Julia2063:
[DEMO LINK](https://Julia2063.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,106 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-export const App = () => {
+import { TodoList } from './components/TodoList';
+
+function getUser(userId: number): User | undefined {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || undefined;
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,25 +1,106 @@
...
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { TodoList } from './components/TodoList';
+
+function getUser(userId: number): User | undefined {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || undefined;
+}
```
Anastasiia-Svintsova, line 13: Would be better to use null here 
```suggestion
function getUser(userId: number): User | null {
  const foundUser = usersFromServer.find(user => user.id === userId);

  return foundUser || null;
}
```

src/App.tsx
```diff
@@ -1,25 +1,106 @@
...
+                <option
+                  value={user.name}
+                  key={user.id}
+                >
+                  {user.name}
+                </option>
+              ))}
+            </select>
+            {(newTodoUserName === '' && needCheck)
+              && (<span className="error">Please choose a user</span>)}
```
Anastasiia-Svintsova, line 102:
```suggestion
            {!newTodoUserName.length && needCheck && (
              <span className="error">Please choose a user</span>
            )}
```

Anastasiia-Svintsova (changes requested):
- I shouldn't be able to add a todo with an empty title

https://user-images.githubusercontent.com/99403198/209941945-154476df-a085-454a-b28f-c8eef5d3aad1.mov

polosanya (approved):
Good job!

PR #1333, opened 2023-01-04
Develop

lefteris-L:
[DEMO LINK](https://lefteris-l.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,111 @@
+import React, { useState } from 'react';
+import { TodoList } from './components/TodoList';
+
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { User, Todo } from './types';
 import './App.scss';
 
-// import usersFromServer from './api/users';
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,61 +1,115 @@
...
+          user: getUserById(+selectedUser),
+          userId: +selectedUser,
+          completed: false,
+          title,
+        })
+      ));
+
+      setTitle('');
+      setUser('');
+    }
```
meforoma, line 47: wait, `.concat` mutates the state - we shouldn't do that
```suggestion
    if (title && selectedUser) {
      const newTodo = {
          id: generateId(),
          user: getUserById(+selectedUser),
          userId: +selectedUser,
          completed: false,
          title,
        }
      
      setTodos([
        ...todos,
        newTodo
      ]);

      setTitle('');
      setUser('');
    }
```

meforoma (changes requested):
all good
EXCEPT you mutated the state ⛔ 
immutability concept prohibits it.

PR #1335, opened 2023-01-05
Initial commit with solution

nurdanokcu:
[DEMO LINK](https://nurdanokcu.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,116 @@
+import React, { useState } from 'react';
 import './App.scss';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-export const App = () => {
+const getUserById = (userId: number) => usersFromServer
+  .find(user => user.id === userId)
+  || null;
+
+const todos = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [selectedUser, setSelectedUser] = useState('');
+  const [hasTitleError, setTitleError] = useState(false);
+  const [hasUserError, setUserError] = useState(false);
+  const [visibleTodos, setVisibleTodos] = useState(todos);
+
+  const createNewTodo = () => {
+    const maxId = Math.max(...visibleTodos.map(todo => todo.id));
+
+    return {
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,124 @@
...
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [selectedUser, setSelectedUser] = useState('');
+  const [hasTitleError, setTitleError] = useState(false);
+  const [hasUserError, setUserError] = useState(false);
+  const [visibleTodos, setVisibleTodos] = useState(todos);
+
+  const createNewTodo = () => {
+    const maxId = Math.max(...visibleTodos.map(todo => todo.id));
```
UmizDemud, line 25: This is relatively heavy. I think you can store the last ID in a variable just to avoid calculating it at each render. Therefore would be calculated only once.

UmizDemud (approved):
I'll approve to see the mentors advice too. Altough I'd go for the best practise and just implement it. Well done all around 👏

src/App.tsx
nurdanokcu: Thank you @UmizDemud . I agree with you. I tried to improve it now.

src/App.tsx
```diff
@@ -1,61 +1,126 @@
...
+}));
+
+let maxId = Math.max(...todos.map(todo => todo.id));
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [selectedUser, setSelectedUser] = useState('');
+  const [hasTitleError, setTitleError] = useState(false);
+  const [hasUserError, setUserError] = useState(false);
+  const [visibleTodos, setVisibleTodos] = useState(todos);
```
naziamoff, line 24:
```suggestion
  const [hasTitleError, setHasTitleError] = useState(false);
  const [hasUserError, setHasUserError] = useState(false);
  const [visibleTodos, setVisibleTodos] = useState(todos);
```

src/App.tsx
```diff
@@ -1,61 +1,126 @@
...
+          <label htmlFor="user">User: </label>
+          <select
+            data-cy="userSelect"
+            id="user"
+            value={selectedUser}
+            onChange={handleUserChange}
+          >
+            <option value="" disabled>Choose a user</option>
+
+            {usersFromServer.map(user => (
```
naziamoff, line 102:
```suggestion
            <option value="" disabled>Choose a user</option>
            {usersFromServer.map(user => (
```

naziamoff (approved):
looks good to me 👍

PR #1344, opened 2023-01-07
done

IoghiStudio:
[DEMO LINK](https://IoghiStudio.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,121 @@
+import React, { useState } from 'react';
 import './App.scss';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/todo';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { User } from './types/user';
 
-export const App = () => {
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,61 +1,121 @@
...
+}));
+
+export const App: React.FC<{}> = () => {
+  const [title, setTitle] = useState('');
+  const [currUser, selectUser] = useState('');
+  const [todos, setTodos] = useState(allTodos);
+  const [errorTitle, setErrorTitle] = useState(false);
+  const [errorUser, setErrorUser] = useState(false);
+
+  const getMaxId = () => {
```
UmizDemud, line 26: I advice you to use another method to find the max id, because sort is a mutating method, and it shouldn't be mutating the state like that. Check this out, should be cleaner:

```ts
return Math.max(...todos.map(todo => todo.id)) + 1;

UmizDemud (approved):
👍 

Just one comment

src/App.tsx
naziamoff: agree with @UmizDemud, that's critical one

naziamoff (approved):
well done
But pay attention, you should not mutate original array unless you really need to

PR #1358, opened 2023-01-10
solution

Sergii-Osedchuk:
[DEMO LINK](https://Sergii-Osedchuk.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,139 @@
 import './App.scss';
+import { useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+
+function findUserById(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+export const todoWithUser: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: findUserById(todo.userId),
+}));
+
+function findUserByName(userName: string): User | null {
+  return usersFromServer.find(user => user.name === userName) || null;
+}
 
 export const App = () => {
+  const [todos, setTodos] = useState(todoWithUser);
+  const [title, setTitle] = useState('');
+  const [selectedUser, setSelectedUser] = useState('');
+  const [isTitleError, setIsTitleError] = useState(false);
+  const [isUserNameError, setIsUserNameError] = useState(false);
+
+  const handleSubmit = (event: React.ChangeEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (selectedUser && title) {
+      setTodos(todosPrev => {
+        const maxTodoId = Math.max(...todosPrev.map(todo => todo.id));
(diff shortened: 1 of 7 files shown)


RedkoSerhii (approved):
Great job!

src/App.tsx
```diff
@@ -1,61 +1,139 @@
...
+  const [isTitleError, setIsTitleError] = useState(false);
+  const [isUserNameError, setIsUserNameError] = useState(false);
+
+  const handleSubmit = (event: React.ChangeEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (selectedUser && title) {
+      setTodos(todosPrev => {
+        const maxTodoId = Math.max(...todosPrev.map(todo => todo.id));
+        const newUser = findUserByName(selectedUser);
```
Ihor-Karpyn, line 39: What u will do if users will have the same name? Let's save user id in state

src/App.tsx
```diff
@@ -1,61 +1,139 @@
...
+      });
+    }
+
+    if (!selectedUser) {
+      setIsUserNameError(true);
+    }
+
+    if (!title) {
+      setIsTitleError(true);
+    }
```
Ihor-Karpyn, line 63: Let's make this validation in the top of this function and add return to remove if in addTodo block. Also check how we made validation on the lesson today

src/App.tsx
```diff
@@ -1,61 +1,139 @@
...
+  };
+
+  const handleChangeTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+    setIsTitleError(false);
+  };
+
+  const handleChangeSelectedUser
+    = (event: React.ChangeEvent<HTMLSelectElement>) => {
```
Ihor-Karpyn, line 72:
```suggestion
  const handleChangeSelectedUser = (
    event: React.ChangeEvent<HTMLSelectElement>
  ) => {
```

src/App.tsx
```diff
@@ -1,61 +1,139 @@
...
+    };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
+        action="/api/users"
+        method="POST"
```
Ihor-Karpyn, line 83: Redundant
```suggestion
```

Ihor-Karpyn (changes requested):
GJ, let's fix few issues

src/App.tsx
Sergii-Osedchuk: Fixed.

src/App.tsx
Sergii-Osedchuk: Changed validation method.

src/App.tsx
```diff
@@ -1,61 +1,132 @@
...
+
+      return;
+    }
+
+    if (selectedUserId && title) {
+      setTodos((todosPrev) => {
+        const maxTodoId = Math.max(...todosPrev.map(todo => todo.id));
+
+        setTitle('');
+        setSelectedUserId('');
```
valerii-derkach, line 56: use number for Id,

valerii-derkach (changes requested):

src/App.tsx
Sergii-Osedchuk: Fixed!

Ihor-Karpyn (approved):

PR #1367, opened 2023-01-11
Add todo form

DiDemkiv:


src/App.tsx
```diff
@@ -1,61 +1,107 @@
...
+          title,
+          userId: +userId,
+          completed: false,
+          user: getUser(+userId),
+        };
+
+        setStoredTodos([...storedTodos, newTodo]);
+
+        setTitle('');
+        setUserId('0');
```
Anastasia4sik, line 61: I think you should create a function and pass it onSubmit
```suggestion
      <form onSubmit={addTodo}>
```

src/App.tsx
```diff
@@ -1,61 +1,107 @@
...
+          <input
+            name="title"
+            value={title}
+            type="text"
+            placeholder="Enter the title"
+            data-cy="titleInput"
+            onChange={(event) => {
+              setShowTitleError(false);
+              setTitle(event.target.value);
+            }}
```
Anastasia4sik, line 74: I think you should create a function and pass it onChange
Fix it in the hole code

src/App.tsx
```diff
@@ -1,61 +1,107 @@
...
+            data-cy="userSelect"
+            value={userId}
+            onChange={(event) => {
+              setShowUserError(false);
+              setUserId(event.target.value);
+            }}
+          >
             <option value="0" disabled>Choose a user</option>
+            {usersFromServer.map(user => (
+              <option value={user.id} key={user.id}>{user.name}</option>
```
Anastasia4sik, line 92: Maybe fix code style here
```suggestion
              <option
              	value={user.id}
              	key={user.id}
              >
              	{user.name}
              </option>
```

Anastasia4sik (changes requested):
I left you some advice

src/App.tsx
DiDemkiv: I don't think that's necessary to move two lines of code :)

Anastasia4sik (approved):

src/App.tsx
```diff
@@ -1,61 +1,116 @@
...
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+
+function getUser(userId: number): User | null {
```
solaryasha, line 10:
```suggestion
function getUserById(userId: number): User | null {
```

src/App.tsx
```diff
@@ -1,61 +1,116 @@
...
+}));
 
 export const App = () => {
+  const [storedTodos, setStoredTodos] = useState(todos);
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState('0');
+  const [showUserError, setShowUserError] = useState(false);
+  const [showTitleError, setShowTitleError] = useState(false);
+
+  const submitTodoForm = () => {
```
solaryasha, line 28:
```suggestion
  const handleFormSubmit = () => {
```

src/App.tsx
```diff
@@ -1,61 +1,116 @@
...
+    let shouldExit = false;
+
+    if (userId === '0') {
```
solaryasha, line 31:
```suggestion
    if (Number(userId) === 0) {
```

src/App.tsx
```diff
@@ -1,61 +1,116 @@
...
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [storedTodos, setStoredTodos] = useState(todos);
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState('0');
```
solaryasha, line 24: Let's store numbers in the state

src/App.tsx
```diff
@@ -1,61 +1,116 @@
...
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form onSubmit={(event) => {
+        event.preventDefault();
+        submitTodoForm();
+      }}
+      >
```
solaryasha, line 67: Avoid using inline functions . You should pass an event to the function
```suggestion
      <form onSubmit={submitTodoForm}>
```

src/App.tsx
solaryasha: >I don't think that's necessary to move two lines of code :)

Why do you think so?  What if these two lines become three lines or more?
Let' create a function for better readibility

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,26 @@
-export const TodoInfo = () => {};
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todo,
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
```
solaryasha, line 9: What about `React.memo` ?

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,16 @@
-export const TodoList = () => {};
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: Todo[],
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
```
solaryasha, line 8: What about `React.memo` ?

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,15 @@
-export const UserInfo = () => {};
+import { User } from '../../types/User';
+
+type Props = {
+  user: User,
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
```
solaryasha, line 7: What about `React.memo` ?

solaryasha (changes requested):

PR #1368, opened 2023-01-11
todo list

smiyka:
[DEMO LINK](https://smiyka.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,130 @@
+import React, { useState } from 'react';
 import './App.scss';
+import { TodoList } from './components/TodoList';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+function getUser(userId: number): User | null {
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,61 +1,132 @@
...
+          <label>
+            User:
+            <select
+              data-cy="userSelect"
+              value={selectedUser}
+              onChange={handleChangeUser}
+            >
+              <option
+                value=""
+              >
```
tetiana-ivanchenko, line 103:
```suggestion
              <option
                value=""
                disabled
              >
```

tetiana-ivanchenko (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+            User:
+            <select
+              data-cy="userSelect"
+              value={selectedUser}
+              onChange={handleChangeUser}
+            >
+              <option
+                value=""
+                disabled
+              >
```
vlad-demchuk, line 104: did you deploy last changes?

after changes `choose a user` option should be disabled
image

vlad-demchuk (changes requested):

vlad-demchuk (approved):

PR #1370, opened 2023-01-11
solution

oleksandrnikora:
[DEMO LINK](https://oleksandrnikora.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,120 @@
 import './App.scss';
+import React, { useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoWithUser } from './types/TodoWithUser';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+const getUserById = (id: number) => {
+  return usersFromServer.find(user => user.id === id) || null;
+};
 
-export const App = () => {
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [todos, setTodos] = useState<Todo[]>(todosFromServer);
+  const [isTitleEmpty, setIsTitleEmpty] = useState(false);
+  const [isUserNotSelected, setIsUserNotSelected] = useState(false);
+
+  const changeTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+    setIsTitleEmpty(false);
+  };
+
+  const changeUser = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setUserId(+event.target.value);
+    setIsUserNotSelected(false);
+  };
+
+  const todosWithUsers: TodoWithUser[] = todos
+    .map(todo => ({
+      ...todo,
+      user: getUserById(todo.userId) || null,
+    }));
+
+  const maxId: number = todosWithUsers.reduce((max, currentUser) => {
+    return max >= currentUser.id
+      ? max
+      : currentUser.id;
+  }, 0);
+
+  const newTodo: Todo = {
+    id: maxId + 1,
+    title,
+    completed: false,
+    userId,
+  };
+  const clearForm = () => {
+    setTitle('');
+    setUserId(0);
+  };
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (userId && title.trim()) {
+      setTodos((currentTodos) => [...currentTodos, newTodo]);
+      clearForm();
+      setIsTitleEmpty(false);
+      setIsUserNotSelected(false);
+    }
+
+    setIsTitleEmpty(!title.trim());
+    setIsUserNotSelected(!userId);
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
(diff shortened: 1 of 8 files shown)


TetKopot (changes requested):
good job! but could you also add some styles?

TetKopot (approved):
great!

src/App.tsx
```diff
@@ -1,61 +1,147 @@
...
+    setIsTitleEmpty(false);
+  };
+
+  const changeUser = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setUserId(+event.target.value);
+    setIsUserNotSelected(false);
+  };
+
+  const todosWithUsers: TodoWithUser[] = todos
+    .map(todo => ({
```
Ihor-Karpyn, line 32:
```suggestion
  const todosWithUsers: TodoWithUser[] = todos.map(todo => ({
```

src/App.tsx
```diff
@@ -1,61 +1,147 @@
...
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+const getUserById = (id: number) => {
+  return usersFromServer.find(user => user.id === id) || null;
+};
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [todos, setTodos] = useState<Todo[]>(todosFromServer);
```
Ihor-Karpyn, line 17: U have to set in state todos with user but not todos from server

src/App.tsx
```diff
@@ -1,61 +1,147 @@
...
+    setTitle(event.target.value);
+    setIsTitleEmpty(false);
+  };
+
+  const changeUser = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setUserId(+event.target.value);
+    setIsUserNotSelected(false);
+  };
+
+  const todosWithUsers: TodoWithUser[] = todos
```
Ihor-Karpyn, line 31: Move this function outside the component, and set the result as the default value in the state

src/App.tsx
```diff
@@ -1,61 +1,147 @@
...
+      ? max
+      : currentUser.id;
+  }, 0);
+
+  const newTodo: Todo = {
+    id: maxId + 1,
+    title,
+    completed: false,
+    userId,
+  };
```
Ihor-Karpyn, line 48: U have to do this inside handleSubmit and get max id inside functional setState

Ihor-Karpyn (changes requested):
I am sure that you could do better, you could try material UI to make better UI

image

src/App.tsx
```diff
@@ -1,61 +1,146 @@
...
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [todos, setTodos] = useState(todosWithUsers);
+  const [isTitleEmpty, setIsTitleEmpty] = useState(false);
+  const [isUserNotSelected, setIsUserNotSelected] = useState(false);
+
+  const changeTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+    setIsTitleEmpty(false);
```
solaryasha, line 34: Why do we need to set this variable to false all the time? Maybe, let's do it when `isTitleEmpty` is true ?

src/App.tsx
```diff
@@ -1,61 +1,146 @@
...
+  };
+
+  const changeUser = (event: SelectChangeEvent<number>) => {
+    setUserId(+event.target.value);
+    setIsUserNotSelected(false);
```
solaryasha, line 39: Same here.

src/App.tsx
```diff
@@ -1,61 +1,146 @@
...
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    setIsTitleEmpty(!title.trim());
+    setIsUserNotSelected(!userId);
+
+    if (userId && title.trim()) {
+      setTitle('');
+      setUserId(0);
```
solaryasha, line 50: It's dangerous, you should do it after creating the new todo

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,34 @@
-export const TodoInfo = () => {};
+import cn from 'classnames';
+import { TodoWithUser } from '../../types/TodoWithUser';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: TodoWithUser,
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
```
solaryasha, line 9: Let's use React.memo

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,14 @@
-export const TodoList = () => {};
+import { TodoWithUser } from '../../types/TodoWithUser';
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: TodoWithUser[],
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
```
solaryasha, line 8: Let's use React.memo

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,15 @@
-export const UserInfo = () => {};
+import { User } from '../../types/User';
+
+type Props = {
+  user: User,
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
```
solaryasha, line 7: Let's use React.memo

solaryasha (approved):
👍

PR #1379, opened 2023-01-11
initial commit, uncompleted solution

DimaKolesnyk:


Mykhailyk-Vadym (approved):
Looks good

src/App.tsx
```diff
@@ -1,61 +1,120 @@
...
+import { Todo } from './Types/Todo';
+
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+function getUser(userId: number): User | null {
```
valerii-derkach, line 12:
```suggestion
function getUserById(userId: number): User | null {
```

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
+              className="input is-small w-max"
+              id="title"
+              type="text"
+              data-cy="titleInput"
+              placeholder="Enter a title"
+              value={title}
+              onChange={handleTitleChange}
+            />
+            {!isTitleError
+              && <span className="noError">Please enter a title</span>}
```
valerii-derkach, line 99: no need to render hidden content, looks wired, just remove it

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
+              <option value="" disabled>Choose a user</option>
+
+              {usersFromServer.map(user => (
+                <option key={user.id}>{user.name}</option>
+              ))}
+            </select>
+            {isUserError
+              && <span className="error">Please choose a user</span>}
+            {!isUserError
+              && <span className="noError">Please choose a user</span>}
```
valerii-derkach, line 127: remove it

src/App.tsx
valerii-derkach: it is better to specify our function - be get user by id here, so name it correctly `getUserById`

valerii-derkach (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,139 @@
...
+              User:
+            </label>
+            <select
+              className="input is-small"
+              data-cy="userSelect"
+              id="userSelect"
+              value={selectedUser}
+              onChange={handleUserChange}
+            >
+              <option value="" disabled>Choose a user</option>
```
valerii-derkach, line 116:
```suggestion
              <option value="" disabled>Choose a user</option>
```

do not left empty value

valerii-derkach (approved):

PR #1382, opened 2023-01-11
solution

Cardlice:
[DEMO LINK](https://Cardlice.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,119 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-export const App = () => {
+import { TodoList } from './components/TodoList';
+
+import { Todo } from './types/Todo';
+import { User } from './types/User';
+
+function getUserById(userId: number): User | null {
+  return usersFromServer.find(user => user.id === userId) || null;
+}
+
+const todosWithUser: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [todos, setTodos] = useState(todosWithUser);
+  const [newTitle, setNewTitle] = useState('');
+  const [newUserId, setNewUserId] = useState(0);
+  const [showTitleError, setTitleError] = useState(false);
+  const [showUserError, setUserError] = useState(false);
+
+  const getNewId = (array: { id: number }[]) => (
+    Math.max(...array.map(el => el.id)) + 1
+  );
+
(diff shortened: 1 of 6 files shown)


OlggaJul (approved):

src/App.tsx
```diff
@@ -1,61 +1,153 @@
...
+            sx={{ m: 2, width: '250px' }}
+            data-cy="titleInput"
+            label="Title"
+            placeholder="Enter a title"
+            helperText={showTitleError ? 'Please enter a title' : ''}
+            error={showTitleError}
+            value={newTitle}
+            onChange={event => {
+              setNewTitle(event.target.value);
+              setTitleError(false);
```
solaryasha, line 99: Why do we need to set this variable to false all the time ? Maybe, let's do it when `showTitleError` is true ?

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,17 @@
-export const TodoList = () => {};
+import React from 'react';
+import { TodoInfo } from '../TodoInfo';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todos: Todo[],
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
```
solaryasha, line 9: Let's use `React.memo`

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,15 @@
-export const UserInfo = () => {};
+import React from 'react';
+import { User } from '../../types/User';
+import './UserInfo.scss';
+
+type Props = {
+  user: User,
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
```
solaryasha, line 9: Let's use `React.memo`

solaryasha (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,163 @@
...
 
-export const App = () => {
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
+        action="/api/users"
+        method="POST"
```
valerii-derkach, line 101:
```suggestion

```

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,15 @@
...
+import React, { memo } from 'react';
+import { User } from '../../types/User';
+import './UserInfo.scss';
+
+type Props = {
+  user: User,
+};
+
+export const UserInfo: React.FC<Props> = memo(({ user }) => {
+  return (
```
valerii-derkach, line 10:
```suggestion
  return (
```
if the component has no body - return is redundant
or destructure your props before return

valerii-derkach (approved):

PR #1387, opened 2023-01-12
Solution React Add TODO Form

volodymir-tymtsias:
[DEMO LINK](https://volodymir-tymtsias.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -24,26 +24,8 @@ function getNewTodoId(todos: Todo[]): number {
   return Math.max(...todos.map(todo => todo.id)) + 1;
 }
 
-function getVisibleTodos(
-  todos: Todo[],
-  selectedUser: string,
-  title: string,
-): Todo[] {
-  const todoId = getNewTodoId(todos);
-  const userId = getUserId(selectedUser);
-  const todo = {
-    id: todoId,
-    userId,
-    title: title.trim(),
-    completed: false,
-    user: getUser(userId),
-  };
-
-  return [...todos, todo];
-}
-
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,61 +1,139 @@
...
+  const userId = getUserId(selectedUser);
+  const todo = {
+    id: todoId,
+    userId,
+    title: title.trim(),
+    completed: false,
+    user: getUser(userId),
+  };
+
+  return [...todos, todo];
```
vlad-demchuk, line 42: logic for adding todo should be inside a react component

src/App.tsx
```diff
@@ -1,61 +1,139 @@
...
+            setErrorUser(true);
+            error = true;
+          }
+
+          if (!error) {
+            setTodos(getVisibleTodos(todos, selectedUser, title));
+            setTitle('');
+            setSelectedUser('Choose a user');
+          }
+        }}
```
vlad-demchuk, line 88: use a separate handle function

src/App.tsx
```diff
@@ -1,61 +1,139 @@
...
+            <input
+              type="text"
+              name="title"
+              placeholder="Enter a title"
+              data-cy="titleInput"
+              value={title}
+              onChange={(event) => {
+                if (inputControl(event)) {
+                  setTitle(event.target.value);
+                  setErrorTitle(false);
```
vlad-demchuk, line 102: the same, better to create a separate functions to make jsx clearer

vlad-demchuk (changes requested):

nazarmatsevych (approved):

PR #1388, opened 2023-01-13
Develop

panianchuk:
https://panianchuk.github.io/react_add-todo-form/


SanyaBratashchuk (changes requested):
the form should not reset if the submission does not pass validation

SanyaBratashchuk (approved):

PR #1400, opened 2023-01-16
sol

dolnaro:
[DEMO LINK](https://dolnaro.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,107 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-export const App = () => {
+import { TodoList } from './components/TodoList';
+
+function getUser(userId: number): User | undefined {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || undefined;
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+  const [newTodoUserName, setNewTodoUserName] = useState('');
+  const [renderedTodos, setRenderedTodos] = useState(todos);
+  const [needCheck, setNeedCheck] = useState(false);
+
+  const handleChangeInput = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = event.target;
+    const rightValue = value.replace(/[^a-zA-Z0-9 ]/g, '');
+
+    setNewTodoTitle(rightValue);
+  };
+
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,25 +1,109 @@
...
+  }));
+
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+  const [newTodoUserName, setNewTodoUserName] = useState('');
+  const [renderedTodos, setRenderedTodos] = useState(todos);
+  const [needCheck, setNeedCheck] = useState(false);
+
+  const handleChangeInput = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = event.target;
+    const rightValue = value.trim().length > 0
```
mmahfuzi, line 31: You don't have to check for this condition actually, if the length is 0, replace will leave it. So this should work also:

```js
const rightValue = value.replace(/[^a-zA-Z0-9 ]/g, '');

**Review by mmahfuzi — APPROVED** [review]
🔥

File: src/App.tsx [hunk 2]
```diff
@@ -1,25 +1,109 @@
...
+import { TodoList } from './components/TodoList';
+
+export const App: React.FC = () => {
+  const getUser = (param: number | string): User | null => {
+    const foundUser
+    = typeof param === 'number'
+      ? usersFromServer.find(user => user.id === param)
+      : usersFromServer.find(user => user.name === param);
+
+    return foundUser || null;
```
naziamoff, line 16:
```suggestion
    const foundUser = usersFromServer.find(user => (
       typeof param === 'number'
         ? user.id === param
         : user.name === param
    ))
    
    return foundUser || null;
```

naziamoff: but anyway it does not seem to be logical - looking for the user by their name. Better to stick to id, cause anyway every user has it

src/App.tsx
```diff
@@ -1,25 +1,109 @@
...
+
+  const todos: Todo[] = todosFromServer.map(todo => ({
+    ...todo,
+    user: getUser(todo.userId),
+  }));
+
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+  const [newTodoUserName, setNewTodoUserName] = useState('');
+  const [renderedTodos, setRenderedTodos] = useState(todos);
+  const [needCheck, setNeedCheck] = useState(false);
```
naziamoff, line 27: what's need check? Unclear to me

naziamoff (changes requested):

src/App.tsx
```diff
@@ -1,25 +1,108 @@
...
+    setNewTodoUserName(value);
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+    const newTodoUser: User | null = getUser(newTodoUserName);
+    const newTodo: Todo = {
+      id: Math.max(...renderedTodos.map(todo => todo.id)) + 1,
+      title: newTodoTitle,
+      userId: newTodoUser ? newTodoUser.id : null,
```
naziamoff, line 49: can userId be null?

naziamoff (approved):

PR #1402, opened 2023-01-16
Develop

sercansimsek:


lefteris-L (approved):
LGTM ✔️

nurdanokcu (approved):
Well done.

src/App.tsx
```diff
@@ -1,61 +1,129 @@
...
+    setTitle(event.target.value);
+    setHasTitleError(false);
+  };
+
+  const selectedUserHandler = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setselectedUserId(+event.target.value);
+    setHasUserError(false);
+  };
+
+  const buttonHandler = (event: React.FormEvent<HTMLFormElement>) => {
```
naziamoff, line 46: you should not name your methods `handlers`.
It should be clear what they do from their name. It should start with a verb
```suggestion
  const changeTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
    setTitle(event.target.value);
    setHasTitleError(false);
  };

  const selectUser= (event: React.ChangeEvent<HTMLSelectElement>) => {
    setselectedUserId(+event.target.value);
    setHasUserError(false);
  };

  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
```

src/App.tsx
```diff
@@ -1,61 +1,129 @@
...
+    event.preventDefault();
+
+    setHasTitleError(!title);
+    setHasUserError(!selectedUserId);
+
+    if (title && selectedUserId) {
+      setCurrentTodos([
+        ...currentTodos,
+        getNewTodo(),
```
naziamoff, line 55: this should be stored into a separate variable instead of calling it on submit. What's more, from its name I guess that it somewhere gets the todo and it is not about input.

naziamoff (changes requested):
not bad at all. There are a couple of suggestions though, need polishing

naziamoff (approved):

PR #1405, opened 2023-01-17
solution

Iksich:
[DEMO LINK](https://Iksich.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,114 @@
+import { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList/TodoList';
+import { User } from './components/types/User';
+import { Todo } from './components/types/Todo';
 
-export const App = () => {
+function getUser(userId: number): User | null {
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [titleError, setTitleError] = useState(false);
+  const [userIdError, setUserIdError] = useState(false);
+
+  const handleSumbit = (event:React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (userId === 0 || title === ' ') {
```
naziamoff, line 30: - the `0` looks kinda hardcoded. I'd say that it is better to store null there and to check whether the userId is `truthy`
- what is `title === ' '`? To me it seems to be very similar to the `trim()` method

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+    setTitle('');
+    setUserId(0);
+  };
+
+  const handleTitle = (event:React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+    setTitleError(false);
+  };
+
+  const handleUser = (event:React.ChangeEvent<HTMLSelectElement>) => {
```
naziamoff, line 62: same here

PR #1407, opened 2023-01-17
finsihed everything there is one weird failed test

alaaashkar:
[DEMO LINK](https://alaaashkar.github.io/react_add-todo-form/)

please help me find the failed test because everything is functioning well ......

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,119 @@
 import './App.scss';
+import React, { useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+function getUser(userId: number) {
+  const foundUser = usersFromServer.find(user => user.id === userId);
 
-export const App = () => {
+  return foundUser || null;
+}
+
+const newTodos = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+let maxId = Math.max(...newTodos.map(todo => todo.id));
+
+export const App: React.FC = () => {
+  const [activeUser, setActiveUser] = useState(0);
+  const [message, setMessage] = useState('');
+  const [visibleTodos, setVisibleTodos] = useState(newTodos);
+  const [hasMessageError, setHasMessageError] = useState(false);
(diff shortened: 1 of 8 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,22 @@
...
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => (
+  <article className={classNames(
+    'TodoInfo',
+    {
+      'TodoInfo--completed': todo.completed,
+    },
+  )}
+  >
```
CharAnagn, line 17: This one should have data-id attribute equal to the id that you pass from the App. This should work I think with the tests.

CharAnagn (changes requested):
Your code seems fine so I think this is the problem with the tests. Hope it works 🙏

src/App.tsx
```diff
@@ -1,61 +1,119 @@
...
+      completed: false,
+      userId: activeUser,
+      user: getUser(activeUser),
+    };
+  };
+
+  const handleAdd = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    setHasMessageError(!message);
+    setHasUserError(!activeUser);
```
ysrckr, line 42: You are setting them here. What happens if I start to write in the input field and don't click submit?

src/App.tsx
```diff
@@ -1,61 +1,119 @@
...
+    setActiveUser(0);
+  };
 
-export const App = () => {
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form action="/api/users" method="Get" onSubmit={handleAdd}>
```
ysrckr, line 61: You don't need action and method, you are not sending or getting anything from the server, anyway you are preventing the default behavior.

ysrckr (changes requested):
Couple of thing for you to look at.

src/App.tsx
alaaashkar: I really do not understand the comment, may you share the solution please ?

ysrckr: [screenshot](https://dsc.cloud/2f6ec4/Screenshot-2023-01-27-at-15.21.01.png)

ysrckr: You only set error on submit but you don't remove to error until it is submitted again.

PR #1411, opened 2023-01-23
final

Alireza-Mohammadhossein:
[DEMO LINK](https://alireza-mohammadhossein.github.io/react_add-todo-form/)

Everything is complete, I don't know how the error appears

### DIFF
--- a/src/components/UserInfo/UserInfo.tsx
+++ b/src/components/UserInfo/UserInfo.tsx
@@ -1,7 +1,5 @@
 import React from 'react';
 
-import './userinfo.scss';
-
 import { User } from '../../types/User';
 
 type Props = {


src/App.tsx
```diff
@@ -1,61 +1,125 @@
...
+
+  function addTodo(id: number) {
+    if (getUser(id) && title.trim()) {
+      const newTodo = {
+        id: newId,
+        userId,
+        title,
+        completed: false,
+        user: getUser(userId),
+      };
```
ysrckr, line 41: you are take in user id as a parameter than you check if the user exists. Then you use userId in the state. If you are going to use one in state, do you really need to take the id as a parameter or vice versa?

src/App.tsx
```diff
@@ -1,61 +1,125 @@
...
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [userId, setUserId] = useState(0);
+  const [title, setTitle] = useState('');
+  const [currentTodos, setCurrentTodos] = useState(todos);
+  const [isUserSelected, setIsUserSelected] = useState(true);
+  const [isTitleEntered, setIsTitleEntered] = useState(true);
```
ysrckr, line 29: Do you really need to keep track of title and if the title is entered? You can get its boolean value instead.

src/App.tsx
```diff
@@ -1,61 +1,125 @@
...
+
+      setUserId(0);
+      setTitle('');
+      setCurrentTodos([...currentTodos, newTodo]);
+    }
+
+    if (!userId) {
+      setUserId(0);
+      setIsUserSelected(false);
+    }
```
ysrckr, line 51: Why do you need this?

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,40 @@
...
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    completed,
+    title,
+    user,
+  } = todo;
+
+  return (
+    <article
+      data-id={todo.id}
```
ysrckr, line 21: You destructered the todo while doing it, why don't you have the id there too?

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,40 @@
...
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    completed,
+    title,
+    user,
```
ysrckr, line 16: You are using id too so you can add it here too

src/components/TodoInfo/TodoInfo.tsx
Alireza-Mohammadhossein: I forgot it :(

naziamoff (approved):
🔥

PR #1414, opened 2023-01-26
initial commit

Malva37:
[DEMO LINK](https://Malva37.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,60 +1,143 @@
 import './App.scss';
+import React, { useState } from 'react';
+// import React from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { FullTodo, Todo } from './types/Todo';
+import { User } from './types/User';
+import { TodoList } from './components/TodoList';
 
-export const App = () => {
+const prepareTodos = (
+  users: User[],
+  todos: Todo[],
+): FullTodo[] => {
+  return todos.map((todo) => ({
+    ...todo,
+    user: users.find((user) => user.id === todo.userId) || null,
+  }));
+};
+
+const todos = prepareTodos(usersFromServer, todosFromServer);
+const users = usersFromServer;
+
+export const App: React.FC = () => {
+  const [userFromSelect, setUser] = useState('0');
+  const [userError, setUserError] = useState(false);
+  const [titleError, setTitleError] = useState(false);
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,60 +1,143 @@
...
+
+const todos = prepareTodos(usersFromServer, todosFromServer);
+const users = usersFromServer;
+
+export const App: React.FC = () => {
+  const [userFromSelect, setUser] = useState('0');
+  const [userError, setUserError] = useState(false);
+  const [titleError, setTitleError] = useState(false);
+  const [title, setTitle] = useState('');
+  const [todosArr, setTodo] = useState(todos);
```
nzrii-kysel, line 29: names of state variables should match their setter names

src/App.tsx
```diff
@@ -1,60 +1,143 @@
...
+        user: currentUser,
+      };
+
+      setTodo(current => ([
+        ...current,
+        todo,
+      ]));
+
+      setTitle('');
+      setUser('0');
```
nzrii-kysel, line 77: **suggestion**
i'd move this whole block into a separate function for readability

src/App.tsx
```diff
@@ -1,60 +1,143 @@
...
+
+  const handleChangeTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitleError(false);
+    const { value } = event.target;
+
+    setTitle(value);
+  };
+
+  const handleAddTodo = () => {
+    const currentUser = users
```
nzrii-kysel, line 50: **suggestion**
i dont think you need handlers on both submit button and form. since they always trigger at the same time, you could merge `handleAddTodo`  and `handleSubmit` into one  handler

nzrii-kysel (changes requested):
good job
just need to fix variable names in state

src/App.tsx
Malva37: moved part of this block, thanks

src/App.tsx
```diff
@@ -1,25 +1,137 @@
...
+    id: maxId + 1,
+    title,
+    completed: false,
+    userId: currentUser.id,
+    user: currentUser,
+  });
+};
+
+const visibleTodos = prepareTodos(usersFromServer, todosFromServer);
+const users = usersFromServer;
```
Anna-Tkachenko, line 37: do you need this additional variable?

Anna-Tkachenko (approved):
Good job! ⭐️

PR #1415, opened 2023-01-27
Dev

nzrii-kysel:
[DEMO LINK](https://nzrii-kysel.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,108 @@
+import { FormEvent, useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import users from './api/users';
+import todosFromServer from './api/todos';
+
+import { TodoList } from './components/TodoList';
 
 export const App = () => {
+  const [inputTitle, setInputTitle] = useState('');
+  const [selectedUserName, setSelectedUserName] = useState('');
+  const [hadFailedSubmit, setHadFailedSubmit] = useState(false);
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,61 +1,117 @@
...
+import users from './api/users';
+import todosFromServer from './api/todos';
+
+import { TodoList } from './components/TodoList';
 
 export const App = () => {
+  const [inputTitle, setInputTitle] = useState('');
+  const [selectedUserName, setSelectedUserName] = useState('');
+  const [hadFailedSubmit, setHadFailedSubmit] = useState(false);
+  const [todos, setTodos] = useState(
```
Malva37, line 13: not sure that it's good practice set initial value in that way

Malva37 (approved):
Good)

src/App.tsx
```diff
@@ -1,61 +1,117 @@
...
+            <option value="" disabled>Choose a user</option>
+            {users.map(({ name, id }) => (
+              <option key={id} value={name}>
+                {name}
+              </option>
+            ))}
           </select>
 
-          <span className="error">Please choose a user</span>
+          {(hadFailedSubmit && !selectedUserName) && (
```
Anna-Tkachenko, line 104: will be better to create separate boolean variable before the `return` statement with the result of this check
and use those variable in `jsx`

Anna-Tkachenko (approved):
Well Done! ⭐️

PR #1416, opened 2023-01-28
React Add TODO Form - completed

GarryRocksman:
[DEMO](https://garryrocksman.github.io/react_add-todo-form/)

 - Have a nice Day

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,127 @@
 import './App.scss';
+import React, { useState } from 'react';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
 
-export const App = () => {
+function getUser(userId: number): User | null {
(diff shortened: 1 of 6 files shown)


svitjojo (approved):
I like it 👍

src/App.tsx
```diff
@@ -1,61 +1,127 @@
...
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [todosTitle, setTodoTitle] = useState('');
+  const [todosUserId, setTodosUserId] = useState(-1);
```
alexandra-protyanova, line 23: make sense to make initial value `0` which is false,  instead of `-1` p.s `id` is never 0 by default

src/App.tsx
```diff
@@ -1,61 +1,127 @@
...
+        title: todosTitle,
+        completed: false,
+        user: getUser(todosUserId),
+      };
+
+      todos.push(todo);
+
+      setTodoTitle('');
+      setTodosUserId(-1);
+    } else {
```
alexandra-protyanova, line 45: you may not use `else` here

alexandra-protyanova (approved):
Good Job 👍

PR #1429, opened 2023-02-03
react / add todo form

valntyn:
[DEMO LINK](https://valntyn.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,152 @@
+import React, { useState } from 'react';
 import './App.scss';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
 
-export const App = () => {
+import { TodoList } from './components/TodoList';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+const todoWithUsers: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState('');
+
+  const [languageError, setlanguageError] = useState(false);
+  const [isTitleError, setTitleError] = useState(false);
+  const [isUserError, setUserError] = useState(false);
+
+  const [visibleTodos, setVisibleTodos] = useState(todoWithUsers);
+
+  let maxId = Math.max(...todoWithUsers.map(todo => todo.id));
+
+  const createTodo = () => {
+    maxId += 1;
+
+    return ({
+      id: maxId,
+      title,
+      completed: false,
+      userId: +userId,
+      user: getUser(+userId),
+    });
+  };
+
+  function checkValue(inputValue: string) {
+    const reg = /[a-zA-Zа-яА-ЯГЄІЇ\d\s]/g;
+    const matches = inputValue.match(reg);
+
+    if (matches === null) {
+      setlanguageError(true);
+    }
+
+    return matches ? matches.join('') : '';
+  }
+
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,61 +1,160 @@
...
+      userId: +userId,
+      user: getUser(+userId),
+    });
+  };
+
+  function checkValue(inputValue: string) {
+    const reg = /[a-zA-Zа-яА-ЯГЄІЇ\d\s]/g;
+    const matches = inputValue.match(reg);
+
+    if (matches === null) {
```
Anastasiia-Svintsova, line 50: You can also use a logical operator here 
```suggestion
    if (!matches) {
```

src/App.tsx
```diff
@@ -1,61 +1,160 @@
...
+      setlanguageError(true);
+    }
+
+    return matches ? matches.join('') : '';
```
Anastasiia-Svintsova, line 54: Also will work. You can check it
```suggestion
    return matches?.join('') || '';
```

src/App.tsx
```diff
@@ -1,61 +1,160 @@
...
+      setTitleError(false);
+      setlanguageError(false);
+
+      setTitle(checkValue(value));
+    }
+
+    if (e.target.id === 'select') {
+      setUserId(value);
+      setUserError(false);
+    }
```
Anastasiia-Svintsova, line 72: You can use switch/case here

src/App.tsx
```diff
@@ -1,61 +1,160 @@
...
+    const matches = inputValue.match(reg);
+
+    if (matches === null) {
+      setlanguageError(true);
+    }
+
+    return matches ? matches.join('') : '';
+  }
+
+  type Event = React.ChangeEvent<HTMLInputElement | HTMLSelectElement>;
```
Anastasiia-Svintsova, line 57: Move this type outside the component. And specify the name, please. Not just Event

src/App.tsx
```diff
@@ -1,61 +1,160 @@
...
+              data-cy="titleInput"
+              placeholder="Enter a title"
+              value={title}
+              onChange={e => handleField(e)}
+            />
+            {languageError
+              && (
+                // eslint-disable-next-line max-len
+                <span className="error">The title supports only EN and UA letters and numbers</span>
+              )}
```
Anastasiia-Svintsova, line 117: Don't disable eslint, try to fix the error
```suggestion
            {languageError && (
              <span className="error">
                The title supports only EN and UA letters and numbers
              </span>
            )}
```

src/App.tsx
```diff
@@ -1,61 +1,160 @@
...
-            <option value="0" disabled>Choose a user</option>
-          </select>
+          <label>
+            User:
+            {' '}
+            <select
+              data-cy="userSelect"
+              value={userId}
+              id="select"
+              onChange={e => handleField(e)}
```
Anastasiia-Svintsova, line 133: Also will work. Event goes to the arguments by default 
```suggestion
              onChange={handleField}
```

Anastasiia-Svintsova (changes requested):
- Now I can add a todo with an empty title. Fix it, please

https://user-images.githubusercontent.com/99403198/216764618-bd281031-69ce-4dd9-8a9d-6f80c203f5cf.mov

src/App.tsx
valntyn: thank you, has been fixed

src/App.tsx
valntyn: thank you, has been fixed

src/App.tsx
valntyn: thank you, has been fixed

src/App.tsx
valntyn: thank you, has been fixed

src/App.tsx
valntyn: thank you, has been fixed

src/App.tsx
valntyn: thanks for the good review and for your advice

nazarmatsevych (approved):

PR #1431, opened 2023-02-03
Solution

KrisMakarovska:
[[DEMO LINK]](https://KrisMakarovska.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,118 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-export const App = () => {
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList/TodoList';
+
+const todosWithUser: Todo[] = todosFromServer.map(todo => {
+  return {
+    ...todo,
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,61 +1,118 @@
...
+
+  const resetForm = () => {
+    setTitleInput('');
+    setUserSelect('');
+    setFormChecked(false);
+  };
+
+  const addNewTodo = () => {
+    const selectedUser = usersFromServer
+      .find(user => user.name === userSelect) as User;
```
Anastasiia-Svintsova, line 34: What will happen if the method doesn't find anything? You can get undefined here. Add an `if` statement for the case when there is no selectedUser.
```suggestion
    const selectedUser = usersFromServer
      .find(user => user.name === userSelect);
      
    if (!selectedUser) ...
```

src/App.tsx
```diff
@@ -1,61 +1,118 @@
...
+      <div className="App">
+        <h1>Add todo form</h1>
+
+        <form
+          action="/api/users"
+          method="POST"
+          onSubmit={handleSubmitTodo}
+        >
+
+          <div className="field" />
```
Anastasiia-Svintsova, line 73:
```suggestion
        >
          <div className="field" />
```

Anastasiia-Svintsova (changes requested):
- I can add a todo with an empty title, fix it, please 

https://user-images.githubusercontent.com/99403198/216759320-b8bef280-3d17-4696-af37-4dfb2fc8a89b.mov

PR #1433, opened 2023-02-04
Solution

OleksandrOse:
[DEMO LINK](https://OleksandrOse.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,143 @@
+import { useState, useEffect } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-export const App = () => {
+import { User } from './types/User';
+import { Todo } from './types/Todo';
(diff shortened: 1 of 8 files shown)


SanyaBratashchuk (changes requested):
I can add todos with title that contains only spaces 
image

Volodymyr-Melnyk-Fron10der (changes requested):
Great, but there is still no labels
image

src/App.tsx
```diff
@@ -1,61 +1,169 @@
...
+
+    setValues({
+      ...values,
+      [name]: value,
+    });
+  };
+
+  const handleSubmit = (e: React.FormEvent) => {
+    e.preventDefault();
+    const { user, title } = values;
```
DarkMistyRoom, line 58: you can do it on the one level higher and use it throughout the whole component

DarkMistyRoom (changes requested):
just make the default value disabled as in the example
image

Volodymyr-Melnyk-Fron10der (approved):
It`s perfect now! Great job!

PR #1443, opened 2023-02-06
solution

BrandonSunrise:
[DEMO LINK](https://BrandonSunrise.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,137 @@
 import './App.scss';
+import React, { useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { FullTodo, Todo } from './types/Todo';
+import { User } from './types/User';
+import { TodoList } from './components/TodoList';
 
-export const App = () => {
+const prepareTodos = (
+  users: User[],
+  todos: Todo[],
+): FullTodo[] => {
+  return todos.map((todo) => ({
+    ...todo,
+    user: users.find((user) => user.id === todo.userId) || null,
+  }));
+};
+
+const prepareTodo = (
+  currentUser: User,
+  title: string,
+  todos: Todo[],
+): FullTodo => {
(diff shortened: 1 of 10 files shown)


alexandra-protyanova (changes requested):
Don't forget to import styles, demo link looks terrible

BrandonSunrise:
> 


I'm so sorry, I was in a hurry due to the blackouts and forgot to connect the styles

src/App.tsx
```diff
@@ -1,25 +1,137 @@
...
+import todosFromServer from './api/todos';
+import { FullTodo, Todo } from './types/Todo';
+import { User } from './types/User';
+import { TodoList } from './components/TodoList';
+
+const prepareTodos = (
+  users: User[],
+  todos: Todo[],
+): FullTodo[] => {
+  return todos.map((todo) => ({
```
alexandra-protyanova, line 14: you can avoid redundant `return`

src/App.tsx
```diff
@@ -1,25 +1,137 @@
...
+    userId: currentUser.id,
+    user: currentUser,
+  });
+};
+
+const visibleTodos = prepareTodos(usersFromServer, todosFromServer);
+const users = usersFromServer;
+
+export const App: React.FC = () => {
+  const [userSelected, setUserSelected] = useState('0');
```
alexandra-protyanova, line 40: in my opinion it's more appropriate to set to `value` not `user.name` but `user.id`

alexandra-protyanova (approved):
Well done 👍

src/App.tsx
BrandonSunrise: Thank you Sasha! I always forgot about those redundant returns! I will try to keep it in my mind all the time(

ihor-soloviov (commented):
Good Job!
P.S. I don't know why I need to check already approved task for my mentors review. But this is it

BrandonSunrise:
> Good Job! P.S. I don't know why I need to check already approved task for my mentors review. But this is it

I pushed my task when platform didnt work. so it seems to me that it;s just a bug)

alSemenyuk:
GJ

PR #1455, opened 2023-02-06
solution to react_add-todo task

andrii-kurash:
[DEMO LINK](https://andrii-kurash.github.io/react_add-todo-form/)

Great mood!😉

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,133 @@
+import React, { useState } from 'react';
+import { TodoList } from './components/TodoList';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
+    id: lastId + 1,
+    title,
+    completed: false,
+    userId: selectedUser.id,
+    user: selectedUser,
+  });
+};
+
+export const App: React.FC = () => {
+  const [todos, setTodos] = useState(getTodos);
```
denys-cheporniuk, line 33: add type for `todos` state

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
+  const [userError, setUserError] = useState(false);
+  const [titleError, setTitleError] = useState(false);
+  const [title, setTitle] = useState('');
+  const [userSelected, setUserSelected] = useState('0');
```
denys-cheporniuk, line 37: why string?
Let's make it number if you use user.id for make it an empty string if you use user.name

denys-cheporniuk (approved):
good job 👍

PR #1458, opened 2023-02-07
solution

imariash86:
[DEMO LINK](https://imariash86.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,115 @@
+import React, { FormEvent, useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+export const App: React.FC = () => {
+  const preparedTodos = todosFromServer.map(todo => {
+    return {
+      ...todo,
+      user: usersFromServer.find(user => user.id === todo.userId) || null,
+    };
+  });
 
-export const App = () => {
+  const [todos, setTodos] = useState(preparedTodos);
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [isTitleEmpty, setIsTitleEmpty] = useState(false);
+  const [isUserSelected, setIsUserSelected] = useState(true);
+
+  const handleTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+    setIsTitleEmpty(false);
+  };
+
+  const handleSubmit = (event: FormEvent) => {
+    event.preventDefault();
+
+    if (!title) {
+      setIsTitleEmpty(true);
+    }
+
+    if (!userId) {
+      setIsUserSelected(false);
+    }
+
+    if (title && userId) {
+      setTodos(currTodos => ([
+        ...currTodos,
+        {
+          id: Math.max(...currTodos.map(todo => todo.id)) + 1,
+          title,
+          completed: false,
+          userId,
+          user: usersFromServer.find(user => user.id === userId) || null,
+        },
+      ]));
+      setTitle('');
+      setUserId(0);
+    }
+  };
+
+  const handleSelect = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setUserId(+event.target.value);
+    setIsUserSelected(true);
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
+        action="/api/users"
+        method="POST"
+        onSubmit={handleSubmit}
+      >
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <input
+            type="text"
+            data-cy="titleInput"
+            placeholder="Enter a title"
+            value={title}
+            onChange={handleTitle}
+          />
+
+          {isTitleEmpty
+            && <span className="error">Please enter a title</span>}
+
         </div>
 
         <div className="field">
-          <select data-cy="userSelect">
+          <select
+            value={userId}
+            data-cy="userSelect"
+            onChange={handleSelect}
+          >
             <option value="0" disabled>Choose a user</option>
+            {usersFromServer.map(user => {
+              return (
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,115 @@
...
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+export const App: React.FC = () => {
+  const preparedTodos = todosFromServer.map(todo => {
+    return {
+      ...todo,
+      user: usersFromServer.find(user => user.id === todo.userId) || null,
+    };
+  });
```
denys-cheporniuk, line 14: this code shouldn't be in the component

src/App.tsx
```diff
@@ -1,61 +1,115 @@
...
+
+  const [todos, setTodos] = useState(preparedTodos);
```
denys-cheporniuk, line 16: add  type for `todos` state

src/App.tsx
```diff
@@ -1,61 +1,115 @@
...
+        {
+          id: Math.max(...currTodos.map(todo => todo.id)) + 1,
+          title,
+          completed: false,
+          userId,
+          user: usersFromServer.find(user => user.id === userId) || null,
+        },
+      ]));
+      setTitle('');
+      setUserId(0);
```
denys-cheporniuk, line 50: create a separate function for resetting form

src/App.tsx
```diff
@@ -1,61 +1,115 @@
...
         <div className="field">
-          <select data-cy="userSelect">
+          <select
+            value={userId}
+            data-cy="userSelect"
+            onChange={handleSelect}
+          >
             <option value="0" disabled>Choose a user</option>
+            {usersFromServer.map(user => {
+              return (
```
denys-cheporniuk, line 90: you can avoid return here

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,19 @@
...
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="TodoList">
+
+      {todos.map(todo => {
+        return <TodoInfo todo={todo} key={todo.id} />;
+      })}
```
denys-cheporniuk, line 15: the same here with return

denys-cheporniuk (changes requested):

src/utils.tsx
```diff
@@ -0,0 +1,12 @@
+import { User } from './types';
+
+export interface Todo {
+  id: number,
+  title: string,
+  completed: boolean,
+  userId: number,
+}
+
+export const getUsersFromServer = (usersFromServer: User[], todo: Todo) => {
```
Anna-Tkachenko, line 10: check function name
this function doesn't return users from server

src/utils.tsx
```diff
@@ -0,0 +1,12 @@
+import { User } from './types';
+
+export interface Todo {
```
Anna-Tkachenko, line 3: move `Todo` interface in separate file

src/types.ts
```diff
@@ -0,0 +1,14 @@
+export interface User {
```
Anna-Tkachenko, line 1: will be better to create separate files for `User` and `Todo` types

src/App.tsx
```diff
@@ -1,61 +1,120 @@
...
+            {usersFromServer.map(user => (
+              <option value={user.id} key={user.id}>
+                {user.name}
+              </option>
+            ))}
           </select>
 
-          <span className="error">Please choose a user</span>
+          {isUserEmpty
+            && <span className="error">Please choose a user</span>}
```
Anna-Tkachenko, line 104:
```suggestion
          {isUserEmpty && (
            <span className="error">Please choose a user</span>
          )}
```

src/App.tsx
```diff
@@ -1,61 +1,120 @@
...
+          <input
+            type="text"
+            data-cy="titleInput"
+            placeholder="Enter a title"
+            value={title}
+            onChange={handleTitle}
+          />
+
+          {isTitleEmpty
+            && <span className="error">Please enter a title</span>}
```
Anna-Tkachenko, line 85:
```suggestion
          {isTitleEmpty && (
            <span className="error">Please enter a title</span>
          )}
```

src/App.tsx
```diff
@@ -1,61 +1,120 @@
...
+
+    if (title && userId) {
+      setTodos(currTodos => ([
+        ...currTodos,
+        {
+          id: Math.max(...currTodos.map(todo => todo.id)) + 1,
+          title,
+          completed: false,
+          userId,
+          user: usersFromServer.find(user => user.id === userId) || null,
```
Anna-Tkachenko, line 53: you're already have the function for getting user by id, check it in your code

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+import { Todo } from './types/todo';
+
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+const preparedTodos = todosFromServer.map(todo => {
+  return {
+    ...todo,
+    user: getUserById(todo.id),
```
Anna-Tkachenko, line 13: probably you need to pass `userId` property in this function?

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+  };
+});
+
+export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>(preparedTodos);
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [isTitleEmpty, setIsTitleEmpty] = useState(false);
+  const [isUserEmpty, setIsUserEmpty] = useState(false);
```
Anna-Tkachenko, line 22: this values are empty in the initial state, so it's not clear why did you set `false` values for them as the initial ones

will be better to rename it for better understanding that it's variables for errors, not just for checking if `title` and `user` are empty for now

Anna-Tkachenko (changes requested):

src/App.tsx
imariash86: Thanks for suggestions, I'm fixing it

Anna-Tkachenko (approved):
Good job!

PR #1459, opened 2023-02-07
only first part

AlexPro100pro:
[DEMO LINK](https://AlexPro100pro.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,107 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+import { ChangeEvent, FormEvent, useState } from 'react';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [selectPersonName, setselectPersonName] = useState('');
+  const [title, setTitle] = useState('');
+  const [listTodos, setListTodos] = useState(todos);
+
+  const sendForm = (event: FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    const selectedUser = usersFromServer.find(
+      user => user.name === selectPersonName,
+    );
+
+    if (!selectedUser) {
+      return;
+    }
+
+    const newTodo = {
+      id: todos.length + 1,
+      title,
+      completed: false,
+      userId: selectedUser.id,
+      user: selectedUser,
+    };
+
+    setListTodos([...listTodos, newTodo]);
+  };
+
+  // const selectInput = (event: FormEventHandler<HTMLOptionElement>) => {
+  //   setselectPersonName(event.target.value);
+  // };
+
+  const handleInput = (event: ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form action="/api/users" method="POST" onSubmit={sendForm}>
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <label>
+            Title:
+            <input
+              type="text"
+              data-cy="titleInput"
+              placeholder="Enter a title"
+              value={title}
+              onChange={handleInput}
+            />
+          </label>
+          {!title && <span className="error">Please enter a title</span>}
+
         </div>
 
         <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>Choose a user</option>
-          </select>
-
+          <label>
+            User:
+            <select
+              data-cy="userSelect"
+              value={selectPersonName}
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,107 @@
...
+    const selectedUser = usersFromServer.find(
+      user => user.name === selectPersonName,
+    );
+
+    if (!selectedUser) {
+      return;
+    }
+
+    const newTodo = {
+      id: todos.length + 1,
```
Serghii29, line 37: you can't use length for id

src/App.tsx
```diff
@@ -1,61 +1,126 @@
...
 export const App = () => {
+  const [selectUserId, setselectUserId] = useState(0);
+  const [title, setTitle] = useState('');
+  const [listTodos, setListTodos] = useState<Todo[]>(todos);
+
+  const [hasNameError, setHasNameError] = useState(false);
+  const [hasTitleError, setHasTitleError] = useState(false);
+
+  const valideData = `zxcvbnmasdfghjklqwertyuiopZXCVBNMASDFGHJKLQWERTY
+  UIOPячсмитьбюЯЧСМИТЬБЮфывапролджэФЫВАПРОЛДЖЭйцукенгшщзхъЙЦУКЕНГШЩЗХёЁ123456789`;
```
denys-cheporniuk, line 30: nope, just check if string empty or not, you don't need this variable

src/App.tsx
```diff
@@ -1,61 +1,126 @@
...
+
+  const selectInput = (event: ChangeEvent<HTMLSelectElement>) => {
+    setselectUserId(+event.target.value);
+    setHasNameError(false);
+  };
+
+  const handleInput = (event: ChangeEvent<HTMLInputElement>) => {
+    if (event.target.value
+      .split('')
+      .every(letter => valideData.includes(letter))) {
```
denys-cheporniuk, line 45: you don't need it

src/App.tsx
```diff
@@ -1,61 +1,126 @@
...
+    if (selectUserId && title) {
+      const newTodo = {
+        id: Math.max(...listTodos.map(todo => todo.id)) + 1,
+        title,
+        completed: false,
+        userId: selectUserId,
+        user: getUser(selectUserId),
+      };
+
+      setListTodos([...listTodos, newTodo]);
```
denys-cheporniuk, line 63: use functional state if you need current value from state

denys-cheporniuk (changes requested):

src/App.tsx
OleksandrProkop: point 5

src/App.tsx
OleksandrProkop: point 5

src/App.tsx
```diff
@@ -1,61 +1,127 @@
...
+  return foundUser || null;
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [selectUserId, setselectUserId] = useState(0);
```
Anna-Tkachenko, line 22:
```suggestion
  const [selectedUserId, setSelectedUserId] = useState(0);
```

src/App.tsx
```diff
@@ -1,61 +1,127 @@
...
+  const [title, setTitle] = useState('');
+  const [listTodos, setListTodos] = useState<Todo[]>(todos);
+
+  const [hasNameError, setHasNameError] = useState(false);
+  const [hasTitleError, setHasTitleError] = useState(false);
+
+  const valideData = `zxcvbnmasdfghjklqwertyuiopZXCVBNMASDFGHJKLQWERTY
```
Anna-Tkachenko, line 29: will be better to handle it via some regular expression instead of creating string with allowed symbols manually
https://stackoverflow.com/questions/388996/regex-for-javascript-to-allow-only-alphanumeric

src/App.tsx
```diff
@@ -1,61 +1,127 @@
...
+  const handleInput = (event: ChangeEvent<HTMLInputElement>) => {
+    if (event.target.value
+      .split('')
+      .every(letter => valideData.includes(letter))) {
+      setTitle(event.target.value);
+      setHasTitleError(false);
+    }
+  };
+
+  const sendForm = (event: FormEvent<HTMLFormElement>) => {
```
Anna-Tkachenko, line 51: this function doesn't send any form
it's just invoked on form submit, so better to name it as some handler: `handleFormSubmit`, for example

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,22 @@
...
+export const TodoInfo: React.FC<Props> = ({ todo }) => (
+
+  <article
+    className={classNames(
+      'TodoInfo',
+      { 'TodoInfo--completed': todo.completed },
+    )}
+  >
+    <h2 className="TodoInfo__title">{todo.title}</h2>
+    {todo.user && <UserInfo user={todo.user} />}
```
Anna-Tkachenko, line 19:
```suggestion
    {todo.user && (
      <UserInfo user={todo.user} />
    )}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,14 @@
...
+import { TodoInfo } from '../TodoInfo';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <section className="TodoList">
+    {todos.map(todo => <TodoInfo todo={todo} key={todo.id} />)}
```
Anna-Tkachenko, line 12:
```suggestion
    {todos.map(todo => (
      <TodoInfo todo={todo} key={todo.id} />
    ))}
```

Anna-Tkachenko (approved):

PR #1461, opened 2023-02-07
Develop

victorionov:
https://victorionov.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -31,7 +31,7 @@ export const App = () => {
   const [errorMessage, setErrorMessage] = useState(false);
 
   const getNewId = () => (
-    todos.length + 1
+    new Date().getTime()
   );
 
   const addTodo = (arr: Todo[]) => {
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,135 @@
 import './App.scss';
+import React, { useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { Todo } from './types/Todo';
+import { User } from './types/User';
+import { TodoList } from './components/TodoList';
+
+function getUserById(userId: number): User | null {
+  const correctUser = usersFromServer.find(user => user.id === userId);
+
+  return correctUser || null;
+}
+
+function getUserByName(userName: string) {
+  const correctUser = usersFromServer.find(({ name }) => userName === name);
+
+  return correctUser || null;
+}
+
+export const preparedTodos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [todos, setTodos] = useState(preparedTodos);
+  const [selectedUser, setSelectedUser] = useState('');
+  const [errorMessage, setErrorMessage] = useState(false);
+
+  const getNewId = () => (
+    todos.length + 1
+  );
+
+  const addTodo = (arr: Todo[]) => {
+    const correctUser = getUserByName(selectedUser);
+
+    if (correctUser) {
+      const newTodo: Todo = {
+        id: getNewId(),
+        userId: correctUser.id,
+        title,
+        completed: false,
+        user: correctUser,
+      };
+
+      setTodos([...arr, newTodo]);
+    }
+  };
+
+  const clearForm = () => {
+    setSelectedUser('');
+    setTitle('');
+    setErrorMessage(false);
+  };
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+  };
+
+  const handleUserChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setSelectedUser(event.target.value);
+  };
+
+  const handleTitleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+}
+
+export const preparedTodos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [todos, setTodos] = useState(preparedTodos);
```
alexandra-protyanova, line 29: set type to non primitive states

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+
+      const newTodo: Todo = {
+        id: newId,
+        userId: correctUser.id,
+        title,
+        completed: false,
+        user: correctUser,
+      };
+
+      setTodos([...arr, newTodo]);
```
alexandra-protyanova, line 47: use functional `setState` instead

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+  };
+
+  const handleUserChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setSelectedUser(event.target.value);
+  };
+
+  const handleTitleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
```
alexandra-protyanova, line 65:
```suggestion
  const handleFormSubmit = (event: React.FormEvent<HTMLFormElement>) => {
```

alexandra-protyanova (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+
+export const preparedTodos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [todos, setTodos] = useState<Todo[]>(preparedTodos);
+  const [selectedUser, setSelectedUser] = useState('');
```
Anna-Tkachenko, line 30: will be better to handle users by id, not by name
because `id` it's main unique field for the users

Anna-Tkachenko: and in this case you don't need to use separate `getUserByName` function

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+      setErrorMessage(true);
+
+      return;
+    }
+
+    addTodo();
+    clearForm();
+  };
+
+  const titleErrorStatus = !title && errorMessage;
```
Anna-Tkachenko, line 78: will be better to name these variables as a boolean ones
for now as user I expect that such variables can contain some status in `string` format, not in the boolean
you can use `hasTitleError`, for example

src/types/Todo.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export interface Todo {
+  id: number,
+  userId: number,
+  title: string,
+  completed: boolean,
+  user: User | null,
+}
+
+export interface AllTodo extends Todo {
```
Anna-Tkachenko, line 11: it's not clear, what is `AllTodo`, think about more understandable type name

Anna-Tkachenko: and your `Todo` type is already contains user property

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,117 @@
...
+export const preparedTodos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [todos, setTodos] = useState<Todo[]>(preparedTodos);
+  const [selectedUserId, setSelectedUserId] = useState(0);
+  const [hasSelectedError, setHasSelectedError] = useState(false);
```
Anna-Tkachenko, line 25:
```suggestion
  const [hasSelectedUserError, setHasSelectedUserError] = useState(false);
```

src/App.tsx
```diff
@@ -1,61 +1,117 @@
...
+  const [hasTitleError, setHasTitleError] = useState(false);
+
+  const handleFormSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    setHasSelectedError(!selectedUserId);
+    setHasTitleError(!title);
+
+    const getMaxId = Math.max(...todos.map(todo => todo.id)) + 1;
```
Anna-Tkachenko, line 34: according to the variable name I expect that it's a function
but it's just the variable for storing max id
check variable name, probably just `maxId` will be better

src/App.tsx
```diff
@@ -1,61 +1,117 @@
...
+    const clearForm = () => {
+      setTitle('');
+      setSelectedUserId(0);
+    };
+
+    const newTodo = {
+      id: getMaxId,
+      title,
+      completed: false,
+      userId: +selectedUserId,
```
Anna-Tkachenko, line 45: do we need to convert the value to the number here? you're already stored the number in the state

src/App.tsx
```diff
@@ -1,61 +1,117 @@
...
+      user: getUserById(+selectedUserId),
```
Anna-Tkachenko, line 46: same here about types conversion

src/types/Todo.ts
Anna-Tkachenko: you didn't fix this comment

Anna-Tkachenko (changes requested):

PR #1462, opened 2023-02-07
Solution

maasiimka:


denys-cheporniuk (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,129 @@
...
+          completed: false,
+          userId,
+          user: findUserByUserId(userId),
+        },
+      ]));
+
+      setTitle('');
+      setUserId(-1);
+
+      return;
```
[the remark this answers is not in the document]
max-rozzhalovets: Yes, I set my `title` and  `userId` values '' and 0 and then check if they === '' or === 0, so I need to return from function before check

src/App.tsx
```diff
@@ -1,61 +1,129 @@
...
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+
+const validationString
+= 'ABCDEFGHIJKLMNOPQRSTUVWXYАБВГҐДЕЄЖЗИІЇЙКЛМНОПРСТУФХЦЧШЩЬЮЯ123456789 ';
```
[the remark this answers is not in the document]
Anna-Tkachenko: will be better to use regular expression instead of creating string with all possible values for the title

src/App.tsx
```diff
@@ -1,61 +1,139 @@
...
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/Todo';
+import { User } from './types/User';
+
+const validationString
+= 'ABCDEFGHIJKLMNOPQRSTUVWXYАБВГҐДЕЄЖЗИІЇЙКЛМНОПРСТУФХЦЧШЩЬЮЯ123456789 ';
+
+type TodoWithUser = Todo & {
```
Anna-Tkachenko, line 13: let's move this type to the file where you described the main `Todo` type

src/App.tsx
```diff
@@ -1,61 +1,139 @@
...
+  user: User | null,
+};
+
+const findUserByUserId = (userId: number) => {
```
Anna-Tkachenko, line 17: don't forget to add the type for return value for the function

src/App.tsx
```diff
@@ -1,61 +1,139 @@
...
+
+  return {
+    ...todo,
+    user: findUserByUserId(userId),
+  };
+});
 
 export const App = () => {
+  const defaultTitle = '';
+  const defaultUserId = 0;
```
Anna-Tkachenko, line 32: do you need these extra variables?

src/App.tsx
```diff
@@ -1,61 +1,139 @@
...
+
+  const [todos, setTodos] = useState<TodoWithUser[]>(todosWithUser);
+  const [title, setTitle] = useState(defaultTitle);
+  const [userId, setUserId] = useState(defaultUserId);
+  const [titleValid, setTitleValidity] = useState(true);
+  const [userIdValid, setUserIdValidity] = useState(true);
+
+  const getNewId = () => (Math.max(...todos.map(({ id }) => id)) + 1);
```
Anna-Tkachenko, line 40: move this function outside of the component
for now this function is recreated on each component rerendering

or let's create it and call only while form submitting action

src/App.tsx
```diff
@@ -1,61 +1,139 @@
...
+          <input
+            type="text"
+            data-cy="titleInput"
+            placeholder="Enter the title"
+            value={title}
+            onChange={handleInputChange}
+          />
+          {
+            !titleValid && <span className="error">Please enter a title</span>
+          }
```
Anna-Tkachenko, line 108:
```suggestion
          {!titleValid && (
            <span className="error">Please enter a title</span>
          )}
```

src/App.tsx
```diff
@@ -1,61 +1,139 @@
...
 
 export const App = () => {
+  const defaultTitle = '';
+  const defaultUserId = 0;
+
+  const [todos, setTodos] = useState<TodoWithUser[]>(todosWithUser);
+  const [title, setTitle] = useState(defaultTitle);
+  const [userId, setUserId] = useState(defaultUserId);
+  const [titleValid, setTitleValidity] = useState(true);
+  const [userIdValid, setUserIdValidity] = useState(true);
```
Anna-Tkachenko, line 38:
```suggestion
  const [isTitleValid, setIsTitleValidity] = useState(true);
  const [isUserIdValid, setIsUserIdValidity] = useState(true);
```

src/App.tsx
```diff
@@ -1,61 +1,139 @@
...
         <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>Choose a user</option>
+          User:
+          <select
+            data-cy="userSelect"
+            value={userId}
+            onChange={handleUserChange}
+          >
+            <option disabled value={defaultUserId}>Choose a user</option>
```
Anna-Tkachenko, line 118:
```suggestion
            <option disabled value={defaultUserId}>
              Choose a user
            </option>
```

src/App.tsx
```diff
@@ -1,61 +1,139 @@
...
+              usersFromServer.map(({ id, name }) => (
+                <option key={id} value={id}>{name}</option>
+              ))
+            }
           </select>
 
-          <span className="error">Please choose a user</span>
+          {
+            !userIdValid && <span className="error">Please choose a user</span>
+          }
```
Anna-Tkachenko, line 128: fix in the same way as in previous comment for `title` error message

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,35 @@
...
+        'TodoInfo--completed': completed,
+      })}
+    >
+      <h2 className="TodoInfo__title">
+        {title}
+      </h2>
+
+      {
+        user !== null && <UserInfo user={user} />
+      }
```
Anna-Tkachenko, line 32:
```suggestion

      {user && (
        <UserInfo user={user} />
      )}
```

Anna-Tkachenko (changes requested):
Check demo link, it doesn't work

denys-cheporniuk (changes requested):
you didn't fix all comments from @Anna-Tkachenko

src/App.tsx
max-rozzhalovets: Yes @denys-cheporniuk told me to create separate variable for resetting form

denys-cheporniuk (approved):

PR #1466, opened 2023-02-07
React_Add_Todo_Form were implemented

Kanzatuwka:
[DEMO LINK](https://kanzatuwka.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,116 @@
+import React, { ChangeEvent, FormEvent, useState } from 'react';
+import { Todo } from './types/Todo';
+import { User } from './types/User';
+import { TodoList } from './components/TodoList';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-export const App = () => {
+const getUserById = (userId: number): User | null => {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+};
+
+export const todosList: Todo[] = todosFromServer.map((todo) => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
+
+export const App : React.FC = () => {
+  const [todos, setTodo] = useState<Todo[]>(todosList);
+  const [titleName, setTitleName] = useState('');
+  const [selectedUserId, SetSelectedUserID] = useState(0);
+  const [hasTitleError, SetTitleError] = useState(false);
+  const [hasUserError, SetUserError] = useState(false);
+
+  const idForNewTodo = Math.max(...todos.map(todo => todo.id));
+
+  const handleTitle = (event: ChangeEvent<HTMLInputElement>) => {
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,61 +1,116 @@
...
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+const getUserById = (userId: number): User | null => {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+};
+
+export const todosList: Todo[] = todosFromServer.map((todo) => ({
```
DarkMistyRoom, line 16: why do you need to export it?

src/App.tsx
```diff
@@ -1,61 +1,116 @@
...
+  ...todo,
+  user: getUserById(todo.userId),
+}));
+
+export const App : React.FC = () => {
+  const [todos, setTodo] = useState<Todo[]>(todosList);
+  const [titleName, setTitleName] = useState('');
+  const [selectedUserId, SetSelectedUserID] = useState(0);
+  const [hasTitleError, SetTitleError] = useState(false);
+  const [hasUserError, SetUserError] = useState(false);
```
DarkMistyRoom, line 26:
```suggestion
  const [selectedUserId, setSelectedUserID] = useState(0);
  const [hasTitleError, setTitleError] = useState(false);
  const [hasUserError, setUserError] = useState(false);
```

src/App.tsx
```diff
@@ -1,61 +1,116 @@
...
+
+  const idForNewTodo = Math.max(...todos.map(todo => todo.id));
```
DarkMistyRoom, line 28: map returns the copy, no need to make a copy one more time. and here you just chose the max id from existed, but it should be unique

src/App.tsx
```diff
@@ -1,61 +1,116 @@
...
+  };
+
+  const handleFormSubmit = (event: FormEvent) => {
+    event.preventDefault();
+
+    SetTitleError(!titleName);
+    SetUserError(!selectedUserId);
+
+    const newTodo: Todo = {
+      id: idForNewTodo + 1,
```
DarkMistyRoom, line 52: Ok, I see that you add 1, but this name confused me. Add 1 in the assignment or change the name

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,33 @@
...
+  } = todo;
+
+  return (
+    <article
+      data-id={todo.id}
+      className={
+        className(
+          'TodoInfo', {
+            'TodoInfo--completed': completed,
+          },
```
DarkMistyRoom, line 25:
```suggestion
          'TodoInfo', 
          {
            'TodoInfo--completed': completed,
          },
```

DarkMistyRoom (changes requested):
look, I can add the todo only with whitespaces
image

src/App.tsx
Kanzatuwka: For nothing, you're right. Fixed.

PR #1472, opened 2023-02-07
--solution

JOWISSA:
--user info is not added in the new todo
--all tasks are always not completed
I need help)


tetiana-ivanchenko (changes requested):
If you need some help - write your questions with screenshots in the slack chat, you should create PR only if your task is done
also, add all folders to github (you didn't add type folder)
also add demo link

JOWISSA:
[DEMO LINK](https://JOWISSA.github.io/react_add-todo-form/)

demetrxx (approved):
Well done 👍 Go ahead!

PR #1474, opened 2023-02-07
react_add-todo-form

rafmamedov:
[DEMO LINK](https://rafmamedov.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,141 @@
+import React, { useState } from 'react';
 import './App.scss';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { Todo } from './components/TodoInfo';
+import { User } from './components/UserInfo';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+export function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
 
-export const App = () => {
+  return foundUser || null;
+}
+
+const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState('0');
+  const [todoList, setTodoList] = useState(todos);
+  const [titleError, setTitleError] = useState(false);
+  const [userError, setUserError] = useState(false);
+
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,61 +1,141 @@
...
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { Todo } from './components/TodoInfo';
+import { User } from './components/UserInfo';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+export function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
```
Volodymyr-Melnyk-Fron10der, line 10: 1. Use decomposition as much as possible 
2. You can write it with one line (optional, your variant is not bad actually)
```suggestion
  return usersFromServer.find(({ id })=> id === userId) || null;
```

src/App.tsx
```diff
@@ -1,61 +1,141 @@
...
+  const [todoList, setTodoList] = useState(todos);
+  const [titleError, setTitleError] = useState(false);
+  const [userError, setUserError] = useState(false);
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    if (titleError) {
+      setTitleError(false);
+    }
+
+    if (/^[a-zA-Z0-9 ]*$/.test(event.target.value)) {
```
Volodymyr-Melnyk-Fron10der, line 32: move regular expression into constant variable

src/App.tsx
```diff
@@ -1,61 +1,141 @@
...
+      isError = true;
+    }
+
+    if (isError) {
+      return;
+    }
+
+    const id = todos
+      .map(todo => todo.id)
+      .sort((a, b) => a - b)[todos.length - 1] + 1;
```
Volodymyr-Melnyk-Fron10der, line 66: You can make it less:
```suggestion
    const id = todos.sort((prevTodo, nextTodo) => nextTodo.id - prevTodo.id)[0]?.id + 1;
```

Volodymyr-Melnyk-Fron10der (changes requested):
Hi,

1. I can add user with just a few spaces instead of letters
image
image

DarkMistyRoom (changes requested):
new todo adding removes the previous one
(https://www.loom.com/share/dca3f4d8cf034ad6b4d32adc81687979)

SanyaBratashchuk (approved):

PR #1476, opened 2023-02-08
solution

miamarichka:
[DEMO LINK](https://miamarichka.github.io/react_add-todo-form/)


miamarichka:
image
Dont have any idea why this checks failed here, work perfectly in terminal

polosanya (changes requested):
The title is deleted after the error appears.
It's wrong behaviour, check yourself with the working example
add

PR #1477, opened 2023-02-09
Initial commit

yaroslav-furhala:
- [DEMO LINK](https://yaroslav-furhala.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,1 +1,23 @@
-export const TodoInfo = () => {};
+import classNames from 'classnames';
+import { Todo } from '../types/Todo';
+import { UserInfo } from '../UserInfo/UserInfo';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => (
+  <article
+    data-id={todo.id}
+    className={classNames(
+      'TodoInfo',
+      {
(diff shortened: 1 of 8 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,23 @@
-export const TodoInfo = () => {};
+import classNames from 'classnames';
+import { Todo } from '../types/Todo';
+import { UserInfo } from '../UserInfo/UserInfo';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => (
```
Volodymyr-Melnyk-Fron10der, line 9: Use decompose structure where is possible 
```suggestion
export const TodoInfo: React.FC<Props> = ({ todo: { id, title, user } }) => (
```

Volodymyr-Melnyk-Fron10der (changes requested):
Great! just add more decomposing pls

yaroslav-furhala:
1. (I would move regexp into single constant) - Done!
2. (Use decompose structure where is possible) - Done!

DarkMistyRoom (changes requested):
there is possibility to add todo with only whitespaces
image

yaroslav-furhala:
1. (there is possibility to add todo with only whitespaces) - Fixed!

SanyaBratashchuk (approved):

PR #1483, opened 2023-02-13
Solution

OksanaBaloh:
[DEMO LINK](https://OksanaBaloh.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,101 @@
+import React, { useState } from 'react';
 import './App.scss';
+import { TodoList } from './components/TodoList';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-export const App = () => {
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [titleError, setTitleError] = useState(false);
+  const [userId, setUserId] = useState(0);
+  const [userError, setUserError] = useState(false);
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    setTitleError(!title);
+    setUserError(!userId);
+
+    if (!title || !userId) {
+      return;
+    }
+
+    const newTodo = {
+      id: Math.max(...todos.map(todo => todo.id)) + 1,
+      title,
+      completed: false,
+      userId,
+      user: getUser(userId),
+    };
+
+    todos.push(newTodo);
+
+    setTitle('');
+    setUserId(0);
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form action="/api/users" method="POST" onSubmit={handleSubmit}>
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <label>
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,61 +1,109 @@
...
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [titleError, setTitleError] = useState(false);
```
loralevitska, line 23: Rename boolean variables according [recommendation](https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig) Fix all cases please

src/App.tsx
```diff
@@ -1,61 +1,109 @@
...
+            <input
+              type="text"
+              data-cy="titleInput"
+              name="title"
+              placeholder="Enter a title"
+              value={title}
+              onChange={event => {
+                setTitle(event.target.value.replace(/[^a-z0-9а-я\s]/gi, ''));
+                setTitleError(false);
+              }}
```
loralevitska, line 68: Move to separate function  (transfer the function to a variable, and here only write the name of the function)

src/App.tsx
```diff
@@ -1,61 +1,109 @@
...
+          <label>
+            User:&nbsp;
+            <select
+              data-cy="userSelect"
+              name="user"
+              value={userId}
+              onChange={event => {
+                setUserId(+event.target.value);
+                setUserError(false);
+              }}
```
loralevitska, line 85: Move to separate function  (transfer the function to a variable, and here only write the name of the function)

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,17 @@
...
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="TodoList">
+      {todos.map(todo => (
+        <TodoInfo todo={todo} key={todo.id} />
+      ))}
+    </section>
```
loralevitska, line 15: Use semantic tags for list `ul` `li`. Don't forget to remove default `li` dots

loralevitska (changes requested):
Keep up the good work! Some comments to fix:

1. I can add empty comment. Fix it with  [`trim`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim)
image

src/App.tsx
```diff
@@ -1,61 +1,115 @@
...
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
```
DarkMistyRoom, line 16: why do you need to export `todos`?

src/App.tsx
```diff
@@ -1,61 +1,115 @@
...
+
+    setTitle('');
+    setUserId(0);
+  };
+
+  const handleChangeTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value
+      .replace(/[^a-z0-9а-я\s]/gi, '')
+      .trimStart());
+    setHasTitle(false);
```
DarkMistyRoom, line 55: looks like there should be `true`, or another name of this state

DarkMistyRoom (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,115 @@
...
+      title,
+      completed: false,
+      userId,
+      user: getUser(userId),
+    };
+
+    todos.push(newTodo);
+
+    setTitle('');
+    setUserId(0);
```
Volodymyr-Melnyk-Fron10der, line 48: It works fine only because you change other states after `todos.push(newTodo);`, but if you don't do it - you wouldn't see new todos, because component wouldn't be rerendered. So move all items into state to make sure component will be rerendered on any changes on next time

Volodymyr-Melnyk-Fron10der (approved):
Great✅ I just left one note to keep in mind

PR #1489, opened 2023-02-14
Solution

IVxVI:
[DEMO LINK](https://ivxvi.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,128 @@
 import './App.scss';
+import './field.scss';
+import { useState } from 'react';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [addedTitle, setAddedTitle] = useState(false);
+  const [userId, setUserId] = useState(0);
+  const [addedUser, setAddedUser] = useState(false);
+
+  const newTodoId = Math.max(...todos.map(todo => todo.id)) + 1;
+
+  const addTodo = (id: number) => {
+    if (getUser(id) && title.trim()) {
+      todos.push({
+        id: newTodoId,
+        title,
+        userId,
+        completed: false,
+        user: getUser(userId),
+      });
+
+      setUserId(0);
+      setTitle('');
+    }
+
+    if (!title.trim()) {
+      setAddedTitle(true);
+    }
+
+    if (!userId) {
+      setAddedUser(true);
+    }
+  };
+
+  const handleForm = (event: React.ChangeEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    addTodo(userId);
+  };
+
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,61 +1,128 @@
...
+  const newTodoId = Math.max(...todos.map(todo => todo.id)) + 1;
+
+  const addTodo = (id: number) => {
+    if (getUser(id) && title.trim()) {
+      todos.push({
+        id: newTodoId,
+        title,
+        userId,
+        completed: false,
+        user: getUser(userId),
```
vlad-demchuk, line 37: move it to a separate variable, just for clarity

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,16 @@
...
+import { TodoInfo } from '../TodoInfo';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="TodoList">
```
vlad-demchuk, line 10: it's not sections, it's just a list, use `ul, li`

vlad-demchuk (approved):

src/App.tsx
```diff
@@ -1,61 +1,128 @@
...
+const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [addedTitle, setAddedTitle] = useState(false);
+  const [userId, setUserId] = useState(0);
+  const [addedUser, setAddedUser] = useState(false);
```
tetiana-ivanchenko, line 26: follow the convention for the boolean variables name

tetiana-ivanchenko (changes requested):

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,53 @@
...
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    id,
+    title,
+    userId,
+    completed,
+    user,
+  } = todo;
+
+  const [isCompleted, setCompleted] = useState(false);
```
SanyaBratashchuk, line 20:
```suggestion
  const [isCompleted, setIsCompleted] = useState(false);
```

src/App.tsx
```diff
@@ -1,61 +1,137 @@
...
+const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [correctTitle, addTitle] = useState(false);
+  const [userId, addUserId] = useState(0);
+  const [chosenUser, chooseUser] = useState(false);
```
SanyaBratashchuk, line 26:
```suggestion
  const [correctTitle, setCorrectTitle] = useState(false);
  const [userId, setUserId] = useState(0);
  const [chosenUser, setChosenUser] = useState(false);
```

SanyaBratashchuk: const [value, setValue] = useState()

That is the only naming pattern for react state values

SanyaBratashchuk (changes requested):

src/App.tsx
IVxVI: OK, thanks. Fixed

src/App.tsx
```diff
@@ -1,61 +1,137 @@
...
+
+  const toDoItem = {
+    id: newTodoId,
+    title,
+    userId,
+    completed: false,
+    user: getUser(userId),
+  };
+
+  const regex = /^[\p{L}\d\s]+$/u;
```
Volodymyr-Melnyk-Fron10der, line 38: (For future) try to move independent variables out of components to avoid it's creating on every re-rendering, it will  save some time

Volodymyr-Melnyk-Fron10der (approved):
PerfectXanderLemmonGIF

PR #1493, opened 2023-02-17
react add todo from completed

MykolaKaradzha:
[DEMO LINK](https://MykolaKaradzha.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,106 @@
 import './App.scss';
-
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import React, { useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/Todo';
+import { User } from './types/User';
 
 export const App = () => {
+  const users:User[] = usersFromServer;
+  const [todos, setTodos] = useState<Todo[]>(todosFromServer);
+  const [userId, setUserId] = useState<number>(0);
+  const [title, setTitle] = useState<string>('');
+  const [errorTitle, setErrorTitle] = useState<boolean>(false);
+  const [errorUser, setErrorUser] = useState<boolean>(false);
+
+  const handleSubmitButton = (event: React.MouseEvent<HTMLButtonElement>) => {
+    event.preventDefault();
+    if (!userId) {
+      setErrorUser(true);
+    }
+
+    if (!title.trim()) {
+      setErrorTitle(true);
+    }
+
+    if (userId && title) {
+      const id = [...todos].sort(
+        (todo1, todo2) => todo2.id - todo1.id,
+      )[0].id + 1;
+
+      setTodos((
+        [
+          ...todos,
+          {
+            id,
+            title,
+            completed: false,
+            userId,
+          }]
+      ));
+      setUserId(0);
+      setTitle('');
+    }
+  };
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value.replace(/[^a-z-а-я-0-9-' ']/gi, ''));
+    setErrorTitle(false);
+  };
+
+  const handleSelect = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setUserId(+event.target.value);
+    setErrorUser(false);
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
       <form action="/api/users" method="POST">
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <label>
+            {'Title: '}
+            <input
+              type="text"
+              value={title}
+              onChange={handleTitleChange}
+              data-cy="titleInput"
+              placeholder="Enter a title"
+            />
+          </label>
+          {errorTitle && <span className="error">Please, enter a title</span>}
         </div>
 
         <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>Choose a user</option>
-          </select>
-
-          <span className="error">Please choose a user</span>
+          <label>
+            {'User: '}
+            <select
+              name="user"
+              data-cy="userSelect"
+              value={userId}
+              onChange={handleSelect}
+            >
+              <option value="0" disabled>Choose a user</option>
+              {
+                users.map(user => (
+                  <option value={user.id} key={user.id}>{user.name}</option>))
+              }
+            </select>
+          </label>
+          {errorUser && <span className="error">Please, choose a user</span>}
         </div>
 
-        <button type="submit" data-cy="submitButton">
+        <button
+          type="submit"
+          data-cy="submitButton"
+          onClick={handleSubmitButton}
+        >
           Add
         </button>
       </form>
-
-      <section className="TodoList">
-        <article data-id="1" className="TodoInfo TodoInfo--completed">
-          <h2 className="TodoInfo__title">
-            delectus aut autem
-          </h2>
-
-          <a className="UserInfo" href="mailto:Sincere@april.biz">
-            Leanne Graham
-          </a>
-        </article>
-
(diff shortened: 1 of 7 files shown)


IvanFesenko (changes requested):
Good job. Check my comments

src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
+
+    if (!title.trim()) {
+      setErrorTitle(true);
+    }
+
+    if (userId && title) {
+      addUser();
+      setUserId(0);
+      setTitle('');
+    }
```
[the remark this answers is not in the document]
MykolaKaradzha: Thanks, for all advices). But this one, I`ve particularly done this way, because in the situation when both title and username are undefined, it will setError only for one of the options.(

src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
+  const [todos, setTodos] = useState<Todo[]>(todosFromServer);
+  const [userId, setUserId] = useState<number>(0);
+  const [title, setTitle] = useState<string>('');
+  const [errorTitle, setErrorTitle] = useState<boolean>(false);
+  const [errorUser, setErrorUser] = useState<boolean>(false);
+
+  const addUser = () => {
+    const id = [...todos].sort(
+      (todo1, todo2) => todo2.id - todo1.id,
+    )[0].id + 1;
```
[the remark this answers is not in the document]
Viktor-Kost: is not fixed

src/App.tsx
```diff
@@ -1,61 +1,107 @@
...
+
+    if (userId && title) {
+      addUser();
+      setUserId(0);
+      setTitle('');
+    }
+  };
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value.replace(/[^a-z-а-я-0-9-' ']/gi, ''));
```
Viktor-Kost, line 49: I would recommend to create a constant for regex (ex: const `MY_REGEX = /{{your_regex}}/`)

src/App.tsx
```diff
@@ -1,61 +1,107 @@
...
+              name="user"
+              data-cy="userSelect"
+              value={userId}
+              onChange={handleSelect}
+            >
+              <option value="0" disabled>Choose a user</option>
+              {
+                usersFromServer.map(user => (
+                  <option value={user.id} key={user.id}>{user.name}</option>))
+              }
```
Viktor-Kost, line 90:
```suggestion
              {usersFromServer.map(user => (
                <option value={user.id} key={user.id}>{user.name}</option>
              ))}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,24 @@
...
+  return (
+    <section className="TodoList">
+      {todos.map(
+        todo => (
+          <TodoInfo
+            todo={todo}
+            user={users.find(user => todo.userId === user.id) || users[0]}
+          />
+        ),
+      )}
```
Viktor-Kost, line 21:
```suggestion
       ))}
```

Viktor-Kost (changes requested):
Nice!
Yet, look at previous mentor comment. I left it as unresolved with a hint "is not fixed"

src/components/TodoList/TodoList.tsx
MykolaKaradzha: Unfortunately, linter don`t like this fix(.

src/App.tsx
```diff
@@ -1,61 +1,107 @@
...
+  const MY_REGEX = /[^a-z-а-я-0-9-' ']/gi;
+
+  const [todos, setTodos] = useState<Todo[]>(todosFromServer);
+  const [userId, setUserId] = useState<number>(0);
+  const [title, setTitle] = useState<string>('');
+  const [errorTitle, setErrorTitle] = useState<boolean>(false);
+  const [errorUser, setErrorUser] = useState<boolean>(false);
+
+  const addUser = () => {
+    const newId = Math.max(...todos.map(({ id }) => id)) + 1;
```
OlEzhA0, line 18: it's bad practice to use latest max id, just create counter with `useState` and increase it

src/App.tsx
```diff
@@ -1,61 +1,107 @@
...
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import React, { useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/Todo';
 
 export const App = () => {
+  const MY_REGEX = /[^a-z-а-я-0-9-' ']/gi;
```
OlEzhA0, line 9: move MY_REGEX outside your component

src/App.tsx
```diff
@@ -1,61 +1,107 @@
...
+
+  const [todos, setTodos] = useState<Todo[]>(todosFromServer);
+  const [userId, setUserId] = useState<number>(0);
+  const [title, setTitle] = useState<string>('');
+  const [errorTitle, setErrorTitle] = useState<boolean>(false);
+  const [errorUser, setErrorUser] = useState<boolean>(false);
```
OlEzhA0, line 15:
```suggestion
  const [userId, setUserId] = useState(0);
  const [title, setTitle] = useState('');
  const [errorTitle, setErrorTitle] = useState(false);
  const [errorUser, setErrorUser] = useState(false);
```

src/App.tsx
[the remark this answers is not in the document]
OlEzhA0: hey! you can create separate validation function, check all errors there
and in your submit function 
```
if (!isValidData()) {
  return
}
```

OlEzhA0 (changes requested):
your demo link unavailable

src/App.tsx
MykolaKaradzha: I`ll fix it, though it was stated in the readme to make it this way.)

src/App.tsx
```diff
@@ -1,61 +1,118 @@
...
+const MY_REGEX = /[^a-z-а-я-0-9-' ']/gi;
 
 export const App = () => {
+  const [todos, setTodos] = useState(todosFromServer);
+  const [userId, setUserId] = useState(0);
+  const [title, setTitle] = useState('');
+  const [errorTitle, setErrorTitle] = useState(false);
+  const [errorUser, setErrorUser] = useState(false);
+  const [counter, setCounter] = useState(
+    Math.max(...todos.map(({ id }) => id)) + 1,
```
OleziO, line 16: `Math.max(...todos) + 1` will be the same.

src/App.tsx
```diff
@@ -1,61 +1,118 @@
...
+  const addUser = () => {
+    setTodos((prevState) => (
+      [
+        ...prevState,
+        {
+          id: counter,
+          title,
+          completed: false,
+          userId,
+        }]));
```
OleziO, line 45: Pass this object to the function props.

src/App.tsx
```diff
@@ -1,61 +1,118 @@
...
+      return;
+    }
+
+    addUser();
+    resetFields();
+    setCounter(state => state + 1);
+  };
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value.replace(MY_REGEX, ''));
```
OleziO, line 60: Why you need this replace() ?

src/App.tsx
```diff
@@ -1,61 +1,118 @@
...
+              name="user"
+              data-cy="userSelect"
+              value={userId}
+              onChange={handleSelect}
+            >
+              <option value="0" disabled>Choose a user</option>
+              {
+                usersFromServer.map(user => (
+                  <option value={user.id} key={user.id}>{user.name}</option>))
+              }
```
OleziO, line 101: Format this code.

OleziO (changes requested):

src/App.tsx
MykolaKaradzha: It`s an optional task:  "Allow entering only letters (ru and en), digits, and spaces in the title."

vlad-demchuk (approved):

PR #1494, opened 2023-02-17
Develop

Deniso4ka23:
[DEMO LINK](https://Deniso4ka23.github.io/react_add-todo-form/)


src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
+        <button
+          type="submit"
+          data-cy="submitButton"
+          onClick={() => {
+            if (selectUserId !== '0' && title) {
+              return addTodo();
+            }
+
+            return setSubmit(true);
+          }}
```
ivbutsykin, line 101: Better to move it in separate function

src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
+  return foundUser || null;
+}
+
+const todos = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [selectUserId, setSelectUserId] = useState('0');
```
ivbutsykin, line 20: Type of your user id is `number`. No need to save other type in state.
Use this `setSelectUserId(+e.target.value);` when you set your user id

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,27 @@
...
+
+type Props = {
+  todo: Todo;
+  largestId: number;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo, largestId }) => {
+  return (
+    <article
+      data-id={largestId + 1}
```
ivbutsykin, line 14: Why we need `largestId` and can't simple use `todo.id`? Try not to pass `largestId` prop it seems strange

ivbutsykin (changes requested):

src/components/TodoInfo/TodoInfo.tsx
Deniso4ka23: image

As I understand it, I need to find the largest id in the list of all cases and add 1 to it, and then assign each component to the date id. Right? If you assign todo.id, then all numbers will be different and the test will fail.

Deniso4ka23:
Corrected the passed value in data-id. But as I said in the test, the error is because it requires the highest value + 1.
image

anstsot (approved):
Looks good from the code perspective

PR #1496, opened 2023-02-18
initial

kostasburnazaki:
[DEMO LINK](https://kostasburnazaki.github.io/react_add-todo-form/)


src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+
+export class App extends React.Component<{}, State> {
+  state = {
+    id: Math.max(...todos.map(todo => todo.id)) + 1,
+    title: '',
+    userId: 0,
+    completed: false,
+  };
+
+  inputStatusMonitoring = false; // this property used in order to comply with this requirement: "errors should appear only after clicking the `Add` button"
```
Volodymyr-Melnyk-Fron10der, line 20: move it into state

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+
+  handleChange = (event:
+  React.FormEvent<HTMLInputElement>
+  | React.ChangeEvent<HTMLSelectElement>) => { // 2 types for events of input change ang select tag change
+    const { name, value } = event.target as HTMLInputElement;
+
+    if (value.match(/^[A-Za-zА-Яа-я0-9 ]*$/i)) {
```
Volodymyr-Melnyk-Fron10der, line 27: move regexp into constant

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+      this.setState((prevState => ({ // used prevState to avoid type script error (2345)
+        ...prevState,
+        [name]: value,
+      })));
+    }
+  };
+
+  handleSubmit = () => {
```
Volodymyr-Melnyk-Fron10der, line 35:
```suggestion
  handleSubmit() {
```

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+    if (this.state.userId && this.state.title.length) {
+      todos.push({
```
Volodymyr-Melnyk-Fron10der, line 37: create state for todos. Don't mutate initial array

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+    }
+  };
+
+  iterateEventId = () => {
+    this.setState(state => ({
+      id: state.id + 1,
+    }));
+  };
+
+  clearTheForm = () => {
```
Volodymyr-Melnyk-Fron10der, line 55:
```suggestion
  handleResetForm= () => {
```

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+      <div className="App">
+        <h1>Add todo form</h1>
+
+        <form
+          action="/api/users"
+          method="POST"
+          onSubmit={(e) => {
+            e.preventDefault();
+            this.handleSubmit();
+          }}
```
Volodymyr-Melnyk-Fron10der, line 74: move prevent into handleSubmit function
```suggestion
          onSubmit={this.handleSubmit}
```

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+              User:&nbsp;
+              <select
+                data-cy="userSelect"
+                name="userId"
+                value={this.state.userId}
+                onChange={this.handleChange}
+              >
+                <option value="">Choose a user</option>
+                {users.map((user: User) => (
+                  <option value={user.id}>{user.name}</option>
```
Volodymyr-Melnyk-Fron10der, line 104: add key

Volodymyr-Melnyk-Fron10der (changes requested):
Hi, 

1. Check reference, this option should be disabled
image
image
2. I can add todo with empty field 
image

src/App.tsx
kostasburnazaki: I've done it to avoid manual definition of the type of the event, to minimize error chances. So the compact code is the priority?

src/App.tsx
kostasburnazaki: In general as I understand I must avoid using such kind of properties? Or is there any case they are more preferable than state properties?

src/App.tsx
```diff
@@ -1,61 +1,158 @@
...
+export class App extends React.Component<{}, State> {
+  state = {
+    currentTodo: {
+      id: Math.max(...todos.map(todo => todo.id)) + 1,
+      title: '',
+      userId: 0,
+      completed: false,
+    },
+    todos: [...todos],
+    re: /^[A-Za-zА-Яа-я0-9 ]*$/i,
```
ivbutsykin, line 30: No need to save this regex in state, it should be constant

src/App.tsx
```diff
@@ -1,61 +1,158 @@
...
+    inputStatusMonitoring: false,
+  };
+
+  handleChange = (event:
+  React.FormEvent<HTMLInputElement>
+  | React.ChangeEvent<HTMLSelectElement>) => {
+    const { name, value } = event.target as HTMLInputElement;
```
ivbutsykin, line 37: Do we need `as HTMLInputElement`? If yes, it's a problem with types. Or try to rewrite unused this `as HTMLInputElement`, or create two different functions for select and input change.

src/App.tsx
```diff
@@ -1,61 +1,158 @@
...
+            newTodo,
+          ],
+        };
+      });
+
+      this.handleResetForm();
+    } else {
+      this.state.inputStatusMonitoring = true;
+
+      this.forceUpdate();
```
ivbutsykin, line 76: Do we need `forceUpdate` here? You should avoid using this: https://reactjs.org/docs/react-component.html#forceupdate

So try to rewrite avoiding this

src/App.tsx
```diff
@@ -1,61 +1,158 @@
...
+    this.setState(prevState => ({
+      ...prevState,
+      currentTodo: {
+        ...prevState.currentTodo,
+        title: '',
+        userId: 0,
+      },
+    }));
+
+    this.state.inputStatusMonitoring = false;
```
ivbutsykin, line 90: I think, we need to call `this.setState` here

src/App.tsx
```diff
@@ -1,61 +1,158 @@
...
+          todos: [
+            ...prevState.todos,
+            newTodo,
+          ],
+        };
+      });
+
+      this.handleResetForm();
+    } else {
+      this.state.inputStatusMonitoring = true;
```
ivbutsykin, line 74: Same here, `this.setState`

ivbutsykin (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,159 @@
...
+    },
+    todos: [...todos],
+    inputStatusMonitoring: false,
+  };
+
+  handleChange = (
+    event: React.ChangeEvent<HTMLInputElement | HTMLSelectElement>,
+  ) => {
+    const { name, value } = event.target;
+    const re = /^[A-Za-zА-Яа-я0-9 ]*$/i;
```
Volodymyr-Melnyk-Fron10der, line 36: **(JUST FYI)** I would move it into `src/constnats` and name it like `LETTERS_NUMBERS_REGEXP`

Volodymyr-Melnyk-Fron10der (approved):
Well done✅

PR #1500, opened 2023-02-26
finished

barantarasNEW:
https://barantarasNEW.github.io/react_add-todo-form/


VitaliyBondarenko1982 (approved):
Good job.
Just please pay attention on anonimous functions in JSX.
It is bad practice.

PR #1502, opened 2023-02-27
React_add-todo-form

07Spirit:
[DEMO LINK](https://07Spirit.github.io/react_add-todo-form/)


qoa11a (changes requested):
user shouldn't get errors right after adding a todo. Only if user clicks "Add" with empty inputs

witflash (approved):
Well done 👍

PR #1526, opened 2023-03-12
solution

vlad-trundenko:
[DEMO LINK](https://vlad-trundenko.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,138 @@
 import './App.scss';
+import React, { useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
 
-export const App = () => {
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,61 +1,146 @@
...
-// import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+function getUser(userId: number): User | null {
```
polosanya, line 12:
```suggestion
function getUserById(userId: number): User | null {
```
It is better to specify it in the name

src/App.tsx
```diff
@@ -1,61 +1,146 @@
...
+  user: getUser(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [todos, setTodo] = useState(addUserTodo);
+
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+
+  const [validInputTitle, setValidInputTitle] = useState(false);
```
polosanya, line 30:
```suggestion
  const [isValidTitle, setIsValidTitle] = useState(false);
```
Name boolean variables according to these rules:
image

polosanya (changes requested):
Almost done! Let's improve your code a bit)

anstsot (approved):
Great job, approved!

PR #1531, opened 2023-03-14
solution

milenadiamonds1:


src/App.tsx
```diff
@@ -1,61 +1,135 @@
...
+    if (title.trim() && selectedUserId) {
+      const newTodo: Todo = {
+        id: Math.max(...todos.map((todo) => todo.id)) + 1,
+        userId: selectedUserId,
+        title,
+        completed: false,
+        user: getUser(selectedUserId),
+      };
+
+      setTodos([...todos, newTodo]);
```
timoxa2t, line 53: It's better to use callback function when we use previous state
```suggestion
      setTodos(prev => [...prev, newTodo]);
```

timoxa2t (approved):
Good work

src/App.tsx
illiarerun: agree with Sasha, use functional setter here

src/App.tsx
```diff
@@ -1,61 +1,135 @@
...
-            <option value="0" disabled>Choose a user</option>
-          </select>
+          <label htmlFor="userSelect">
+            User:
+            <select
+              data-cy="userSelect"
+              id="userSelect"
+              placeholder="Choose a user"
+              value={selectedUserId}
+              onChange={(event) => handleUser(event)}
```
illiarerun, line 106: use shortened version, just pass your callback

illiarerun (changes requested):

illiarerun (approved):

PR #1533, opened 2023-03-15
solution

sofiatryasko:
[DEMO LINK](https://sofiatryasko.github.io/react_add-todo-form/)


src/App.tsx
```diff
@@ -1,60 +1,95 @@
...
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [list, setList] = useState(todos);
+  const [input, setInput] = useState('');
+  const [userId, setUserId] = useState('');
+
+  const addTodo = (event: any) => {
```
tetiana-ivanchenko, line 28: don't use any as type

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,16 @@
...
+import { TodoInfo } from '../TodoInfo';
+import { Todo } from '../types/Todo';
+
+type Todos = {
+  todos: Todo[];
+};
+export const TodoList = ({ todos } : Todos) => (
+  <ul>
+    {todos.map(todo => (
+      <li data-id={todo.id} className={`TodoInfo ${todo.completed ? 'TodoInfo--completed' : ''}`}>
```
tetiana-ivanchenko, line 10: add key attr

src/App.tsx
```diff
@@ -1,60 +1,95 @@
...
-          <span className="error">Please enter a title</span>
+          Title:
+          <input
+            type="text"
+            data-cy="titleInput"
+            onChange={event => setInput(event.target.value)}
+          />
+          <span
+            className="error"
+            style={{ display: input === '' ? 'inline' : 'none' }}
```
tetiana-ivanchenko, line 55: change classes instead of using inline styles

tetiana-ivanchenko (changes requested):
1. I can add todo whrn the title and the user are empty
2.  I don't see an error if my title contains only spaces
image

src/App.tsx
sofiatryasko: could you advise which type of event I should write?

tetiana-ivanchenko (changes requested):
image
please, before requesting a review, check:
1. if all points from the task are done
2. if all points from the checklist done
3. if all your links are working (demo, test report, if exist)
4. if it's not your first review - if all comments fixed

mykhalenych (approved):

PR #1536, opened 2023-03-18
● Solution

d-hubych:
- [DEMO LINK](https://d-hubych.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,123 @@
+import { useState } from 'react';
 import './App.scss';
+import { TodoList } from './components/TodoList';
+import { User } from './components/types/User';
+import { Todo, TodoForRender } from './components/types/Todo';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+const localUsers: User[] = usersFromServer.map(user => ({ ...user }));
+let localTodos: Todo[] = todosFromServer.map(todo => ({ ...todo }));
+
+function prepareTodos(): TodoForRender[] {
+  return (
+    localTodos.map(todo => {
+      return ({
+        ...todo,
+        user: localUsers.find(user => user.id === todo.userId) || null,
+      });
+    })
+  );
+}
 
 export const App = () => {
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+import { TodoList } from './components/TodoList';
+import { User } from './components/types/User';
+import { Todo, TodoForRender } from './components/types/Todo';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+const localUsers: User[] = usersFromServer.map(user => ({ ...user }));
```
OleziO, line 10: You don't need this map, just set the data to the state.

OleziO: Check other similar cases.

OleziO (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,132 @@
...
 export const App = () => {
+  const [todosToRender, setTodosToRender] = useState(initialData);
+  const [newTitle, setNewTitle] = useState('');
+  const [selectedUserId, setSelectedUserId] = useState('');
+  const [errTitle, setErrTitle] = useState(false);
+  const [errUser, setErrUser] = useState(false);
+
+  const todoUpdater = () => {
+    const maxId = todosToRender.reduce((max, { id }) => {
+      return (max < id) ? id : max;
```
VitaliyBondarenko1982, line 31:
```suggestion
      return Math.max(max, id);
```

src/App.tsx
```diff
@@ -1,61 +1,132 @@
...
-            <option value="0" disabled>Choose a user</option>
-          </select>
+          <label>
+            <select
+              data-cy="userSelect"
+              value={selectedUserId}
+              onChange={(event) => {
+                setSelectedUserId(event.target.value);
+                setErrUser(false);
+              }}
```
VitaliyBondarenko1982, line 104: it is bad practice to create anonymous arrow functions in jsx
better to create function in component and pass reference here

VitaliyBondarenko1982 (approved):

PR #1546, opened 2023-03-25
add todo

Mykola-Busarhin:
[DEMO LINK](https://Mykola-Busarhin.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,119 @@
 import './App.scss';
+import { useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { User } from './components/types/User';
+import { Todo } from './components/types/Todo';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+function findUser(todoId: number): User | null {
+  const user = usersFromServer.find(person => person.id === todoId);
+
+  return user || null;
+}
+
+const todosWithUser: Todo[] = todosFromServer.map(todo => (
+  {
+    ...todo,
+    user: findUser(todo.userId),
+  }
+));
 
 export const App = () => {
+  const [title, setNewTitle] = useState('');
+  const [userId, setNewId] = useState(0);
+  const [userSelected, isUserSelected] = useState(true);
+  const [titleCreated, isTitleCreated] = useState(true);
+
+  const maxTodoId = Math.max(...todosWithUser.map(todo => todo.id)) + 1;
+
+  const addNewTodo = (todoId: number) => {
+    const newUser = usersFromServer.find(user => user.id === todoId);
+
+    if (newUser && title) {
+      todosWithUser.push({
+        id: maxTodoId,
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,61 +1,119 @@
...
 export const App = () => {
+  const [title, setNewTitle] = useState('');
+  const [userId, setNewId] = useState(0);
+  const [userSelected, isUserSelected] = useState(true);
+  const [titleCreated, isTitleCreated] = useState(true);
+
+  const maxTodoId = Math.max(...todosWithUser.map(todo => todo.id)) + 1;
+
+  const addNewTodo = (todoId: number) => {
+    const newUser = usersFromServer.find(user => user.id === todoId);
```
Volodymyr-Melnyk-Fron10der, line 31: i don't understand why todoId === user.id ? And why it's named as newUser? if it is already exist?

Volodymyr-Melnyk-Fron10der: what is common between user and todo id?

src/App.tsx
```diff
@@ -1,61 +1,119 @@
...
+
+    if (newUser && title) {
+      todosWithUser.push({
```
Volodymyr-Melnyk-Fron10der, line 34: create state for it and change the state

src/App.tsx
```diff
@@ -1,61 +1,119 @@
...
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
+        action="/api/users"
+        method="POST"
+        onSubmit={(event) => {
+          event.preventDefault();
+          addNewTodo(userId);
+        }}
```
Volodymyr-Melnyk-Fron10der, line 65: move into function , fix everywhere

Volodymyr-Melnyk-Fron10der (changes requested):

src/App.tsx
Mykola-Busarhin: todoAdd-form-userInfo
I am trying to find a user who created a new to do ( as fas as I get it, todo`s userId and user`s id are the same numbers, actually todoId is userId cause it is passed as an argument in addNewTodo function). Then I use found user as a value for key 'user' in each new todo. Smth like this. As for the name 'newUser', would it be better if i rename it to 'newTodocreator' for example to make it more explicit?

src/App.tsx
```diff
@@ -1,61 +1,130 @@
...
 export const App = () => {
+  const [title, setNewTitle] = useState('');
+  const [userId, setNewId] = useState(0);
+  const [userSelected, isUserSelected] = useState(true);
+  const [titleCreated, isTitleCreated] = useState(true);
+  const [todoList, updateTodoList] = useState(todosWithUser);
+
+  const maxTodoId = Math.max(...todoList.map(todo => todo.id)) + 1;
+
+  const addNewTodo = (todoId: number) => {
```
OleziO, line 31: Rename to something like `userId`

OleziO (changes requested):

OleziO (approved):

PR #1563, opened 2023-03-27
react_add-todo-form solution

stanislavkulakovskyi:
[DEMO LINK](https://stanislavkulakovskyi.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React, { useState, useEffect, ChangeEvent } from 'react';
 import './App.scss';
+import { v4 as uuidv4 } from 'uuid';
 
 import usersFromServer from './api/users';
 import todosFromServer from './api/todos';
@@ -56,10 +57,10 @@ export const App = () => {
   ) => {
     const { name, value } = event.target;
 
-    setValues({
-      ...values,
+    setValues(prevValues => ({
+      ...prevValues,
       [name]: value,
-    });
+    }));
   };
 
   useEffect(() => {
@@ -76,14 +77,16 @@ export const App = () => {
     setTodo({
       ...todo,
       id,
-      userId: addUser !== null ? addUser.id : 0,
+      userId: addUser !== null
+        ? addUser.id
+        : uuidv4(),
       title,
       completed,
       user: getUserByName(user),
     });
   }, [title, user]);
 
-  const handleSubmit = (event: React.FormEvent) => {
+  const handleAddingTodo = (event: React.FormEvent) => {
     event.preventDefault();
 
     if (!user || !title) {
@@ -125,7 +128,7 @@ export const App = () => {
       <form
         action="/api/users"
         method="POST"
-        onSubmit={handleSubmit}
+        onSubmit={handleAddingTodo}
       >
         <div className="field">
           <label>
--- a/package-lock.json
+++ b/package-lock.json
@@ -1282,6 +1282,12 @@
             "psl": "^1.1.28",
             "punycode": "^2.1.1"
           }
+        },
+        "uuid": {
+          "version": "8.3.2",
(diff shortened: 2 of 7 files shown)


MarkBabich (approved):
GJ :)

src/App.tsx
```diff
@@ -1,61 +1,178 @@
...
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+
+const getUserById = (userId: number) => {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
```
Andr1yK, line 13:
```suggestion
  const user = usersFromServer.find(user => user.id === userId);

  return user || null;
```

src/App.tsx
```diff
@@ -1,61 +1,178 @@
...
+};
+
+const getUserByName = (username: string) => {
+  const foundUser = usersFromServer.find(user => user.name === username);
+
+  return foundUser || null;
+};
```
Andr1yK, line 20: What happened if we will have more than one user with this name? use better way to find user)

src/App.tsx
```diff
@@ -1,61 +1,178 @@
...
+
+  const handleChange = (
+    event: ChangeEvent<HTMLInputElement | HTMLSelectElement>,
+  ) => {
+    const { name, value } = event.target;
+
+    setValues({
+      ...values,
+      [name]: value,
+    });
```
Andr1yK, line 62: In ReactJS, it’s recommended to use functional `setState` instead of the regular `setState`. The reason for this is that the functional form of `setState` can help avoid issues with stale state, and ensure that the state is updated correctly.

Here are some rules to consider when using functional `setState` in React:

1. **Always use functional `setState` when updating state based on previous state:** If you need to update the state based on its previous value, it’s recommended to use functional `setState` to ensure that the state is updated correctly. Here’s an example:

```jsx
// using functional setState
this.setState(prevState => ({ count: prevState.count + 1 }));
```

2. **Use regular `setState` for simple updates:** If you’re updating the state with a simple value, such as a boolean or a string, you can use regular `setState`. Here’s an example:

```jsx
// using regular setState
this.setState({ isModalOpen: true });
```

3. **Use functional `setState` for complex state updates:** If you’re updating the state with a complex value, such as an array or an object, it’s recommended to use functional `setState` to ensure that the state is updated correctly. Here’s an example:

```jsx
// using functional setState with object spread syntax
this.setState(prevState => ({ ...prevState, items: newItems }));
```

4. **Be careful with nested state updates:** If you have nested state updates, it’s important to use functional `setState` to ensure that the state is updated correctly. Here’s an example:

```jsx
// incorrect use of setState with nested state updates
this.setState({ items: [] });
this.setState(prevState => ({ items: [...prevState.items, newItem] }));

// correct use of functional setState with nested state updates
this.setState(prevState => ({ items: [...prevState.items, newItem] }));
```

In summary, using functional `setState` in ReactJS can help avoid issues with stale state and ensure that the state is updated correctly. Always use functional `setState` when updating state based on its previous value, and be careful with nested state updates.

src/App.tsx
```diff
@@ -1,61 +1,178 @@
...
+    }
+
+    if (title.length > 0) {
+      setErrorTitle(false);
+    }
+
+    setTodo({
+      ...todo,
+      id,
+      userId: addUser !== null ? addUser.id : 0,
```
Andr1yK, line 79:
```suggestion
      userId: addUser !== null
        ? addUser.id
        : 0,
```

Andr1yK, line 79: wow.. super bad way.. avoid adding 0 as userId

Andr1yK (changes requested):

PR #1570, opened 2023-03-27
solution 1.0

olecsiy-gavrilyuk:
[DEMO LINK](https://olecsiy-gavrilyuk.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,116 @@
+import React, { ChangeEvent, useState } from 'react';
 import './App.scss';
+import { TodoList } from './components/TodoList';
+import { User } from './components/types/User';
+import { Todo } from './components/types/Todo';
 
-// import usersFromServer from './api/users';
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,61 +1,124 @@
...
+  const [selectedTodos, setSelectedTodos] = useState(todos);
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (!userId || !title.trim()) {
+      setIsGoodToSubmit(true);
+    }
+
+    if (userId && title.trim()) {
```
Andr1yK, line 38: Instead of checking if `userId` exists, it is better to check if `user` exists after calling `getUserById(userId)` because it ensures that a valid user is selected. Change the condition to use `user` object. 🕵️

Andr1yK (changes requested):

PR #1582, opened 2023-04-01
the task was solved

batiukova-nataliia:
[DEMO LINK](https://batiukova-nataliia.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,103 @@
 import './App.scss';
+import React, { useState } from 'react';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  // if there is no user with a given userId
+  return foundUser || null;
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [option, setOption] = useState(0);
+  const [title, setTitle] = useState('');
+  const [initialTodos, setNewTodo] = useState(todosFromServer
+    .map(todo => ({ ...todo, user: getUser(todo.userId) })));
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    const newTodo: Todo = {
+      id: Math.max(...initialTodos.map(todo => todo.id)) + 1,
+      title,
+      completed: false,
+      userId: option,
+      user: getUser(option),
+    };
+
+    setNewTodo([...initialTodos, newTodo]);
+  };
+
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,103 @@
...
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [option, setOption] = useState(0);
+  const [title, setTitle] = useState('');
+  const [initialTodos, setNewTodo] = useState(todosFromServer
+    .map(todo => ({ ...todo, user: getUser(todo.userId) })));
```
VitaliyBondarenko1982, line 26:
```suggestion
  const [todos, setTodos] = useState(initialTodos);
```
it will be more understandable naming

src/App.tsx
```diff
@@ -1,61 +1,103 @@
...
+import todosFromServer from './api/todos';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  // if there is no user with a given userId
+  return foundUser || null;
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
```
VitaliyBondarenko1982, line 17:
```suggestion
export const initialTodos: Todo[] = todosFromServer.map(todo => ({
```

src/App.tsx
```diff
@@ -1,61 +1,103 @@
...
+    event.preventDefault();
+
+    const newTodo: Todo = {
+      id: Math.max(...initialTodos.map(todo => todo.id)) + 1,
+      title,
+      completed: false,
+      userId: option,
+      user: getUser(option),
+    };
+
```
VitaliyBondarenko1982, line 38: add validation for inputs
Screenshot from 2023-04-01 17-25-43

VitaliyBondarenko1982 (changes requested):

etojeDenys (changes requested):
let's add this useState errors and set them if user tries to submit with empty fields 
```
  const [selectUserError, setSelectUserError] = useState(false);
```
2. also let's reset our fields after submitting

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+    }
+  }, [title]);
+
+  useEffect(() => {
+    if (option === 0) {
+      setSelectUserError(true);
+    } else {
+      setSelectUserError(false);
+    }
+  }, [option]);
```
etojeDenys, line 43: let's show errors when user submits the form, not when he only opens the page

so we need to make this calculation in handleSubmit

etojeDenys (changes requested):

OleziO (changes requested):
Don't show errors before the user first click on submit button.

src/App.tsx
```diff
@@ -1,61 +1,120 @@
...
+            Title:
+            {' '}
+            <input
+              type="text"
+              data-cy="titleInput"
+              placeholder="Enter a title"
+              value={title}
+              onChange={titleChange}
+            />
+            {title === ''
```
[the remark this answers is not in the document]
batiukova-nataliia: it doesn't work properly, so I replaced it by tittleError

IvanFesenko (approved):

PR #1597, opened 2023-04-05
react todo form v1

fenderukr:
https://fenderukr.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,102 @@
 import './App.scss';
+import { useState } from 'react';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [user, setUser] = useState('0');
+  const [isClicked, setIsClicked] = useState(false);
+  const [todos, setTodos] = useState(todosFromServer);
+
+  const clearForm = () => {
+    setUser('0');
+    setTitle('');
+    setIsClicked(false);
+  };
+
+  const handleSubmit
+   = (event:React.SyntheticEvent) => {
+     event.preventDefault();
+     setIsClicked(true);
+
+     if (!title || user === '0') {
+       return;
+     }
+
+     const id: number = todosFromServer.sort((a, b) => b.id - a.id)[0].id + 1;
+
+     setTodos((prevState) => [...prevState, {
+       id,
+       title,
+       completed: false,
+       userId: +user,
+     }]);
+
+     clearForm();
+   };
+
+  const handleTitleChange: React.ChangeEventHandler<HTMLInputElement>
+    = (event) => {
+      setTitle(event.target.value);
+    };
+
+  const handleUserChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setUser(event.target.value);
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
+        action="/api/users"
+        method="POST"
+        onSubmit={handleSubmit}
+      >
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <label>
+            Title:
+            <input
+              type="text"
+              data-cy="titleInput"
+              placeholder="Please enter a title"
+              value={title}
+              onChange={handleTitleChange}
+            />
+
+            {(!title && isClicked)
+            && <span className="error">Please enter a title</span>}
+          </label>
         </div>
 
         <div className="field">
-          <select data-cy="userSelect">
+          <label htmlFor="userSelect">User: </label>
+          <select
+            data-cy="userSelect"
+            defaultValue="0"
+            id="userSelect"
+            value={user}
+            onChange={handleUserChange}
+          >
             <option value="0" disabled>Choose a user</option>
+            {usersFromServer.map((userFromServer) => {
+              return (
+                <option
+                  value={userFromServer.id.toString()}
+                  key={userFromServer.id}
+
+                >
+                  {userFromServer.name}
+                </option>
+              );
+            })}
           </select>
 
-          <span className="error">Please choose a user</span>
+          {(user === '0' && isClicked)
+          && <span className="error">Please choose a user</span>}
         </div>
 
         <button type="submit" data-cy="submitButton">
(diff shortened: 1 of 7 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,18 @@
...
+interface TodoListInt {
+  todos: Todo[]
+}
+
+export const TodoList = (props: TodoListInt) => {
+  const { todos } = props;
+
+  return (
+    <section className="TodoList">
+      {todos.map((todo: Todo) => (
```
agniya-i, line 13: Consider adding a `key` prop to the `TodoInfo` component to improve performance. 🚀

src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
+  const handleSubmit
+   = (event:React.SyntheticEvent) => {
+     event.preventDefault();
+     setIsClicked(true);
+
+     if (!title || user === '0') {
+       return;
+     }
+
+     const id: number = todosFromServer.sort((a, b) => b.id - a.id)[0].id + 1;
```
agniya-i, line 29: Consider using a more descriptive variable name instead of `id`. For example, `newTodoId`. 🤔

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,35 @@
...
+  const getUserById = (id: number): User => {
+    return usersFromServer.find((user) => user.id === id) as User;
+  };
+
+  return (
+
+    <article
+      key={todo.id}
+      data-id={todo.id}
+      className={todo.completed ? 'TodoInfo TodoInfo--completed' : 'TodoInfo'}
```
agniya-i, line 26: It's a perfect case to use `classNames` 

```suggestion
      className={classNames('TodoInfo', {
        'TodoInfo--completed': todo.completed,
      })}
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,35 @@
...
+
+export const TodoInfo = ({ todo }: TodoInfoInt) => {
+  const getUserById = (id: number): User => {
+    return usersFromServer.find((user) => user.id === id) as User;
+  };
+
+  return (
+
+    <article
+      key={todo.id}
```
agniya-i, line 24: remove `key`
you should set this property only inside `.map()`

src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
+            value={user}
+            onChange={handleUserChange}
+          >
             <option value="0" disabled>Choose a user</option>
+            {usersFromServer.map((userFromServer) => {
+              return (
+                <option
+                  value={userFromServer.id.toString()}
+                  key={userFromServer.id}
+
```
agniya-i, line 90:
```suggestion
```

agniya-i (changes requested):

src/components/TodoInfo/TodoInfo.tsx
fenderukr: hm, i am not sure that I understand the difference

could you send me a few links to the resources about the difference in using my approach and classnames?

just want to make sure that I understand what's happening here

TarasHoliuk: There is an article about classnames: [link](https://medium.com/2359media/why-react-classnames-is-a-lifesaver-and-how-you-should-use-it-if-you-arent-already-b25a44ed5603) - about general advantages

I'd say with `classNames` it's more clear (visually) that `TodoInfo` class exists always and `--completed` is conditional
But it's not critical here

src/App.tsx
```diff
@@ -1,61 +1,111 @@
...
+            defaultValue="0"
+            id="userSelect"
+            value={user}
+            onChange={handleUserChange}
+          >
             <option value="0" disabled>Choose a user</option>
+            {usersFromServer.map((userFromServer) => {
+              return (
+                <option
+                  value={userFromServer.id.toString()}
```
TarasHoliuk, line 90: why stringified?

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,37 @@
-export const TodoInfo = () => {};
+import classNames from 'classnames';
+import { Todo } from '../../types/todo';
```
TarasHoliuk, line 2: add this file to git

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,37 @@
...
+  email: string;
+}
+
+export const TodoInfo = ({ todo }: TodoInfoInt) => {
+  const getUserById = (id: number): User => {
+    return usersFromServer.find((user) => user.id === id) as User;
+  };
+
+  return (
+
```
TarasHoliuk, line 23:
```suggestion
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,37 @@
...
+interface User {
+  id: number;
+  name: string;
+  username: string;
+  email: string;
+}
+
+export const TodoInfo = ({ todo }: TodoInfoInt) => {
+  const getUserById = (id: number): User => {
+    return usersFromServer.find((user) => user.id === id) as User;
```
TarasHoliuk, line 19:
```suggestion
    return usersFromServer.find((user) => user.id === id) || null;
```
it may not find the user, then it will lead to a page crash
it's better to handle the case when result of call `getUserById` returns `null`

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,18 @@
-export const TodoList = () => {};
+import { Todo } from '../../types/todo';
```
TarasHoliuk, line 1: add this file

TarasHoliuk (changes requested):

src/components/TodoInfo/TodoInfo.tsx
fenderukr: @TarasHoliuk not sure what I should do herE:
image

src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState('0');
+  const [isClicked, setIsClicked] = useState(false);
```
mykhalenych, line 11: Consider using a more descriptive variable name like 'isFormSubmitted' instead of 'isClicked' to make it clear what the variable represents.

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,36 @@
...
+
+interface User {
+  id: number;
+  name: string;
+  username: string;
+  email: string;
+}
+
+export const TodoInfo = ({ todo }: TodoInfoProps) => {
+  const getUserById = (id: number): User => {
```
mykhalenych, line 18: The `getUserById` function should be moved outside of the `TodoInfo` component to prevent it from being recreated on every render.

src/components/TodoInfo/TodoInfo.tsx
mykhalenych: didn't fix it, I don' see folder types 
Screenshot 2023-04-21 at 16 26 51

src/components/TodoList/TodoList.tsx
mykhalenych: the same

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,18 @@
...
+import { TodoInfo } from '../TodoInfo';
+
+interface TodoListInt {
```
mykhalenych, line 4: keep interfaces in folder types

lerastarynets (changes requested):
Well done with fixing comments) However now user can add a todo consisting of white spaces only, pls fix

PR #1601, opened 2023-04-06
add-todo-form was done

pushkarskiyrodion:
[DEMO LINK](https://%3Cyour_account%3E.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,141 @@
+import { useEffect, useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+
+import { ToDo } from './types/ToDo';
+
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+const todosInfo: ToDo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: usersFromServer.find(user => user.id === todo.userId),
+}));
 
 export const App = () => {
+  const [todos, setTodos] = useState(todosInfo);
+  const [user, setUser] = useState('');
+  const [title, setTitle] = useState('');
+  const [hasUser, setHasUser] = useState(false);
+  const [hasTitle, setHasTitle] = useState(false);
+  const [formSubmitted, setFormSubmitted] = useState(false);
+
+  useEffect(() => {
+    if (title) {
+      setHasTitle(true);
+    }
+
+    if (user) {
+      setHasUser(true);
+    }
+  }, [title, user]);
+
+  function handleChange(e: React.BaseSyntheticEvent) {
+    const { name } = e.target;
+
+    if (name === 'title') {
+      setTitle(e.target.value);
+    }
+
+    if (name === 'user') {
+      setUser(e.target.value);
+    }
+  }
+
+  function handleSubmit(e: React.FormEvent) {
+    e.preventDefault();
+
+    setFormSubmitted(true);
+
+    if (!title) {
+      setHasTitle(false);
+    }
+
+    if (!user) {
+      setHasUser(false);
+    }
+
+    if (title && user) {
+      const newId = Math.max(...todos.map(todo => todo.id));
+      const userOfToDo = usersFromServer.find(userFromServer => (
+        userFromServer.name.includes(user)
+      ));
+
+      const newToDo = {
+        title,
+        completed: false,
+        userId: userOfToDo?.id,
+        id: newId + 1,
+        user: userOfToDo,
+      };
+
+      setTodos(current => ([
+        ...current,
+        newToDo,
+      ]));
+
+      setUser('');
+      setTitle('');
+      setHasTitle(false);
+      setHasUser(false);
+      setFormSubmitted(false);
+    }
+  }
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
+        action="/api/users"
+        method="POST"
+        onSubmit={handleSubmit}
+      >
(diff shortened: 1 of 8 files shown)


tetiana-ivanchenko (changes requested):
now I can add title with just spaces

src/App.tsx
```diff
@@ -1,61 +1,149 @@
...
+  function handleChange(e: React.BaseSyntheticEvent) {
+    const { name } = e.target;
+
+    if (name === 'title') {
+      setTitle(e.target.value);
+    }
+
+    if (name === 'user') {
+      setUser(e.target.value);
+    }
```
loralevitska, line 43:
```suggestion
    const { name, value } = e.target;

    if (name === 'title') {
      setTitle(value);
    }

    if (name === 'user') {
      setUser(value);
    }
```

src/App.tsx
```diff
@@ -1,61 +1,149 @@
...
+  }
+
+  function handleSubmit(e: React.FormEvent) {
+    e.preventDefault();
+
+    setFormSubmitted(true);
+
+    if (!title) {
```
loralevitska, line 51:
```suggestion
    if (!title.trim()) {
```

src/App.tsx
```diff
@@ -1,61 +1,149 @@
...
+              value={user}
+              onChange={handleChange}
+            >
+              <option value="" disabled>Choose a user</option>
+
+              {
+                usersFromServer.map(userFromServer => (
+                  <option key={userFromServer.id}>{userFromServer.name}</option>
+                ))
+              }
```
loralevitska, line 130:
```suggestion
              {usersFromServer.map(userFromServer => (
                  <option key={userFromServer.id}>{userFromServer.name}</option>
                ))}
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,29 @@
...
+
+import { UserInfo } from '../UserInfo';
+
+import { ToDo } from '../../types/ToDo';
+
+type Props = {
+  todo: ToDo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => (
```
loralevitska, line 11: Destructure `todo`

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,20 @@
...
+  <section className="TodoList">
+    {
+      todos.map(todo => (
+        <TodoInfo
+          todo={todo}
+          key={todo.id}
+        />
+      ))
+    }
+  </section>
```
loralevitska, line 19: Use semantic tags for list `ul` `li`. Don't forget to remove default `li` dots

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,11 @@
-export const UserInfo = () => {};
+import { User } from '../../types/User';
+
+type Props = {
+  user: User;
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => (
```
loralevitska, line 7: Destucture `user`

loralevitska (changes requested):
Almost perfect) Let's improve
Would be better to show the error after click to `add` button and field is empty
image

src/App.tsx
```diff
@@ -1,61 +1,154 @@
...
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+const todosInfo: ToDo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: usersFromServer.find(user => user.id === todo.userId),
+}));
 
 export const App = () => {
+  const [todos, setTodos] = useState(todosInfo);
```
Volodymyr-Melnyk-Fron10der, line 17: add a type
```suggestion
  const [todos, setTodos] = useState<ToDo[]>(todosInfo);
```

src/App.tsx
Volodymyr-Melnyk-Fron10der: previous comment was not fixed yet
```suggestion
    if (!title.trim()) {
```

src/App.tsx
```diff
@@ -1,61 +1,154 @@
...
+  const [user, setUser] = useState('');
+  const [title, setTitle] = useState('');
+  const [hasUser, setHasUser] = useState(false);
+  const [hasTitle, setHasTitle] = useState(false);
```
Volodymyr-Melnyk-Fron10der, line 21: are you sure you need hasUser and hasTitle states? You can just check is there some value inside user and title, isn't it enough?

src/App.tsx
```diff
@@ -1,61 +1,154 @@
...
+    if (!user) {
+      setHasUser(false);
+    }
+
+    if (title.startsWith(' ')) {
+      setHasTitle(false);
+      setTitle('');
+
+      return;
+    }
```
Volodymyr-Melnyk-Fron10der, line 64: don't clearly understand the purpose of this check actually))

Volodymyr-Melnyk-Fron10der (changes requested):
Hi! Almost there)

src/App.tsx
pushkarskiyrodion: You right. It's useless. Thanks for advice :)

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+    const { name, value } = e.target;
+
+    if (name === 'title') {
+      setTitle(value);
+    }
+
+    if (name === 'user') {
+      setUser(value);
+    }
+  }
```
Volodymyr-Melnyk-Fron10der, line 32: good for now, but for future - for different inputs - create separated handlers plz

Volodymyr-Melnyk-Fron10der (approved):
WellDoneGIF

PR #1603, opened 2023-04-11
Solution

DaveBeetle:
[DEMO LINK](https://DaveBeetle.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,148 @@
 import './App.scss';
+import React, { useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-export const App = () => {
+import { TodoList } from './components/TodoList';
+import { User } from './types/User';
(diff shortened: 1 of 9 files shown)


SerhiiKirik (changes requested):
Well done 🚀 
Let's fix some bugs 🧑‍💻 
I have an error when in the user field i want to add todo with an empty title ( only spaces )
image

src/App.tsx
```diff
@@ -1,61 +1,146 @@
...
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [newList, setNewList] = useState(todos);
+
+  const [showErrorTitle, setShowErrorTitle] = useState(false);
```
loralevitska, line 27: Please fix all boolean variables
```suggestion
  const [isShowErrorTitle, setIsShowErrorTitle] = useState(false);
```

loralevitska (approved):
Would be better to add an error message when todo field have only spaces and we click with this to `add` button

PR #1612, opened 2023-04-16
Dev

VladGraund:
[DEMO LINK](https://vladgraund.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,127 @@
+import { ChangeEvent, MouseEvent, useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { User } from './types/User';
+import { TodoList } from './components/TodoList';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+const todos = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [initTodos, setTodos] = useState(todos);
+  const [title, setTitle] = useState('');
+  const [emptyTitle, setEmptyTitle] = useState(false);
+  const [userId, setUserId] = useState(0);
+  const [emptyUser, setEmptyUser] = useState(false);
+
+  let count = 15;
+
+  const handleSelect = (event: ChangeEvent<HTMLSelectElement>) => {
+    setEmptyUser(false);
+    setUserId(+event.target.value);
+  };
+
+  const handleTitle = (event: ChangeEvent<HTMLInputElement>) => {
+    setEmptyTitle(false);
+    setTitle(event.target.value);
+  };
+
+  const createTodo = (event: MouseEvent) => {
+    event.preventDefault();
+
+    if (!title.trim()) {
+      setEmptyTitle(true);
+    }
+
+    if (!userId) {
+      setEmptyUser(true);
+    }
+
+    if (!title.trim() || userId === 0) {
+      return false;
+    }
+
+    count += 1;
+
+    const newTodo = {
+      id: count,
+      title,
+      completed: false,
+      userId,
+      user: getUser(userId),
+    };
+
+    const form = document.querySelector('form');
+
+    if (!form) {
+      return false;
+    }
+
+    form.reset();
+
+    return setTodos([...initTodos, newTodo]);
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
       <form action="/api/users" method="POST">
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <label>
+            Title:&nbsp;
+            <input
+              type="text"
+              data-cy="titleInput"
+              onChange={handleTitle}
+              placeholder="Enter a title"
+            />
+          </label>
+          {emptyTitle
+            && (<span className="error">Please enter a title</span>)}
         </div>
 
         <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>Choose a user</option>
-          </select>
-
-          <span className="error">Please choose a user</span>
+          <label>
+            User:&nbsp;
+            <select
+              data-cy="userSelect"
+              onChange={handleSelect}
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,127 @@
...
+import { User } from './types/User';
+import { TodoList } from './components/TodoList';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+const todos = todosFromServer.map(todo => ({
```
VitaliyBondarenko1982, line 16:
```suggestion
const initialTodos = todosFromServer.map(todo => ({
```
consider more meaningful naming

src/App.tsx
```diff
@@ -1,61 +1,127 @@
...
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [initTodos, setTodos] = useState(todos);
```
VitaliyBondarenko1982, line 22:
```suggestion
  const [todos, setTodos] = useState(todos);
```
and here todos. general convention for naming in useState. todos, setTodos ( for example for `initTodos` setter should have name `setInitTodos`)

src/App.tsx
```diff
@@ -1,61 +1,127 @@
...
+  const [title, setTitle] = useState('');
+  const [emptyTitle, setEmptyTitle] = useState(false);
+  const [userId, setUserId] = useState(0);
+  const [emptyUser, setEmptyUser] = useState(false);
```
VitaliyBondarenko1982, line 26:
```suggestion
  const [isValidTitle, setIsValidTitle] = useState(false);
  const [userId, setUserId] = useState(0);
  const [isValidUser, setIsValidUser] = useState(false);
```
consider more meaningful naming.
general convenion in JS to name boolean variables with prefix `is` (sometime in some cases `has`)

src/App.tsx
```diff
@@ -1,61 +1,127 @@
...
+
+    if (!title.trim()) {
+      setEmptyTitle(true);
+    }
+
+    if (!userId) {
+      setEmptyUser(true);
+    }
+
+    if (!title.trim() || userId === 0) {
```
VitaliyBondarenko1982, line 51:
```suggestion
    if (!title.trim() || !userId) {
```

src/App.tsx
```diff
@@ -1,61 +1,127 @@
...
+      user: getUser(userId),
+    };
+
+    const form = document.querySelector('form');
+
+    if (!form) {
+      return false;
+    }
+
+    form.reset();
```
VitaliyBondarenko1982, line 71:
```suggestion

```
no need to work with DOM tools and methods
we work with React)
just update state to initial values

src/App.tsx
```diff
@@ -1,61 +1,127 @@
...
+
+    return setTodos([...initTodos, newTodo]);
```
VitaliyBondarenko1982, line 73:
```suggestion
    setTodos([...initTodos, newTodo]);
```
redundant

src/App.tsx
```diff
@@ -1,61 +1,127 @@
...
+          <label>
+            User:&nbsp;
+            <select
+              data-cy="userSelect"
+              onChange={handleSelect}
+              defaultValue="0"
+            >
+              <option value="0" disabled>Choose a user</option>
+              {usersFromServer.map(user => (
+                <option value={`${user.id}`} key={`${user.id}`}>{user.name}</option>
```
VitaliyBondarenko1982, line 105:
```suggestion
                <option value={user.id} key={user.id}>{user.name}</option>
```

VitaliyBondarenko1982 (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+}));
 
 export const App = () => {
+  const [todos, setTodos] = useState(initialTodos);
+  const [title, setTitle] = useState('');
+  const [isValidTitle, setIsValidTitle] = useState(false);
+  const [userId, setUserId] = useState(0);
+  const [isValidUser, setIsValidUser] = useState(false);
+
+  let count = 15;
```
Volodymyr-Melnyk-Fron10der, line 28: 1. count of what?
2. using let inside compoent is bad practice - > move it to a state or use const and move it out of component

Volodymyr-Melnyk-Fron10der: I am not sure you really need it at all

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+      setIsValidTitle(true);
+    }
+
+    if (!userId) {
+      setIsValidUser(true);
+    }
+
+    if (!title.trim() || !userId) {
+      return;
+    }
```
Volodymyr-Melnyk-Fron10der, line 53:
```suggestion
    if (!title.trim()) {
      return setIsValidTitle(true);
    }

    if (!userId) {
      return setIsValidUser(true);
    }
```

Volodymyr-Melnyk-Fron10der (changes requested):

src/App.tsx
VladGraund: it's don't work great. If you don't change user and title, you have to see two errors: "Please enter a title" and "Please choose a user". Your code don't complete this task.

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,25 @@
...
+  todo: Todo
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    id, completed, title, user,
+  } = todo;
+
+  return (
+    <article data-id={`${id}`} className={`TodoInfo ${completed ? 'TodoInfo--completed' : ''}`}>
```
etojeDenys, line 15: Consider using the 'classnames' library to conditionally apply classes to the element. It makes the code cleaner and more readable. Example: import cx from 'classnames'; ... <article className={cx('TodoInfo', {'TodoInfo--completed': completed})}>

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,15 @@
-export const UserInfo = () => {};
+import { User } from '../../types/User';
+
+type Props = {
+  user: User;
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
```
etojeDenys, line 7: It's a good practice to explicitly import React when using JSX. Add `import React from 'react';` at the beginning of the file. 📦

etojeDenys (changes requested):

PR #1615, opened 2023-04-18
solution

MSA1991:
[DEMO LINK](https://msa1991.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,138 @@
+import { useState } from 'react';
+
+import { TodoList } from './components/TodoList';
+
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+const getUser = (userId: number): User | null => {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+};
+
+const newTodos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [todos, setTodos] = useState(newTodos);
+  const [user, setUser] = useState('0');
+  const [title, setTitle] = useState('');
+
+  const [invalidUser, setInvalidUser] = useState(false);
+  const [invalidTitle, setInvalidTitle] = useState(false);
+
+  const handleChangeUser = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    const { value } = event.target;
+
+    setUser(value);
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,61 +1,147 @@
...
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [todos, setTodos] = useState(newTodos);
+  const [user, setUser] = useState('0');
+  const [title, setTitle] = useState('');
+
+  const [invalidUser, setInvalidUser] = useState(false);
+  const [invalidTitle, setInvalidTitle] = useState(false);
```
DarkMistyRoom, line 30:
```suggestion
  const [isValidUser, setIsValidUser] = useState(true);
  const [isValidTitle, setIsValidTitle] = useState(true);
```

DarkMistyRoom: https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig

src/App.tsx
```diff
@@ -1,61 +1,147 @@
...
+
+  const resetForm = () => {
+    setTitle('');
+    setUser('0');
+  };
+
+  const handleSubmitForm = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    let invalidValues = false;
```
DarkMistyRoom, line 54: why do you need this additional variable? It is enough to check if `isValidUser && isValidTitle` is true

src/App.tsx
```diff
@@ -1,61 +1,147 @@
...
-          </h2>
-
-          <a className="UserInfo" href="mailto:Julianne.OConner@kory.org">
-            Patricia Lebsack
-          </a>
-        </article>
-      </section>
+      <TodoList todos={todos} />
+
     </div>
```
DarkMistyRoom, line 145:
```suggestion
      <TodoList todos={todos} />
    </div>
```

DarkMistyRoom (changes requested):
Good job, just a few fixes are needed
I can add todo with only white spaces. Fix it
image

src/App.tsx
MSA1991: Я не знаю почему, но это не работает

src/App.tsx
```diff
@@ -1,61 +1,153 @@
...
+
+  const handleSubmitForm = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    let isValidValues = true;
+
+    const titleOnlyWithLetters = title.replace(/[0-9]/g, '')
+      .split(' ')
+      .filter(letter => letter !== '')
+      .join(' ');
```
IvanFesenko, line 61: You can use [string.trim()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim)

IvanFesenko (approved):

PR #1618, opened 2023-04-20
Solution react_add-todo-form

NazariiAlieksieiev:
[DEMO LINK](https://NazariiAlieksieiev.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -78,6 +78,16 @@ export const App: React.FC = () => {
     }
   };
 
+  const titleHandler = (event: React.FormEvent<HTMLInputElement>) => {
+    setTitle(event.currentTarget.value);
+    setTitleError(false);
+  };
+
+  const nameHandler = (event: React.FormEvent<HTMLSelectElement>) => {
+    setName(event.currentTarget.value);
+    setNameError(false);
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
@@ -96,7 +106,9 @@ export const App: React.FC = () => {
             data-cy="titleInput"
             value={title}
             placeholder="Enter a title"
-            onChange={(event) => setTitle(event.target.value)}
+            onChange={(event) => {
+              titleHandler(event);
+            }}
           />
           {titleError && <span className="error">Please enter a title</span>}
         </div>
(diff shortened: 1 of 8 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,20 @@
...
+import { TodoInfo } from '../TodoInfo';
+
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todos: Todo[]
+};
+
+// Add the required types and props
+export const TodoList: React.FC<Props> = ({ todos = [] }) => (
```
IvanFesenko, line 11: You should not set a default value for `todos` prop. The `todos` prop is required in the Props

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
+  const foundUser: User | undefined = users
+    .find(user => user.name === nameUser);
+
+  if (!foundUser) {
+    throw new Error(`User '${nameUser}' not found.`);
+  }
+
+  const todoMaxId = todosList.reduce((prev, curr) => {
+    return prev.id > curr.id ? prev : curr;
+  });
```
IvanFesenko, line 37: Better use Math.max
```suggestion
  const todoMaxId = Math.max(...todosList.map(({id}) => id));
```

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
+        action="/api/users"
+        method="POST"
+        onSubmit={(event) => {
+          handleAddTodo(event);
+        }}
```
IvanFesenko, line 100: Add separate handler instead using anonymous function. Fix all similar places

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
         </div>
 
         <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>Choose a user</option>
+          <select
+            data-cy="userSelect"
+            name="name"
+            value={name}
+            onChange={(event) => nameHandler(event)}
```
IvanFesenko, line 121: No need additional callback
```suggestion
            onChange={nameHandler}
```

IvanFesenko (changes requested):
Check my comments

IvanFesenko (approved):

PR #1619, opened 2023-04-21
todo list ready for use

PavloReutskiy:
[DEMO LINK](https://PavloReutskiy.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,134 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
 
-export const App = () => {
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [userId, setUserId] = useState('0');
+  const [title, setTitle] = useState('');
+  const [todoList, setTodoList] = useState(todos);
+  const [isTitleEmpty, setIsTitleEmpty] = useState(false);
+  const [isUserEmpty, setIsUserEmpty] = useState(false);
+  const maxId = todoList.reduce((max, todo) => Math.max(max, todo.id), 0);
+
+  const handleUserSelect = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setUserId(event.target.value);
+    if (isUserEmpty) {
+      setIsUserEmpty(false);
+    }
+  };
+
+  const handleTodoTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const inputText = event.target.value;
+    const regex = /^[a-zA-Zа-яА-Я0-9\s]+$/;
+
+    if (regex.test(inputText) || inputText === '') {
+      setTitle(event.target.value);
+      if (isTitleEmpty) {
+        setIsTitleEmpty(false);
+      }
+    }
+  };
+
+  const handleTodoAddition = () => {
+    setIsTitleEmpty(title === '');
+    setIsUserEmpty(userId === '0');
+
+    if (userId !== '0' && title !== '') {
+      setTodoList([
+        ...todoList,
+        {
+          id: maxId + 1,
+          userId: +userId,
+          title,
+          completed: false,
+          user: getUser(+userId),
+        },
+      ]);
+
+      setUserId('0');
+      setTitle('');
+    }
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
+        action="/api/users"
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,135 @@
...
+
+    if (userId !== '0' && title.trim() !== '') {
+      setTodoList([
+        ...todoList,
+        {
+          id: maxId + 1,
+          userId: +userId,
+          title,
+          completed: false,
+          user: getUser(+userId),
```
jstmpelowycz, line 63:
```suggestion
          user: getUser(Number(userId)),
```

src/App.tsx
```diff
@@ -1,61 +1,135 @@
...
+  return foundUser || null;
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [userId, setUserId] = useState('0');
```
jstmpelowycz, line 22: saving user id as a string seems not to be reliable, try saving it as a number

src/App.tsx
```diff
@@ -1,61 +1,135 @@
...
+  const [title, setTitle] = useState('');
+  const [todoList, setTodoList] = useState(todos);
+  const [isTitleEmpty, setIsTitleEmpty] = useState(false);
+  const [isUserEmpty, setIsUserEmpty] = useState(false);
+
+  const maxId = todoList.reduce((max, todo) => Math.max(max, todo.id), 0);
+
+  const handleUserSelect = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setUserId(event.target.value);
+    if (isUserEmpty) {
```
jstmpelowycz, line 32:
```suggestion
    setUserId(event.target.value);

    if (isUserEmpty) {
```

src/App.tsx
```diff
@@ -1,61 +1,135 @@
...
+      setIsUserEmpty(false);
```
jstmpelowycz, line 33: I believe, it's better to perform this validation inside submit callback

src/App.tsx
```diff
@@ -1,61 +1,135 @@
...
+
+  const handleTodoTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const inputText = event.target.value;
+    const regex = /^[a-zA-Zа-яА-Я0-9\s]+$/;
+
+    if (regex.test(inputText) || inputText === '') {
+      setTitle(event.target.value);
+      if (isTitleEmpty) {
+        setIsTitleEmpty(false);
+      }
```
jstmpelowycz, line 45: > I believe, it's better to perform this validation inside submit callback

jstmpelowycz (changes requested):

src/App.tsx
PavloReutskiy: I tried, but I don't know how. If I move it to form handler it doesn't work like on screen(( it has to change immediately so I put it in the handler for input
image

PavloReutskiy:
all done except one because don't know how((

src/App.tsx
```diff
@@ -1,61 +1,145 @@
...
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+
+function getUser(userId: number): User | null {
```
SerhiiKirik, line 10:
```suggestion
function getUserById(id: number): User | null {
```

src/App.tsx
```diff
@@ -1,61 +1,145 @@
...
+    if (isTitleEmpty) {
+      setIsTitleEmpty(false);
+    }
+  };
+
+  const handleTodoAddition = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    setIsTitleEmpty(title.trim() === '');
+    setIsUserEmpty(userId === 0);
```
SerhiiKirik, line 55:
```suggestion
    if (!title.trim()) {
      setIsTitleEmpty(true);
    }
    
    const user = getUserById(userId);
    
    if (!user) {
      setIsUserEmpty(true);
    }
    
    if (!title.trim() || !user) {
      return;
    }
```

src/App.tsx
```diff
@@ -1,61 +1,145 @@
...
+
+    if (userId !== 0 && title.trim() !== '') {
```
SerhiiKirik, line 57:
```suggestion
```

src/App.tsx
```diff
@@ -1,61 +1,145 @@
...
+      setTodoList([
+        ...todoList,
+        {
+          id: maxId + 1,
+          userId: +userId,
+          title,
+          completed: false,
+          user: getUser(userId),
```
SerhiiKirik, line 65:
```suggestion
          user,
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,34 @@
-export const TodoInfo = () => {};
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+// import { User } from '../../types/User';
```
SerhiiKirik, line 3: remove all comments

SerhiiKirik (changes requested):
Great 🔥 
Let's make some improvements 🧑‍💻 
I can add todo just with spaces
image

PavloReutskiy:
all done!

PR #1620, opened 2023-04-21
solution

ladrom:
[demo](https://ladrom.github.io/react_add-todo-form/)


src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,23 @@
...
+  users: User[],
+};
+
+export const TodoList: React.FC<Props> = ({ todos, users }) => {
+  return (
+    <section className="TodoList">
+      {todos.map(todo => (
+        <TodoInfo
+          todo={todo}
+          users={users}
```
Anastasiia-Svintsova, line 16: Why do you need to pass all users to the TodoInfo? You should find the user before

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,36 @@
...
+        )
+      }
+    >
+      <h2 className="TodoInfo__title">
+        {title}
+      </h2>
+
+      <a className="UserInfo" href={users[userId - 1].email}>
+        {users[userId - 1].name}
+      </a>
```
Anastasiia-Svintsova, line 33: So refactor it as well according to my previous comment

Anastasiia-Svintsova (changes requested):

ivanchuk-yurii (changes requested):

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,27 @@
...
+type Props = {
+  todos: Todo[],
+  users: User[],
+};
+
+export const TodoList: React.FC<Props> = ({ todos, users }) => {
+  return (
+    <section className="TodoList">
+      {todos.map(todo => {
+        const targetUser = users[todo.userId - 1];
```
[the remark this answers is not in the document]
ladrom: It doesn't work. Gives an error message 'No overload matches this call'.

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,25 @@
...
+  todos: Todo[],
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="TodoList">
+      {todos.map(todo => {
+        const { user, id } = todo;
+
+        return user && (
```
agniya-i, line 14:
```suggestion
        return (
```
this check could be redundant

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,39 @@
-export const TodoInfo = () => {};
+import classNames from 'classnames';
+import { Todo } from '../../types/todo';
+import { User } from '../../types/user';
+
+type Props = {
+  todo: Todo,
+  user: User,
```
agniya-i, line 7: you don't use user in props, so you can delete it

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,25 @@
...
+          <TodoInfo
+            todo={todo}
+            user={user}
```
agniya-i, line 17:
```suggestion
     
```

agniya-i (changes requested):

nazarmatsevych (approved):

PR #1622, opened 2023-04-22
Develop

zynkd:
[DEMO LINK](https://zynkd.github.io/react_add-todo-form/)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,25 @@
-export const TodoInfo = () => {};
+import React from 'react';
+import getClassName from 'classnames';
```
witflash, line 2: It's not critical, but the general naming for this lib is "cx" or just "classNames".
So you can do you import like this:
```
import cx from "classNames"
```

witflash (approved):
Great! I like it!
Well done 👍🏻

PR #1637, opened 2023-04-25
solution without all passed tests

VenomMB:
[DEMO LINK](https://VenomMB.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,100 @@
+import { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { Todo } from './types/interface';
+import { TodoList } from './components/TodoList';
+
+function getUserByName(nameUser: string) {
+  return usersFromServer.find(({ name }) => name === nameUser) || null;
+}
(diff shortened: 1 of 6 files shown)


Nazarii98 (changes requested):
Great job!
Go try to fix one bug. 
I don't think this is a useful feature.

image


You should make your default value for the list unselectable

SerhiiKirik (changes requested):
Well done 🔥 
Let's fix some bugs 🧑‍💻 
I can't add todo just with spaces - it's good, but i don't see some error
image

src/App.tsx
```diff
@@ -1,61 +1,136 @@
...
+              data-cy="userSelect"
+              value={userSelect}
+              onChange={handleChangeSelect}
+            >
+              <option value="" disabled>Choose a user</option>
+              {usersFromServer.map(({ name, id }) => (
+                <option value={name} key={id}>
+                  {name}
+                </option>
+              ))}
```
SerhiiKirik, line 117:
```suggestion
              <option value="" disabled>Choose a user</option>
              
              {usersFromServer.map(({ name, id }) => (
                <option value={name} key={id}>
                  {name}
                </option>
              ))}
```

SerhiiKirik (approved):
LGTM 🔥

PR #1640, opened 2023-04-26
Solution

voronevska:
[DEMO LINK](https://voronevska.github.io/react_add-todo-form/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,4 +27,4 @@ Implement the ability to add TODOs to the `TodoList` implemented in the [Static 
 - Implement a solution following the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline).
 - Use the [React TypeScript cheat sheet](https://mate-academy.github.io/fe-program/js/extra/react-typescript).
 - Open one more terminal and run tests with `npm test` to ensure your solution is correct.
-- Replace `<your_account>` with your Github username in the [DEMO LINK](https://MylaCrystal.github.io/react_add-todo-form/) and add it to the PR description.
+- Replace `<your_account>` with your Github username in the [DEMO LINK](https://voronevska.github.io/react_add-todo-form/) and add it to the PR description.


src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+      setIsValidTitle(Boolean(title.length));
+      setIsValidUser(Boolean(userId));
+    }
+  };
+
+  const handleChangeTitle = (
+    { target: { value } }: ChangeEvent<HTMLInputElement>,
+  ) => {
+    setTitle(value);
+    setIsValidTitle(Boolean(value) || !userId);
```
roshirin, line 56: If you choose a user, type something into the title input and then erase it, you'll get the error message shown.

I think it's better to just pass true as an argument here and in the next case.

roshirin (approved):
Works nice, check some suggestions in the comments.

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,27 @@
...
+  } = todo;
+
+  return (
+    <article
+      data-id={id}
+      className={classNames('TodoInfo', { 'TodoInfo--completed': completed })}
+    >
+      <h2 className="TodoInfo__title">{title}</h2>
+
+      {user && <UserInfo user={user} />}
```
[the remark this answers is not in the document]
MylaCrystal: in general, agree with this, but we have a check that does not allow posts without user input (here is an error message for this)

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/Todo';
+
+const getTodoId = (todos: Omit<Todo, 'user'>[]) => {
+  return todos.reduce((maxId, todo) => Math.max(maxId, todo.id), 0);
+};
+
+const preparedTodo = (todos: Omit<Todo, 'user'>[]) => {
```
SerhiiKirik, line 13:
```suggestion
const getPreparedTodo = (todos: Omit<Todo, 'user'>[]) => {
```

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+          userId,
+        };
+
+        return [...prevTodos, newTodo];
+      });
+      setTitle('');
+      setUserId(0);
+    } else {
+      setIsValidTitle(Boolean(title.length));
+      setIsValidUser(Boolean(userId));
```
SerhiiKirik, line 48:
```suggestion
      if (!title.length) {
        setIsValidTitle(false);
      }
      
    ......
```

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+    }
+  };
+
+  const handleChangeTitle = (
+    { target: { value } }: ChangeEvent<HTMLInputElement>,
+  ) => {
```
SerhiiKirik, line 54: just add type to the event
```suggestion
  const handleChangeTitle = (event: ...) => {
```

src/App.tsx
SerhiiKirik, line 56: just remove error if you have error

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+    setTitle(value);
+    setIsValidTitle(Boolean(value) || !userId);
+  };
+
+  const handleChangeUserId = (
+    { target: { value } }: ChangeEvent<HTMLSelectElement>,
+  ) => {
+    setUserId(+value);
+    setIsValidUser(Boolean(value) || !title);
```
SerhiiKirik, line 63: the same

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
-          <h2 className="TodoInfo__title">
-            quis ut nam facilis et officia qui
-          </h2>
-
-          <a className="UserInfo" href="mailto:Julianne.OConner@kory.org">
-            Patricia Lebsack
-          </a>
-        </article>
-      </section>
+      <TodoList todos={preparedTodo(todos)} />
```
SerhiiKirik, line 128: move to the variable
```suggestion
      <TodoList todos={...} />
```

SerhiiKirik (changes requested):
Well done 🔥 
Let's fix some bugs 👩‍💻 
I can add todo just with spaces
image

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { ChangeEvent, FormEvent, useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/Todo';
+
+const getTodoId = (todos: Omit<Todo, 'user'>[]) => {
```
SerhiiKirik, line 9:
```suggestion
const getTodoById = (todos: Omit<Todo, 'user'>[]) => {
```

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+          title,
+          completed: false,
+          userId,
+        };
+
+        return [...prevTodos, newTodo];
+      });
+      setTitle('');
+      setUserId(0);
+    } else {
```
SerhiiKirik, line 48: try to avoid `if / else` construction

SerhiiKirik (approved):
Good job 🔥

PR #1646, opened 2023-04-26
react todo-form

ofenchuk:
[DEMO LINK](https://ofenchuk.github.io/react_add-todo-form/)

one test dont pass for some reason, its when need to add id + 1 to title, but it works, it can be seen in state Id

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import { FC, useState } from 'react';
+import { FC, FormEvent, useState } from 'react';
 
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,61 +1,127 @@
...
+    }
+  }
+
+  function handleSubmit(event: FormEvent<HTMLFormElement>) {
+    event.preventDefault();
+
+    const target = event.target as typeof event.target & {
+      title: { value: string };
+      user: { value: string };
+    };
```
Kavon-Skela, line 55: Using aliases is a bad practice. Try to avoid it

Kavon-Skela (changes requested):
You didn't reach the target, but you are so close, good job! You can try to fix your test with my comments. Other comments related to code style and avoiding overload of your code

PR #1675, opened 2023-04-28
solution

AnastassiaH:
[DEMO LINK](https:/anastassiah.github.io/react_add-todo-form/)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,13 @@
-export const UserInfo = () => {};
+import { User } from '../../types/User';
+
+type Props = {
+  user: User | undefined,
+};
+
+export const UserInfo:React.FC<Props> = ({ user }) => {
```
Volodymyr-Melnyk-Fron10der, line 7: Add a space between 'React.FC' and '<Props>' for better readability.

src/App.tsx
```diff
@@ -1,61 +1,106 @@
...
-          <span className="error">Please enter a title</span>
+          <label htmlFor="titleInput">Title: </label>
+          <input
+            id="titleInput"
+            type="text"
+            data-cy="titleInput"
+            placeholder="Enter a title"
+            value={title}
+            onChange={(e) => {
+              setTitle(e.target.value.replace(/[^A-Za-z0-9 ]/, ''));
```
Volodymyr-Melnyk-Fron10der, line 68: Instead of using replace() with a regular expression to filter out unwanted characters, consider using an onChange handler that only sets the state if the input value matches the desired pattern.

src/App.tsx
```diff
@@ -1,61 +1,106 @@
...
+          completed: false,
+          userId: users?.find(user => user.name === userName)?.id || -1,
+        },
+      ];
+    });
+  };
+
+  const validateForm = () => {
+    if (!title.trim()) {
+      setError({ ...error, title: true });
```
Volodymyr-Melnyk-Fron10der, line 30: use callback if you need the current state

src/App.tsx
```diff
@@ -1,61 +1,106 @@
...
-          <select data-cy="userSelect">
-            <option value="0" disabled>Choose a user</option>
+          <label htmlFor="userSelect">User: </label>
+          <select
+            id="userSelect"
+            data-cy="userSelect"
+            value={userName}
+            onChange={(e) => {
+              setUserName(e.target.value);
+              setError({ ...error, user: false });
```
Volodymyr-Melnyk-Fron10der, line 82: 2+ lines -> move into a separated function

src/App.tsx
```diff
@@ -1,61 +1,106 @@
...
-
-          <span className="error">Please choose a user</span>
+          {error.user && <span className="error">Please choose a user</span>}
         </div>
 
-        <button type="submit" data-cy="submitButton">
+        <button
+          type="submit"
+          data-cy="submitButton"
+          onClick={(e) => handleAdd(e)}
```
Volodymyr-Melnyk-Fron10der, line 96:
```suggestion
          onClick={ handleAdd}
```

Volodymyr-Melnyk-Fron10der (changes requested):
1. There is no error for the title, it should be filled in as well
image

2. option should be disabled according to the design
image

nazarmatsevych (approved):

PR #1677, opened 2023-04-30
Develop

NedobezhkinOleksandr:
https://NedobezhkinOleksandr.github.io/react_add-todo-form/


src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+  user: getUser(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [todo, setTodo] = useState(todos);
+  const [title, setTitle] = useState('');
+  const [user, setUser] = useState(0);
+
+  const [userError, setUserError] = useState(false);
+  const [titleError, setTitleError] = useState(false);
```
tetiana-ivanchenko, line 28: follow the convention for the boolean variables name

tetiana-ivanchenko (changes requested):
fix demo link

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+
+  const addTodo = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+    const titleTrim = title.trim();
+
+    setTitleErrored(!titleTrim);
+
+    setUserErrored(!user);
+
+    if (titleTrim && user) {
```
Anastasiia-Svintsova, line 48: You could just add return to an opposite condition to avoid nesting
```suggestion
    if (!titleTrim || !user) {
      return
    }
    
    const newTodo = { ...
```

src/components/TodoInfo/index.ts
```diff
@@ -1 +1,2 @@
+// eslint-disable-next-linem
```
Anastasiia-Svintsova, line 1: Try not to disable eslint

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,18 @@
-export const TodoList = () => {};
+// eslint-disable-next-line
```
Anastasiia-Svintsova, line 1: Same. Why do you need these disablings?

Anastasiia-Svintsova (approved):

PR #1678, opened 2023-04-30
solution

gabriella135:
[DEMO LINK](https://gabriella135.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,108 @@
+import { ChangeEvent, FormEvent, useState } from 'react';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/Todo';
+import { User } from './types/User';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
 
-export const App = () => {
+  return foundUser || null;
+}
+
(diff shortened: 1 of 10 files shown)


polosanya (changes requested):
Almost done!
The error should disappear after user is chosen
image
image

src/App.tsx
```diff
@@ -1,61 +1,116 @@
...
+
+export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>(updatedTodos);
+  const [title, setTitle] = useState('');
+  const [userName, setUserName] = useState(0);
+
+  const [isTitleError, setTitleError] = useState(false);
+  const [isUserError, setUserError] = useState(false);
+
+  const newTodoId = Math.max(...todos.map(todo => todo.id));
```
polosanya, line 28:
```suggestion
  const newTodoId = Math.max(...todos.map(todo => todo.id)) + 1;
```
Better save the final version of new id

src/App.tsx
```diff
@@ -1,61 +1,116 @@
...
+  user: getUser(todo.userId) || null,
+}));
+
+export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>(updatedTodos);
+  const [title, setTitle] = useState('');
+  const [userName, setUserName] = useState(0);
+
+  const [isTitleError, setTitleError] = useState(false);
+  const [isUserError, setUserError] = useState(false);
```
polosanya, line 26:
```suggestion
  const [isTitleError, setIsTitleError] = useState(false);
  const [isUserError, setIsUserError] = useState(false);
```
setState function should be named according to the noun variable

polosanya (approved):
Good job!

PR #1681, opened 2023-05-06
react add todo form

Oleksandr-Maslo:
https://oleksandr-maslo.github.io/react_add-todo-form/


src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+import { Todo } from './types/Todo';
+
+const getUserByID = (id: number) => {
+  const foundUser = usersFromServer.find(user => user.id === id);
+
+  return foundUser || null;
+};
+
+const findMaxId = (todos: Todo[]) => {
+  let maxId = 0;
```
TarasHoliuk, line 16: Use Math.max with map to simplify the findMaxId function

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,13 @@
-export const UserInfo = () => {};
+import { User } from '../../types/User';
+
+interface Props {
+  user: User
+}
+
+export const UserInfo = ({ user }: Props) => {
```
TarasHoliuk, line 7: Consider destructuring 'user' in the function argument to improve readability.

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
+        action="/api/users"
+        method="POST"
+        onSubmit={(event) => {
+          event.preventDefault();
+
+          if (!id || !title) {
```
TarasHoliuk, line 46: you should trim the title to prevent adding smth like that `      `

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
+        action="/api/users"
+        method="POST"
+        onSubmit={(event) => {
```
TarasHoliuk, line 43: move it into a separate function instead of passing anonymous function

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+          <input
+            id="title"
+            name="title"
+            type="text"
+            data-cy="titleInput"
+            placeholder="Enter a title"
+            value={title}
+            onChange={(event) => {
+              setTitle(event.target.value);
+              setError({ ...error, title: false });
```
TarasHoliuk, line 80: move change handlers into separate functions
use functional setState if you need the current state value

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,25 @@
...
+
+interface Props {
+  todo: Todo
+}
+
+export const TodoInfo = ({ todo }: Props) => {
+  return (
+    <article
+      data-id={todo.id}
+      className={classNames(
```
TarasHoliuk, line 13: wrong usage of `classNames` - check it [here](https://www.npmjs.com/package/classnames)

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,25 @@
...
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+interface Props {
+  todo: Todo
+}
+
+export const TodoInfo = ({ todo }: Props) => {
+  return (
```
TarasHoliuk, line 10: you may do the destructuring assignment of `todo` object here, before `return`

src/components/TodoInfo/index.ts
```diff
@@ -1 +1,2 @@
+// eslint-disable-next-line
```
TarasHoliuk, line 1: why do you need it (i.e. try to figure out with eslint rule, or ask about that in the chat)

TarasHoliuk (changes requested):

tetiana-ivanchenko (changes requested):
if i try to add a title that only consists of spaces, it doesn't get added, but i don't see any errors
image

OleziO (approved):

PR #1717, opened 2023-05-19
Develop

kq333:
[DEMO LINK](https://<your_account>.github.io/react_add-todo-form/)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,22 +1,14 @@
...
-  // onTodosUpdated: (todoToUpdate: Todo) => void
-};
+interface TodoListProps {
+  todos: Todos[];
+}
 
-export const TodoList: React.FC<Props> = ({ todos, onTodosDeleted }) => (
-  <section className="TodoList">
+export const TodoList: React.FC<TodoListProps> = ({ todos }) => (
+  <>
```
d-wasilewski, line 9: this should be wrapped in that section tag here

src/App.tsx
```diff
@@ -1,59 +1,115 @@
...
+    ...todo,
+    user: getUserById(todo.userId),
+  }));
+}
 
 export const App = () => {
-  const [todos, setTodos] = useState<Todo[]>([]);
+  const [sectionValue, setSectionValue] = useState('');
+  const [inputValue, setInputValue] = useState('');
+  const [CopyToDoList, setCopyToDoList] = useState<Todos[]>(InitialArray());
```
d-wasilewski, line 27: It's not a copy, you just set the initial value as the existing array
```suggestion
  const [todoList, setToDoList] = useState<Todos[]>(InitialArray());
```

src/App.tsx
```diff
@@ -1,59 +1,115 @@
...
+      id: newId + 1,
+      title: inputValue,
+      completed: false,
+      userId: 0,
+      user: getUserById(userId),
+    };
+
+    if (inputValue.length === 0) {
+      setTitleState(true);
+    }
```
d-wasilewski, line 47: shouldnt it be named as errors not state?

d-wasilewski (changes requested):
It should be fixed
image

src/App.tsx
```diff
@@ -1,59 +1,121 @@
...
+}
+
+function InitialArray(): Todo[] {
+  return todosFromServer.map(todo => ({
+    ...todo,
+    user: getUserById(todo.userId),
+  }));
+}
+
+const startingArray = InitialArray();
```
d-wasilewski, line 24:
```suggestion
const startingArray: Todo[] = todosFromServer.map(todo => ({
    ...todo,
    user: getUserById(todo.userId),
  }));

```

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #1722, opened 2023-05-21
add todo form was implement

2pasha:
[DEMO LINK](https://2pasha.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,120 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
 
-export const App = () => {
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+export const getTodos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+type HandleChangeType = React.ChangeEvent<HTMLInputElement | HTMLSelectElement>;
+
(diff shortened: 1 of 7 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,27 @@
...
+  },
+}) => (
+  <article
+    data-id={id}
+    className={classNames('TodoInfo',
+      { 'TodoInfo--completed': completed })}
+  >
+    <h2 className="TodoInfo__title">{title}</h2>
+
+    {user && <UserInfo user={user} key={user.id} />}
```
etojeDenys, line 25: Remove the 'key' prop from the 'UserInfo' component as it's not needed here. Key prop is used when rendering a list of elements.

src/App.tsx
```diff
@@ -1,61 +1,128 @@
...
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+export const getTodos: Todo[] = todosFromServer.map(todo => ({
```
etojeDenys, line 18: named as a function, but it's an array

better to rename this var

etojeDenys (changes requested):

IvanFesenko (changes requested):
There are possibility to add just spaces - fix it
2023-05-24_14h13_57

src/App.tsx
```diff
@@ -1,61 +1,128 @@
...
+
+type HandleChangeType = React.ChangeEvent<HTMLInputElement | HTMLSelectElement>;
+
+export function correctTitle(str: string) {
+  return str.trim().replace(/[^a-zA-Zа-яА-Я0-9\s]/g, '');
+}
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [userSelect, setUserSelect] = useState(0);
```
DarkMistyRoom, line 31:
```suggestion
  const [selectedUserId, setSelectedUserId] = useState(0);
```

PR #1726, opened 2023-05-22
To-do list form (react app) - solution

ctgdprzyb:
[DEMO LINK](https://ctgdprzyb.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,3 +1,4 @@
+import { FormEventHandler, useState } from 'react';
 import './App.scss';
 import { TodoList } from './components/TodoList';
 
@@ -9,21 +10,40 @@ import { User } from './types/User';
 function getUser(userId: number): User | null {
   const foundUser = usersFromServer.find(user => user.id === userId);
 
-  // if there is no user with a given userId
   return foundUser || null;
 }
 
-export const todos: Todo[] = todosFromServer.map(todo => ({
-  ...todo,
-  user: getUser(todo.userId),
-}));
-
 export const App = () => {
+  const [newTaskTitle, setNewTaskTitle] = useState('asdasd');
+  const [newTaskUserId, setNewTaskUserId] = useState(1);
+  // const [todoList, setTodoList] = useState<Todo[]>(todos);
+
+  const todos: Todo[] = todosFromServer.map(todo => ({
+    ...todo,
+    user: getUser(todo.userId),
+  }));
+
+  const addNewTask: FormEventHandler = (event) => {
+    event.preventDefault();
+    todosFromServer.push({
+      title: newTaskTitle,
+      completed: false,
+      userId: newTaskUserId,
+      id: Math.random(),
+    });
+    setNewTaskTitle('');
+    setNewTaskUserId(0);
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
+        action="/api/users"
+        method="POST"
+        onSubmit={addNewTask}
+      >
         <div className="field">
           <label htmlFor="titleInput">Title: </label>
           <input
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,118 @@
...
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
 
 export const App = () => {
+  const [newTaskTitle, setNewTaskTitle] = useState('');
+  const [newTaskUserId, setNewTaskUserId] = useState(0);
+  const [titleError, setTitleError] = useState(false);
```
d-wasilewski, line 19: Name of the boolean variable has to start with `is` or `has`

src/App.tsx
```diff
@@ -1,61 +1,118 @@
...
+    user: getUser(todo.userId),
+  }));
+
+  const addNewTask: FormEventHandler = (event) => {
+    event.preventDefault();
+    const emptyTaskTitle = (newTaskTitle === '');
+    const invalidUserId = (newTaskUserId === 0);
+
+    setTitleError(emptyTaskTitle);
+    setUserError(invalidUserId);
```
d-wasilewski, line 33: maybe set this only if the condition is meet instead of setting it everytime

src/App.tsx
```diff
@@ -1,61 +1,118 @@
...
+
+    if (!emptyTaskTitle && !invalidUserId) {
+      todosFromServer.push({
+        title: newTaskTitle,
+        completed: false,
+        userId: newTaskUserId,
+        id: Math.random(),
```
d-wasilewski, line 40: Task description:
> id is the largest id in the array + 1

src/App.tsx
```diff
@@ -1,61 +1,118 @@
...
+            value={newTaskTitle}
+            onChange={(event) => {
+              setTitleError(false);
+              validateTitleField(event);
+            }}
+          />
+
+          {titleError
+            ? <span className="error">Please enter a title</span>
+            : undefined}
```
d-wasilewski, line 79: use && instead of returning undefined

src/App.tsx
```diff
@@ -1,61 +1,118 @@
...
+          <label htmlFor="userSelect">User: </label>
+          <select
+            name="userSelect"
+            data-cy="userSelect"
+            defaultValue={0}
+            value={newTaskUserId}
+            onChange={(event) => {
+              setUserError(false);
+              setNewTaskUserId(Number(event.target.value));
+            }}
```
d-wasilewski, line 92: you can create a handler for this

d-wasilewski (changes requested):

src/App.tsx
ctgdprzyb: removed entirely

ctgdprzyb: unremoved because i was stupid

src/App.tsx
```diff
@@ -1,61 +1,115 @@
...
+          />
+
+          {hasEmptyTitle && <span className="error">Please enter a title</span>}
         </div>
 
         <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>Choose a user</option>
+          <label htmlFor="userSelect">User: </label>
+          <select
```
d-wasilewski, line 83: Add `id` attribute to the select element for better accessibility. It should match the `htmlFor` attribute in the label

src/App.tsx
```diff
@@ -1,61 +1,115 @@
...
+  }));
+
+  const addNewTask: FormEventHandler = (event) => {
+    event.preventDefault();
+
+    setHasEmptyTitle(!newTaskTitle);
+    setHasEmptyUser(!newTaskUserId);
+
+    if (newTaskTitle && newTaskUserId) {
+      highestId += 1;
```
d-wasilewski, line 36: Do not mutate the external `todosFromServer` array. Use the local `todos` array instead, and change it using state management

src/App.tsx
```diff
@@ -1,61 +1,115 @@
...
+import { Todo } from './types/Todo';
+import { User } from './types/User';
+
+function getUserById(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+let highestId = Math.max.apply(null, todosFromServer.map(todo => todo.id));
```
d-wasilewski, line 16: Avoid using `Math.max.apply(null, ...)` to find the highest id. Use `Math.max(...todosFromServer.map(todo => todo.id))` instead

src/App.tsx
```diff
@@ -1,61 +1,115 @@
...
+        userId: newTaskUserId,
+        id: highestId,
+      });
+      setNewTaskTitle('');
+      setNewTaskUserId(0);
+    }
+  };
+
+  const validateTitleField = (event: ChangeEvent<HTMLInputElement>) => {
+    const validatedTitle
```
d-wasilewski, line 49: Use a more descriptive function name for `validateTitleField` such as `handleTitleChange`.

src/App.tsx
```diff
@@ -1,61 +1,115 @@
...
+            value={newTaskUserId}
+            onChange={(event) => {
+              setHasEmptyUser(false);
+              setNewTaskUserId(Number(event.target.value));
+            }}
+          >
+            <option value={0} disabled>Choose a user</option>
+            {usersFromServer.map((user: User) => {
+              return (
+                <option value={user.id}>{user.name}</option>
```
d-wasilewski, line 96: Add a `key` attribute to the option elements

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #1729, opened 2023-05-26
solution

AnastasiiaVorobets:
[DEMO LINK](https://AnastasiiaVorobets.github.io/react_add-todo-form/)


nazarmatsevych (approved):
Well done!
image

PR #1742, opened 2023-06-06
done

generatedman:
[DEMO LINK](https://generatedman.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,130 @@
 import './App.scss';
+import { useState } from 'react';
+import { TodoList } from './components/TodoList';
+import { User } from './components/UserInfo';
+import { Todo } from './components/TodoInfo';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  // if there is no user with a given userId
+  return foundUser || null;
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [titleError, setTitleError] = useState(false);
+  const [pickUserError, setPickUserError] = useState(false);
+  const [pickedUser, pickUser] = useState(0);
+  const [actualTodos, setTodos] = useState(todos);
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    const newId = Math.max(...todos.map(todo => todo.id)) + 1;
+
+    const addedTodo: Todo = {
+      id: newId,
+      completed: false,
+      user: getUser(Number(pickedUser)),
+      title: title.trim(),
+      userId: Number(pickedUser),
+    };
+
+    if (!title) {
+      setTitleError(true);
+    }
+
+    if (!pickedUser) {
+      setPickUserError(true);
+    }
+
+    if (!pickedUser || !title) {
+      return;
+    }
+
+    setTodos([...actualTodos, addedTodo]);
+    setTitle('');
+    pickUser(0);
+  };
+
+  const handleChange = (
+    event: React.ChangeEvent<HTMLInputElement | HTMLSelectElement>,
+  ) => {
+    const { name, value } = event.target;
+
+    if (name === 'title') {
+      setTitle(value);
+      setTitleError(false);
+    } else {
+      pickUser(Number(value));
+      setPickUserError(false);
+    }
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
+        action="/api/users"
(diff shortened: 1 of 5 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,35 @@
...
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    id,
+    completed,
+    title,
+    user,
+  } = todo;
+
+  return (
+    <article data-id={id} className={`TodoInfo ${completed ? 'TodoInfo--completed' : ''}`}>
```
silchyn, line 25: use `cn`

silchyn (changes requested):
I can submit the form with a title consisting of only spaces

silchyn (changes requested):
> I can submit the form with a title consisting of only spaces

still

src/App.tsx
```diff
@@ -1,61 +1,130 @@
...
+    event.preventDefault();
+
+    const newId = Math.max(...todos.map(todo => todo.id)) + 1;
+
+    const addedTodo: Todo = {
+      id: newId,
+      completed: false,
+      user: getUser(Number(user)),
+      title: title.trim(),
+      userId: Number(user),
```
dffUqp, line 39: I think `user` is already a number, or am I wrong?🤔

src/App.tsx
```diff
@@ -1,61 +1,132 @@
...
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+function getUserById(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  // if there is no user with a given userId
```
33Orange, line 13: remove redundant comments

src/App.tsx
```diff
@@ -1,61 +1,132 @@
...
+    const validTitle = title.trim();
+
+    const newId = Math.max(...todos.map(todo => todo.id)) + 1;
+
+    const addedTodo: Todo = {
+      id: newId,
+      completed: false,
+      user: getUserById(Number(user)),
+      title: validTitle,
+      userId: Number(user),
```
33Orange, line 41: Let's create new todo after all our checks. 
We don't need to create such a massive data object when there is a chance that we will not pass the test.

src/App.tsx
```diff
@@ -1,61 +1,132 @@
...
+    const { name, value } = event.target;
+
+    if (name === 'title') {
+      setTitle(value);
+      setIsTitleValid(false);
+    } else {
+      setUser(Number(value));
+      setIsUserValid(false);
+    }
+  };
```
33Orange, line 73: **Let's create 2 separate handlers for input and select**

src/App.tsx
```diff
@@ -1,61 +1,132 @@
...
+  const [isUserValid, setIsUserValid] = useState(false);
+  const [user, setUser] = useState(0);
+  const [actualTodos, setTodos] = useState(todos);
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    const validTitle = title.trim();
+
+    const newId = Math.max(...todos.map(todo => todo.id)) + 1;
```
33Orange, line 34: Let's create helper for get new todo id and move it to separate file `helpers`

src/App.tsx
```diff
@@ -1,61 +1,132 @@
...
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [isTitleValid, setIsTitleValid] = useState(false);
+  const [isUserValid, setIsUserValid] = useState(false);
+  const [user, setUser] = useState(0);
```
33Orange, line 26: Is it `user` ? Or just user id ?

33Orange (changes requested):
Almost done

src/App.tsx
```diff
@@ -1,61 +1,132 @@
...
+    if (!userId || validTitle) {
+      return;
+    }
+
+    const addedTodo: Todo = {
+      id: newId,
+      completed: false,
+      user: getUserById(Number(userId)),
+      title: validTitle,
+      userId: Number(userId),
```
dffUqp, line 53: in `handleSelect` function you use `Number(event.target.value)`.
why do you need it here?🤔

dffUqp (changes requested):
some fixes needed:upside_down_face:

src/App.tsx
```diff
@@ -1,61 +1,130 @@
...
+  ...todo,
+  user: getUserById(todo.userId),
+}));
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [isTitleValid, setIsTitleValid] = useState(false);
+  const [isUserIdValid, setIsUserIdValid] = useState(false);
+  const [userId, setUserId] = useState(0);
+  const [actualTodos, setListTodos] = useState(todos);
```
dffUqp, line 27: In useStare our variables should have the same name for more readability. 

I mean if the first variable is `actualTodos` then the second should be `setActualTodos`

dffUqp (approved):
Completed:fire:

PR #1745, opened 2023-06-11
Develop

haduigon:
https://haduigon.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,120 @@
 import './App.scss';
-
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import React, { useState } from 'react';
+import classNames from 'classnames';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
 export const App = () => {
+  const [newToDo, setNewToDo] = useState({
+    title: '',
+    user: '',
+  });
+  const todosCopy = [...todosFromServer];
+  const copy = [...usersFromServer];
+
+  const [toDoList, setToDoList] = useState(todosCopy);
+
+  function submitHandler(event: React.FormEvent<HTMLFormElement>) {
+    event.preventDefault();
+    const maxId = Math.max(...todosCopy.map(todo => +todo.id));
+    const eventUser = (event.currentTarget[1] as HTMLInputElement).value;
+    const eventTitle = (event.currentTarget[0] as HTMLInputElement).value;
+
+    if (eventUser === '0') {
+      setNewToDo(prevState => ({
+        title: prevState.title,
+        user: 'alarm',
+      }));
+      setTimeout(() => setNewToDo(prevState => ({
+        title: prevState.title,
+        user: '',
+      })), 1000);
+    } else if (eventTitle === '') {
+      setNewToDo(prevState => ({
+        title: 'test',
+        user: prevState.user,
+      }));
+      setTimeout(() => setNewToDo(prevState => ({
+        title: '',
+        user: prevState.user,
+      })), 1000);
+    }
+
+    const post = {
+      id: maxId + 1,
+      title: eventTitle,
+      completed: false,
+      userId: +eventUser,
+    };
+
+    if (eventTitle !== '' && eventUser !== '0') {
+      setToDoList([post, ...toDoList]);
+      setNewToDo({
+        title: '',
+        user: '',
+      });
+    }
+  }
+
+  function inputHandler(event: React.ChangeEvent<HTMLInputElement>) {
+    setNewToDo({
+      title: event.target.value.replace(/[^a-z0-9 ]/gi, ''),
+      user: newToDo.user,
+    });
+  }
+
+  function handleSelect(event: React.ChangeEvent<HTMLSelectElement>) {
+    setNewToDo({
+      title: newToDo.title,
+      user: event.target.value === '0' ? '' : event.target.value,
+    });
+  }
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form onSubmit={(event) => submitHandler(event)}>
         <div className="field">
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,25 +1,120 @@
 import './App.scss';
```
VitaliyBodnar, line 1: according to the style guide, import styles should be at the bottom of the imports list

src/App.tsx
```diff
@@ -1,25 +1,120 @@
...
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
 export const App = () => {
+  const [newToDo, setNewToDo] = useState({
+    title: '',
+    user: '',
+  });
+  const todosCopy = [...todosFromServer];
+  const copy = [...usersFromServer];
```
VitaliyBodnar, line 13: I don't see any sense to make two identical copies of same array

src/App.tsx
```diff
@@ -1,25 +1,120 @@
...
+    if (eventUser === '0') {
+      setNewToDo(prevState => ({
+        title: prevState.title,
+        user: 'alarm',
+      }));
+      setTimeout(() => setNewToDo(prevState => ({
+        title: prevState.title,
+        user: '',
+      })), 1000);
+    } else if (eventTitle === '') {
```
VitaliyBodnar, line 32: don't forget about empty space in an input value, so please use .trim()

src/App.tsx
```diff
@@ -1,25 +1,120 @@
...
+
+  const [toDoList, setToDoList] = useState(todosCopy);
+
+  function submitHandler(event: React.FormEvent<HTMLFormElement>) {
+    event.preventDefault();
+    const maxId = Math.max(...todosCopy.map(todo => +todo.id));
+    const eventUser = (event.currentTarget[1] as HTMLInputElement).value;
+    const eventTitle = (event.currentTarget[0] as HTMLInputElement).value;
+
+    if (eventUser === '0') {
```
VitaliyBodnar, line 23: what is the magic number '0'? please move it to the constant

src/App.tsx
```diff
@@ -1,25 +1,120 @@
...
+  const todosCopy = [...todosFromServer];
+  const copy = [...usersFromServer];
+
+  const [toDoList, setToDoList] = useState(todosCopy);
+
+  function submitHandler(event: React.FormEvent<HTMLFormElement>) {
+    event.preventDefault();
+    const maxId = Math.max(...todosCopy.map(todo => +todo.id));
+    const eventUser = (event.currentTarget[1] as HTMLInputElement).value;
+    const eventTitle = (event.currentTarget[0] as HTMLInputElement).value;
```
VitaliyBodnar, line 21: make sense to destructure event.currentTarget

src/App.tsx
```diff
@@ -1,25 +1,120 @@
...
+
+    if (eventUser === '0') {
+      setNewToDo(prevState => ({
+        title: prevState.title,
+        user: 'alarm',
+      }));
+      setTimeout(() => setNewToDo(prevState => ({
+        title: prevState.title,
+        user: '',
+      })), 1000);
```
VitaliyBodnar, line 31: why do we need this setTimeout() ?

src/App.tsx
```diff
@@ -1,25 +1,120 @@
...
+            onChange={(event) => handleSelect(event)}
+          >
+            <option value="0">Choose a user</option>
+            {copy.map((user) => {
+              return (
+                <option value={user.id} key={user.id}>
+                  {user.name}
+                </option>
+              );
+            })}
```
VitaliyBodnar, line 108:
```suggestion
            {copy.map((user) => (
                <option value={user.id} key={user.id}>
                  {user.name}
                </option>
              ))}
```

src/App.tsx
```diff
@@ -28,33 +123,22 @@ export const App = () => {
...
+            <article
+              data-id={todo.id}
+              className="TodoInfo TodoInfo--completed"
+            >
+              <h2 className="TodoInfo__title">
+                {todo.title}
+              </h2>
+
+              <a className="UserInfo" href="mailto:Sincere@april.biz">
+                {copy.find(user => user.id === todo.userId)?.name}
```
VitaliyBodnar, line 137: don't call the function from the 'template render', move it to state

VitaliyBodnar (changes requested):
You need to show these validation errors only when the user submitted the form ( clicked add )
Screenshot 2023-06-11 at 16 19 13

src/App.tsx
haduigon: spaces should be allowed )
Screenshot 2023-06-11 at 21 47 05

src/App.tsx
haduigon: it is different arrays, one of them is about todos, another is about the users )

src/App.tsx
haduigon: it is all due to my logic of element )

src/App.tsx
haduigon: The logic is - to hide warning messages only if select and title are not empty, if you press 'add' messages change color for 1 sec, actually I need setTimeout especially for this purpose ) It seemed to me more interesting but I can rewrite strictly according to the task)

haduigon: fixed

src/App.tsx
haduigon: fixed

src/App.tsx
haduigon: fixed

src/App.tsx
haduigon: fixed

src/App.tsx
haduigon: fixed)

src/App.tsx
haduigon: fixed

demetrxx (approved):
Well done 👍 Go ahead!

PR #1746, opened 2023-06-11
Developed

devWizHub:
[DEMO LINK] https://devWizHub.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,26 @@
 import './App.scss';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
 
-export const App = () => {
-  return (
-    <div className="App">
-      <h1>Add todo form</h1>
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
 
-      <form action="/api/users" method="POST">
-        <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
-        </div>
+  return foundUser || null;
+}
 
-        <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>Choose a user</option>
-          </select>
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
-          <span className="error">Please choose a user</span>
-        </div>
+export const App: React.FC = () => (
+  <div className="App">
+    <h1 className="App__title">Dynamic list of todos</h1>
+    <TodoList todos={todos} />
 
-        <button type="submit" data-cy="submitButton">
-          Add
-        </button>
-      </form>
-
-      <section className="TodoList">
-        <article data-id="1" className="TodoInfo TodoInfo--completed">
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
-          <span className="error">Please enter a title</span>
+          <span className="">Title: </span>
+          <input
+            type="text"
+            data-cy="titleInput"
+            placeholder="Enter a title"
+            value={inputValue}
+            onChange={(e) => setInputValue(e.target.value)}
+          />
+          {inputError && <span style={{ color: 'red' }}>{inputError}</span>}
```
Anastasiia-Svintsova, line 83: It's better to use a className for styling error messages instead of inline styles.

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+                <option
+                  key={person.id}
+                  value={person.id}
+                >
+                  {person.name}
+                </option>
+              );
+            })}
+          </select>
+          {userError && <span style={{ color: 'red' }}>{userError}</span>}
```
Anastasiia-Svintsova, line 108: It's better to use a className for styling error messages instead of inline styles.

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+}
+
+const initialTodos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [selectedUser, setSelectedUser]
+  = useState< number | undefined>(undefined);
```
Anastasiia-Svintsova, line 23: You can use a 0 here for the initial value 
```suggestion
  const [selectedUser, setSelectedUser] = useState(0);
```

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+      setUserError('Please choose a user');
+
+      return;
+    }
+
+    if (!inputValue.trim()) {
+      setInputError('Please enter a title');
+
+      return;
+    }
```
Anastasiia-Svintsova, line 42: The user should see both errors if the user is not selected and the title is empty at the same time. 
Now it shows only one error at a time. Try to refactor it a bit.

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
-          </select>
+          <span className="">User: </span>
 
-          <span className="error">Please choose a user</span>
+          <select
+            data-cy="userSelect"
+            value={selectedUser}
+            onChange={handleUserChange}
+          >
+            <option value="">Choose a user</option>
```
Anastasiia-Svintsova, line 95: Let's change the value to 0 to match the other values type.

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,15 @@
-export const TodoList = () => {};
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo';
+
+interface ToDoListProps {
+  todos: Todo[];
+}
+
+export const TodoList: React.FC <ToDoListProps> = ({ todos }) => (
```
Anastasiia-Svintsova, line 9: Remove the unnecessary space before the angle bracket.
```suggestion
export const TodoList: React.FC<ToDoListProps> = ({ todos }) => (
```

Anastasiia-Svintsova (changes requested):
- Implement this requirement, please: If the form is valid, add a todo to the list and clear the form. Clear the user as well

https://github.com/mate-academy/react_add-todo-form/assets/99403198/1feed043-be58-4c49-b2f5-346515a55e5d

PR #1750, opened 2023-06-14
Fixed

IvanFront-end:
[DEMO LINK](https://IvanFront-end.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,1 +1,28 @@
-export const TodoInfo = () => {};
+import usersFromServer from '../../api/users';
+import { User } from '../../types/User';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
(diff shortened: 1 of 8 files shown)


TarasHoliuk (changes requested):
To test purpose

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,28 @@
...
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const user = getUser(todo.userId);
+
+  return (
+    <article data-id="1" className={`TodoInfo ${todo.completed ? 'TodoInfo--completed' : ''}`}>
```
TarasHoliuk, line 20: FYI: it's better to use `classnames` library in case when you have conditional className

TarasHoliuk (approved):

PR #1759, opened 2023-06-18
solution

Yuliya1907:
[DEMO LINK](https://Yuliya1907.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,136 @@
+import React, { useState } from 'react';
+import { TodoList } from './components/TodoList';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-export const App = () => {
+const getUserById = (userId: number): User | null => {
+  const findUser = usersFromServer.find((user) => (user.id === userId));
+
+  return findUser || null;
+};
+
+const getTodoId = (listTodos: Todo[]): number => {
+  const todosId = listTodos.map(todo => todo.id);
+
+  return Math.max(...todosId) + 1;
+};
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [listTodos, setListTodos] = useState(todos);
+  const [title, setTitle] = useState('');
+  const [user, setUser] = useState(0);
+  const [titleError, setTitleError] = useState(false);
+  const [userError, setUserError] = useState(false);
+
+  const changeTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitleError(false);
+    setTitle(event.target.value);
+  };
+
+  const changeUser = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setUserError(false);
+    setUser(Number(event.target.value));
+  };
+
+  const submit = (event: React.ChangeEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (!title) {
+      setTitleError(true);
+    }
+
+    if (!user) {
+      setUserError(true);
+    }
+
+    if (title && user) {
+      const newTodo: Todo = {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,136 @@
...
+        title,
+        userId: user,
+        completed: false,
+        user: getUserById(user),
+      };
+
+      setListTodos(prevlist => [...prevlist, newTodo]);
+      setTitle('');
+      setUser(0);
+    }
```
dffUqp, line 67: We can get rid of the last if statement. 
Let's add `return` in error case.

src/App.tsx
```diff
@@ -1,61 +1,140 @@
...
+  const findUser = usersFromServer.find((user) => (user.id === userId));
+
+  return findUser || null;
+};
+
+export const getTodoId = (listTodos: Todo[]): number => {
+  const todosId = listTodos.map(todo => todo.id);
+
+  return Math.max(...todosId) + 1;
+};
```
33Orange, line 20: Do you need this exports when you used this functions in same file?
Move this functions to a separate file `helpers`.

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,31 @@
...
+    user,
+  } = todo;
+
+  return (
+    <article
+      data-id={id}
+      className={classNames(
+        'TodoInfo',
+        { 'TodoInfo--completed': completed },
+      )}
```
33Orange, line 24: Remove unnecessary linebreaks
```suggestion
      className={classNames('TodoInfo',{ 
       'TodoInfo--completed': completed 
       })}
```

33Orange (changes requested):
Almost done

src/App.tsx
```diff
@@ -1,61 +1,128 @@
...
+      setTitleError(true);
+    }
+
+    if (!user) {
+      setUserError(true);
+    }
+
+    if (!title.trim() || !user) {
+      return;
+    }
```
dffUqp, line 47: we repeat `title.trim()` 2 times, let's create a variable for it

src/types/User.ts
```diff
@@ -0,0 +1,5 @@
+export interface User {
```
roman-mirzoian, line 1: Now the name `User` is well understood in the type file itself, but when we reuse this type, it will be more convenient to give a name that will directly inform that it is a type, interface, enam, and so on. (For example: `IUser`, `UserInterface`, `TUser`, `UserType`).

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,17 @@
-export const TodoList = () => {};
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
```
roman-mirzoian, line 5: You've already started nicely to export typing to separate files, so it would be cool to continue here as well, and export the type for the `TodoList` component to the `Todo.ts` file and give this type a more convenient name (For example: `TodoListType`).

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,30 @@
-export const TodoInfo = () => {};
+import React from 'react';
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
```
roman-mirzoian, line 6: You've already started nicely to export typing to separate files, so it would be cool to continue here as well, and export the type for the `TodoInfo` component to the `Todo.ts` file and give this type a more convenient name (For example: `TodoInfoType`).

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,30 @@
...
+    id,
+    completed,
+    title,
+    user,
+  } = todo;
+
+  return (
+    <article
+      data-id={id}
+      className={classNames('TodoInfo', {
```
roman-mirzoian, line 21: It will be much easier to read the code if you transfer such class definitions to separate variables and use the ready result. You can see a usage example on the [classnames](https://www.npmjs.com/package/classnames) package site.

naziamoff (approved):

PR #1762, opened 2023-06-19
solution

okosohor:
[DEMO LINK](https://okosohor.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,110 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types';
+
+const getUser = (userId: number) => {
+  const foundedUser = usersFromServer.find(user => (
+    userId === user.id
+  ));
+
+  return foundedUser || null;
+};
+
+const todosWithUser: Todo[] = todosFromServer.map(todo => (
+  {
+    ...todo,
+    user: getUser(todo.userId),
+  }
+));
 
 export const App = () => {
+  const [user, setUser] = useState(0);
+  const [title, setTitle] = useState('');
+  const [todos, setTodos] = useState([...todosWithUser]);
+  const [titleError, setTitleError] = useState(false);
+  const [userError, setUserError] = useState(false);
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
+        action="/api/users"
+        method="POST"
+        onSubmit={(event) => {
+          event.preventDefault();
+          if (title === '') {
+            setTitleError(true);
+          }
+
+          if (user === 0) {
+            setUserError(true);
+          }
+
+          if (title === '' && user === 0) {
+            setTitleError(true);
+            setUserError(true);
+          }
+
+          if (title !== '' && user !== 0) {
+            const maxId = todos.map((todo) => (todo.id));
+            const newId = Math.max(...maxId) + 1;
+            const newTodo: Todo = {
+              id: newId,
+              userId: user,
+              title,
+              completed: false,
+              user: getUser(user),
+            };
+
+            setTodos([...todos, newTodo]);
+
+            setTitle('');
+            setUser(0);
+          }
+        }}
+      >
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <input
+            type="text"
+            data-cy="titleInput"
+            placeholder="Enter a title"
+            value={title}
+            onChange={(event) => {
+              setTitle(event.target.value);
+              setTitleError(false);
+            }}
+          />
+          {titleError && (<span className="error">Please enter a title</span>)}
         </div>
 
         <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>Choose a user</option>
+          <select
+            data-cy="userSelect"
+            value={user}
+            onChange={(event) => {
+              setUser(+event.target.value);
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,61 +1,123 @@
...
+    if (title !== '' && user !== 0) {
+      const newTodo: Todo = {
+        id: newId(todos),
+        userId: user,
+        title,
+        completed: false,
+        user: getUser(user),
+      };
+
+      setTodos([...todos, newTodo]);
```
33Orange, line 66: Good practice to use prevstate in such cases.

```suggestion
      setTodos(prevTodos => [ ...prevTodos, newTodo]);
```

src/App.tsx
```diff
@@ -1,61 +1,123 @@
...
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
+        action="/api/users"
+        method="POST"
+        onSubmit={(event) => (handleSubmit(event))}
```
33Orange, line 80: Why are you creating this anonymous function here ?

```suggestion
        onSubmit={handleSubmit}
```

src/App.tsx
```diff
@@ -1,61 +1,123 @@
...
+      >
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <input
+            type="text"
+            data-cy="titleInput"
+            placeholder="Enter a title"
+            value={title}
+            onChange={(event) => (handleTitleSelect(event))}
```
33Orange, line 88: Same here

src/App.tsx
```diff
@@ -1,61 +1,123 @@
...
+          {titleError && (<span className="error">Please enter a title</span>)}
         </div>
 
         <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>Choose a user</option>
+          <select
+            data-cy="userSelect"
+            value={user}
+            onChange={(event) => (handleUserSelect(event))}
```
33Orange, line 97: And same here

src/App.tsx
```diff
@@ -1,61 +1,123 @@
...
+            <option value={0} disabled>
+              Choose a user
+            </option>
+
+            {usersFromServer.map(userFromServer => (
+              userFromServer && (
+                <option key={userFromServer.id} value={userFromServer.id}>
+                  {userFromServer.name}
+                </option>
+              )))}
```
33Orange, line 108: In which cases might `userFromServer` not exist in the array? What is this check for?
Use destructuring assignment here

```suggestion
            {usersFromServer.map(({name, id})=> (
                <option key={id} value={id}>{name}</option>
              ))}
```

33Orange (changes requested):
Almost done

src/App.tsx
```diff
@@ -5,42 +5,29 @@ import usersFromServer from './api/users';
...
-  const [todos, setTodos] = useState([...todosWithUser]);
+  const [todos, setTodos] = useState(todosWithUser);
   const [titleError, setTitleError] = useState(false);
   const [userError, setUserError] = useState(false);
 
-  const handleTitleSelect = (event: React.ChangeEvent<HTMLInputElement>) => {
-    setTitle(event.target.value);
-    setTitleError(false);
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    if (event.target.value.trim() !== '') {
```
And678, line 23: Because of this, I can't remove the first symbol in the input.
image

You should prevent **submitting** wrong data, not **inputting** it.

And678 (changes requested):
Your demo link displays the previous version. Don't forget to run `npm run deploy` after pushing new changes

src/App.tsx
```diff
@@ -1,61 +1,109 @@
...
+      setTitle('');
+
+      return;
+    }
+
+    if (!userId) {
+      setUserError(true);
+    }
+
+    if (title === '' || userId === 0) {
```
dffUqp, line 45:
```suggestion
    if (!title' || !userId) {
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,30 @@
...
+    <article
+      data-id={id}
+      className={cn('TodoInfo', { 'TodoInfo--completed': completed })}
+    >
+      <h2 className="TodoInfo__title">
+        {title}
+      </h2>
+
+      {user && <UserInfo user={user} />}
+
```
dffUqp, line 27:
```suggestion

```

src/App.tsx
```diff
@@ -1,61 +1,109 @@
...
+            data-cy="userSelect"
+            value={userId}
+            onChange={handleUserSelect}
+          >
+            <option value={0} disabled>
+              Choose a user
+            </option>
+
+            {usersFromServer.map(({ name, id }) => (
+              <option key={id} value={id}>{ name }</option>
```
dffUqp, line 93:
```suggestion
              <option key={id} value={id}>{name}</option>
```

dffUqp (approved):
:fire:

src/App.tsx
```diff
@@ -1,61 +1,109 @@
...
+    setUserError(false);
+  };
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    if (!title.trim()) {
+      setTitleError(true);
+      setTitle('');
+
+      return;
```
dffUqp, line 38: to pass the test, you need to remove this `return` and add `!title.trim()` on line 45

if i submit the form with empty field i will only see one error message
image

src/App.tsx
```diff
@@ -1,61 +1,109 @@
...
+    }
```
dffUqp, line 39: to pass the test, you need to remove this `return` and add `!title.trim()` on line 45

if i submit the form with empty field i will only see one error message
image

PR #1765, opened 2023-06-19
solution

annholovko:
[DEMO LINK](https://annholovko.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,118 @@
 import './App.scss';
+import { FC, useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
 
-export const App = () => {
+import { TodoList } from './components/TodoList';
+import { Todo } from './components/Types/Todo';
+
+import {
+  findUserById,
+  preparedTodos,
+  newTodoId,
+} from './components/Helpers/helpers';
+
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,118 @@
...
+    if (!title || !selectedUser) {
+      return;
+    }
+
+    const newTodo: Todo = {
+      id: newTodoId,
+      title: title.trim(),
+      completed: false,
+      userId: Number(selectedUser),
+      user: findUserById(Number(selectedUser)),
```
dffUqp, line 41: I think `selectedUser ` is already a number, or am I wrong?🤔

src/App.tsx
```diff
@@ -1,61 +1,118 @@
...
+    };
+
+    setTodos([...todos, newTodo]);
+    setTitle('');
+    setSelectedUser(0);
+  };
+
+  const changeTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
```
dffUqp, line 49: good practice to name the event handler as handle[eventName]:smiley:

`handleTitleChange`
`handleUserChange`
`handleSubmit`

dffUqp (changes requested):
LGTM🔥 but needs a few fixes:upside_down_face:

image

33Orange (approved):
Well done

PR #1767, opened 2023-06-19
add todo form

vchervinko:
[DEMO LINK](https://vchervinko.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,124 @@
+import React, { useState } from 'react';
+import { TodoList } from './components/TodoList';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
 import './App.scss';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+const getUserById = (userId: number): User | null => (
+  usersFromServer.find((user) => (user.id === userId)) || null
+);
 
-export const App = () => {
+const getUserByName = (userName: string): User | null => (
+  usersFromServer.find((user) => (user.name === userName)) || null
+);
+
+const newTodos: Todo[] = todosFromServer.map((todo) => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
+
+export const App: React.FC = () => {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,132 @@
+import React, { useState } from 'react';
+import { TodoList } from './components/TodoList';
+import { User } from './types/User';
```
33Orange, line 3: You didn't add files to commit

image

src/App.tsx
```diff
@@ -1,61 +1,132 @@
...
+
+    setIsTitleValid(Boolean(title.trim()));
+    setIsUserValid(Boolean(selectedUser));
+
+    if (!title.trim() || !selectedUser) {
+      return;
+    }
+
+    setTodos((currentTodos) => {
+      const maxToDoId = Math.max(...currentTodos.map(todo => todo.id));
```
33Orange, line 54: Create a separate function which will be creating new id and move it to separate file `helpers`

src/App.tsx
```diff
@@ -1,61 +1,132 @@
...
+      return [
+        ...currentTodos,
+        {
+          id: maxToDoId + 1,
+          title,
+          completed: false,
+          userId: newUser ? newUser.id : null,
+          user: newUser,
+        },
+      ];
```
33Orange, line 66: Let's create a variable fro new todo to improve readability

```suggestion
      cosnt newTodo = {...}
      return [ ...currentTodos, newTodo];
```

src/App.tsx
```diff
@@ -1,61 +1,132 @@
...
+    setIsTitleValid(Boolean(title.trim()));
+    setIsUserValid(Boolean(selectedUser));
+
+    if (!title.trim() || !selectedUser) {
+      return;
+    }
+
+    setTodos((currentTodos) => {
+      const maxToDoId = Math.max(...currentTodos.map(todo => todo.id));
+      const newUser = getUserByName(selectedUser);
```
33Orange, line 55: Use `user.id` as value in select and you can use `getUserById`  here and remove redundant `getUserByName`

33Orange (changes requested):
Almost done

A few moments to fix

PR #1773, opened 2023-06-20
solution

maryanmats:
[DEMO LINK](https://maryanmats.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,106 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { Todo } from './types/Todo';
+import { User } from './types/User';
+import { TodoList } from './components/TodoList';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  // if there is no user with a given userId
+  return foundUser || null;
+}
+
+export const getTodos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [todos, setTodos] = useState(getTodos);
+  const [selectedUser, setSelectedUser] = useState(0);
+  const [titleError, setTitleError] = useState(false);
+  const [userError, setUserError] = useState(false);
+
+  const formOnSubmit = (event: React.ChangeEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (title === '') {
+      setTitleError(true);
+    }
+
+    if (selectedUser === 0) {
+      setUserError(true);
+    }
+
+    if (title === '' && selectedUser === 0) {
+      setTitleError(true);
+      setUserError(true);
+    }
+
+    if (title !== '' && selectedUser !== 0) {
+      const maxId = todos.map((todo) => (todo.id));
+      const newId = Math.max(...maxId) + 1;
+
+      const newTodo: Todo = {
+        id: newId,
+        completed: false,
+        user: getUser(selectedUser),
+        title,
+        userId: selectedUser,
+      };
+
+      setTodos([...todos, newTodo]);
+    }
+
+    setTitle('');
+    setSelectedUser(0);
+  };
+
+  const handleChangeUser = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setSelectedUser(+event.target.value);
+    setUserError(false);
+  };
+
+  const handleChangeTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+    setTitleError(false);
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form action="/api/users" method="POST" onSubmit={formOnSubmit}>
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <input
+            type="text"
+            data-cy="titleInput"
+            placeholder="Enter a title"
+            value={title}
+            onChange={handleChangeTitle}
+          />
+          {titleError && (<span className="error">Please enter a title</span>)}
         </div>
 
         <div className="field">
-          <select data-cy="userSelect">
+          <select
+            data-cy="userSelect"
+            value={selectedUser}
+            onChange={handleChangeUser}
+          >
             <option value="0" disabled>Choose a user</option>
-          </select>
 
-          <span className="error">Please choose a user</span>
+            {usersFromServer.map(user => (user
+              && (<option value={user.id} key={user.id}>{user.name}</option>)
+            ))}
+
+          </select>
+          {userError && (<span className="error">Please choose a user</span>)}
         </div>
 
         <button type="submit" data-cy="submitButton">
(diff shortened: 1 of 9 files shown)


silchyn (changes requested):
- Enter title -> submit -> observe title being cleared and validation error for user appearing
- Form can be submitted with a title consisting of spaces

dinaminahi (changes requested):
Still can add todo without title(with spaces)
image
Let's validate it

src/types/User.ts
```diff
@@ -0,0 +1,5 @@
+export interface User {
```
roman-mirzoian, line 1: Now the name is clear in the context of this file, but if you use this interface in many places, confusion may arise, so it is better to give more precise names, for example `UserInterface`/`IUser` or `UserType`/`TUser` (if it will be a type).

src/types/Todo.ts
```diff
@@ -0,0 +1,9 @@
+import { User } from './User';
+
+export interface Todo {
```
roman-mirzoian, line 3: Now the name is clear in the context of this file, but if you use this interface in many places, confusion may arise, so it is better to give more precise names, for example `TodoInterface`/`ITodo` or `TodoType`/`TTodo` (if it will be a type).

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,11 @@
-export const UserInfo = () => {};
+import { User } from '../../types/User';
+
+type Props = {
```
roman-mirzoian, line 3: Similar names are okay for a small project, but on large projects we will have many interfaces and types, so it is better to give more detailed names, for example `UserInfoProps` or `UserInfoType`.
This applies to all code.

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,17 @@
-export const TodoList = () => {};
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
```
roman-mirzoian, line 4: Similar names are okay for a small project, but on large projects we will have many interfaces and types, so it is better to give more detailed names, for example `TodoListProps` or `TodoListType`.
This applies to all code.

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,17 @@
...
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="TodoList">
+      {todos.map(todo => (
+        <TodoInfo key={todo.id} todo={todo} />
+      ))}
+
```
roman-mirzoian, line 14: Don't forget about good formatting (don't leave extra blank spaces) :)

src/App.tsx
```diff
@@ -1,61 +1,111 @@
...
+
+    const newTodo: Todo = {
+      id: newId,
+      completed: false,
+      user: getUser(Number(selectedUser)),
+      title: title.trim(),
+      userId: selectedUser,
+    };
+
+    setTodos([...todos, newTodo]);
```
roman-mirzoian, line 55: When you work with updating a state that already has some data, it is better to update it using a callback function, the first argument of which will be the actual current state. This will help avoid errors when rerendering the component.

roman-mirzoian (changes requested):
The work is great and the app works as it should, only a little optimization of the code remains to be done.

src/App.tsx
```diff
@@ -1,61 +1,111 @@
...
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { Todo } from './types/Todo';
+import { User } from './types/User';
+import { TodoList } from './components/TodoList';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  // if there is no user with a given userId
```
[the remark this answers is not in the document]
naziamoff: not fixed

src/App.tsx
```diff
@@ -1,61 +1,111 @@
...
+  const [userError, setUserError] = useState(false);
+
+  const formOnSubmit = (event: React.ChangeEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (!title.trim()) {
+      setTitleError(true);
+    }
+
+    if (!selectedUser) {
```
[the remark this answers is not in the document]
naziamoff: not fixed

src/App.tsx
```diff
@@ -1,61 +1,111 @@
...
+      setUserError(true);
+    }
+
+    if (!selectedUser || !title.trim()) {
+      return;
+    }
+
+    const maxId = todos.map((todo) => (todo.id));
+    const newId = Math.max(...maxId) + 1;
```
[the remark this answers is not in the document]
naziamoff: not fixed

naziamoff (changes requested):
you did not fix the requested changes from the previous reviews

src/App.tsx
```diff
@@ -1,61 +1,105 @@
...
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [todos, setTodos] = useState(getTodos);
+  const [selectedUserID, setSelectedUserID] = useState(0);
+  const [titleError, setTitleError] = useState(false);
+  const [userError, setUserError] = useState(false);
+
+  const isTitleExist = !selectedUserID || !title.trim();
+
+  const formOnSubmit = (event: React.ChangeEvent<HTMLFormElement>) => {
```
naziamoff, line 24: how do we usually name our methods?

src/App.tsx
```diff
@@ -1,61 +1,105 @@
...
+      completed: false,
+      user: getUser(Number(selectedUserID)),
+      title: title.trim(),
+      userId: selectedUserID,
+    };
+
+    setTodos((prevTodos) => [...prevTodos, newTodo]);
+
+    setTitle('');
+    setSelectedUserID(0);
```
naziamoff, line 52: we use `null` for explicitly not defined values

src/App.tsx
```diff
@@ -1,61 +1,105 @@
...
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <input
+            type="text"
+            data-cy="titleInput"
+            placeholder="Enter a title"
+            value={title}
+            onChange={handleChangeTitle}
+          />
+          {titleError && (<span className="error">Please enter a title</span>)}
```
naziamoff, line 78:
```suggestion
          {titleError && <span className="error">Please enter a title</span>}
```

src/App.tsx
```diff
@@ -1,61 +1,105 @@
...
+            onChange={handleChangeUser}
+          >
             <option value="0" disabled>Choose a user</option>
-          </select>
 
-          <span className="error">Please choose a user</span>
+            {usersFromServer.map(user => (user
+              && (<option value={user.id} key={user.id}>{user.name}</option>)
+            ))}
+
```
naziamoff, line 92: If user can be nullish in this array - something went wrong in the code above
```suggestion
            {usersFromServer.map(({ id, name }) => (
                <option value={id} key={id}>{name}</option>
              )))}
```

src/App.tsx
```diff
@@ -1,61 +1,105 @@
...
+          </select>
+          {userError && (<span className="error">Please choose a user</span>)}
```
naziamoff, line 94:
```suggestion
          {userError && <span className="error">Please choose a user</span>}
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,23 @@
...
+import { UserInfo } from '../UserInfo';
+
+type TodoInfoType = {
+  todo: TodoType;
+};
+
+export const TodoInfo: React.FC<TodoInfoType> = ({ todo }) => {
+  const todoInfoClasses = classNames('TodoInfo', {
+    'TodoInfo--completed': todo.completed,
+  });
```
naziamoff, line 12: looks super weird. Let's just move this logic to the `className` - to where it belongs

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,16 @@
-export const TodoList = () => {};
+import { TodoType } from '../../types/TodoType';
+import { TodoInfo } from '../TodoInfo';
+
+type TodoListProps = {
+  todos: TodoType[];
+};
+
+export const TodoList: React.FC<TodoListProps> = ({ todos }) => {
+  return (
```
naziamoff, line 9: as you don't have any logic before return - we should make right-away return statement in our arrow funciton
```suggestion
export const TodoList: React.FC<TodoListProps> = ({ todos }) => (
```

naziamoff (changes requested):

PR #1781, opened 2023-06-20
solution

thienlaoo:
[DEMO LINK](https://thienlaoo.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,123 @@
 import './App.scss';
+import React, { useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Todo, User } from './components/Types';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+function getUserById(userId: number): User | null {
+  const userById = usersFromServer.find(user => user.id === userId);
+
+  return userById || null;
+}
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+      return;
+    }
+
+    if (!userSelect) {
+      setUserError(true);
+
+      return;
+    }
+
+    const receivedUser = getUserByName(userSelect);
```
33Orange, line 47: Use `user.id` as value in select and use `getUserById` to find.

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+      ...todos,
+      {
+        id: maxTodoId + 1,
+        title: titleInput,
+        completed: false,
+        userId: receivedUser?.id,
+        user: receivedUser,
+      },
+    ]);
+  };
```
33Orange, line 62: Let's create a separate variable for new todo to improve readability 

```suggestion
     const newTodo = {
        id: maxTodoId + 1,
        title: titleInput,
        completed: false,
        userId: receivedUser?.id,
        user: receivedUser,
    },
    setTodos([ ...todos, newTodo ]);
  };
```

33Orange (changes requested):
Almost done

**Please use `npm run test` before push PR and fix all problems which tests indicate**

PR #1785, opened 2023-06-20
add task solution

Yuliia-kucheriavenko:
[DEMO LINK](https://Yuliia-kucheriavenko.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,136 @@
+import React, { useState } from 'react';
 import './App.scss';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-export const App = () => {
+const getUserById = (userId: number): User | null => (
+  usersFromServer.find(user => (user.id === userId)) || null
+);
+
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,136 @@
...
+      setIsUserValid(true);
+    }
+
+    if (!userSelect || !title) {
+      return;
+    }
+
+    setTodos((currentTodos) => {
+      const maxTodoId = Math.max(...currentTodos.map(todo => todo.id));
+      const newTodo = getUserByName(userSelect);
```
dffUqp, line 47: when we change user we store his id on our state, can we remove `getUserByName` and start using `getUserById`?

src/App.tsx
```diff
@@ -1,61 +1,136 @@
...
+    if (!userSelect) {
+      setIsUserValid(true);
+    }
+
+    if (!userSelect || !title) {
+      return;
+    }
+
+    setTodos((currentTodos) => {
+      const maxTodoId = Math.max(...currentTodos.map(todo => todo.id));
```
dffUqp, line 46: let's create a separate function to calculate todo id

dffUqp (changes requested):


image

roman-mirzoian (changes requested):
The app is nicely updated, but there are some suggestions for code optimization and function naming. I also see that there are still some issues left after the previous review.

PR #1786, opened 2023-06-20
add solution

ulositska:
[DEMO LINK](https://ulositska.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,133 @@
+import { useState, ChangeEvent } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { ToDo } from './types/types';
+
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+const findUserById = (userId:number | null) => {
+  return usersFromServer.find(user => user.id === userId) || null;
+};
+
+const extendedTodos = todosFromServer.map(todo => ({
+  ...todo,
+  user: findUserById(todo.userId),
+} || null));
 
 export const App = () => {
+  const [todos, setTodos] = useState<ToDo[]>(extendedTodos);
+  const [title, setTitle] = useState('');
+  const [selectedUserId, setselectedUserId] = useState<number | null>(null);
+
+  const [titleError, setTitleError] = useState('');
+  const [userError, setUserError] = useState('');
+
+  const handleTitleChange = (event: ChangeEvent<HTMLInputElement>) => {
+    if (titleError !== '') {
+      setTitleError('');
+    }
+
+    setTitle(event.target.value);
+  };
+
+  const handleUserChange = (event: ChangeEvent<HTMLSelectElement>) => {
+    if (userError !== '') {
+      setUserError('');
+    }
+
+    setselectedUserId(Number(event.target.value));
+  };
+
+  const handleFormSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (title.trim() === '') {
+      setTitleError('Please enter a title');
+    }
+
+    if (!selectedUserId) {
+      setUserError('Please choose a user');
+    }
+
+    if (!selectedUserId || !title.trim()) {
+      return;
+    }
+
+    setTodos((prevTodos) => {
+      const maxTodoId = Math.max(...prevTodos.map(todo => todo.id));
+
+      const newTodo: ToDo = {
+        id: maxTodoId + 1,
+        title,
+        userId: maxTodoId,
+        completed: false,
+        user: findUserById(selectedUserId),
+      };
+
+      return [
+        ...prevTodos,
+        newTodo,
+      ];
+    });
+
+    setTitle('');
+    setselectedUserId(null);
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+      setTitleError('Please enter a title');
+    }
+
+    if (!selectedUserId) {
+      setUserError('Please choose a user');
+    }
+
+    if (!selectedUserId || !title.trim()) {
+      return;
+    }
```
dffUqp, line 56: we repeat `title.trim()` 2 times, let's create a variable for it.

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+
+    setTodos((prevTodos) => {
+      const maxTodoId = Math.max(...prevTodos.map(todo => todo.id));
```
dffUqp, line 59: let's create a separate function for `todoId` calculation

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+
+            {usersFromServer.map(user => (
+              <option key={user.id} value={user.id}>
+                {user.name}
+              </option>
+            ))}
           </select>
 
-          <span className="error">Please choose a user</span>
+          {userError && <span className="error">Please choose a user</span>}
```
dffUqp, line 120: you store a error message in `userError` so let's use it:smiley:
```suggestion
          {userError && <span className="error">{userError}</span>}
```

dffUqp (changes requested):

src/types/types.tsx
```diff
@@ -0,0 +1,14 @@
+export type User = {
```
roman-mirzoian, line 1: Now the name `User` and `Todo` is well understood in the types file itself, but when we reuse this types, it will be more convenient to give a name that will directly inform that it is a types, interfaces, enams, and so on. (For example: `IUser/ITodo`, `UserInterface/TodoInterface`, `TUser/TTodo`, `UserType/TodoType`).

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,16 @@
-export const UserInfo = () => {};
+import { FC } from 'react';
+import { User } from '../../types/types';
+
+type Props = {
```
roman-mirzoian, line 4: You've already started nicely to export typing to separate files, so it would be cool to continue here as well, and export the type for the `UserInfo` component to the types.tsx file and give this type a more convenient name (For example: `UserInfoType`).

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,16 @@
-export const TodoList = () => {};
+import { TodoInfo } from '../TodoInfo';
+import { ToDo } from '../../types/types';
+
+type Props = {
```
roman-mirzoian, line 4: You've already started nicely to export typing to separate files, so it would be cool to continue here as well, and export the type for the `TodoList` component to the types.tsx file and give this type a more convenient name (For example: `TodoListType`).

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,37 @@
-export const TodoInfo = () => {};
+import cn from 'classnames';
+import { FC } from 'react';
+import { UserInfo } from '../UserInfo';
+import { ToDo } from '../../types/types';
+
+type Props = {
```
roman-mirzoian, line 6: You've already started nicely to export typing to separate files, so it would be cool to continue here as well, and export the type for the `TodoInfo` component to the types.tsx file and give this type a more convenient name (For example: `TodoInfoType`).

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,37 @@
-export const TodoInfo = () => {};
+import cn from 'classnames';
```
roman-mirzoian, line 1: In the context of a small application, an import with the name `"cn"` is possible, but in larger applications, where there may be many imports, it is better not to shorten the names to unreadable ones, but to use longer but more obvious names (`import classNames from 'classnames'`).

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,37 @@
...
+    id,
+    title,
+    completed,
+    user,
+  } = todo;
+
+  return (
+    <article
+      data-id={id}
+      className={cn(
```
roman-mirzoian, line 21: It will be much easier to read the code if you transfer such class definitions to separate variables and use the ready result. You can see a usage example on the [classnames](https://www.npmjs.com/package/classnames) package site.

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
 export const App = () => {
+  const [todos, setTodos] = useState<ToDo[]>(extendedTodos);
+  const [title, setTitle] = useState('');
+  const [selectedUserId, setselectedUserId] = useState<number | null>(null);
+
+  const [titleError, setTitleError] = useState('');
+  const [userError, setUserError] = useState('');
+
+  const handleTitleChange = (event: ChangeEvent<HTMLInputElement>) => {
+    if (titleError !== '') {
```
roman-mirzoian, line 28: Here it would be better to check for an empty `title` because now I can create a new todo with an empty title.

roman-mirzoian (changes requested):
The app is nicely updated, but some updates can be made to optimize the code.
Also, some comments after the previous review remained unresolved.

PR #1787, opened 2023-06-20
add solution

ivankhrsh:


src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
+        setSelectedOption(0);
+      } else {
+        if (selectedOption === 0) {
+          setIsValidSelectedOption(false);
+        }
+
+        if (title.length === 0) {
+          setIsValidTitle(false);
+        }
+      }
```
dffUqp, line 60: try to come up with a better solution for this case🤔

I think we can remove nested if statement.

dffUqp (changes requested):


image

dffUqp (approved):
it seems you forgot to update the demo link, but I see the code looks good. 

Approved:fire:

PR #1790, opened 2023-06-20
Solution

julshvets:
[DEMO LINK](https://julshvets.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,121 @@
 import './App.scss';
+import { useState } from 'react';
+import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import { getCurrentTodos, getNewId } from './helpers';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { InitialTodo } from './types/Todo';
+import { TodoList } from './components/TodoList';
 
-export const App = () => {
+export const App: React.FC = () => {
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,129 @@
...
+          <label htmlFor="userSelector">
+            {'User: '}
+          </label>
+          <select
+            id="userSelector"
+            data-cy="userSelect"
+            onChange={((event) => {
+              setIsUserSelectError(false);
+              setSelectedUserId(+event.target.value);
+            })}
```
dffUqp, line 96: Move it to separate function

dffUqp (changes requested):

image

src/App.tsx
```diff
@@ -1,61 +1,137 @@
...
+import usersFromServer from './api/users';
+import { findUserById, getNewId } from './helpers';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { InitialTodo, Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+
+export const App: React.FC = () => {
+  const users = usersFromServer;
```
dinaminahi, line 11: `users` is the same reference as `usersFromServer` so we can use directly `usersFromServer`

dinaminahi (approved):
Well done 👍

PR #1796, opened 2023-06-21
...

victor-buldenko:
[DEMO LINK](https://victor-buldenko.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,108 @@
+import { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList/TodoList';
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [userName, setUserName] = useState('');
+  const [todoList, setTodoList] = useState([...todosFromServer]);
+  const [warningTitle, setWarningTitle] = useState(false);
+  const [warningUser, setWarningUser] = useState(false);
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/users" method="POST">
+      <form
+        action="/api/users"
+        method="POST"
+        // eslint-disable-next-line consistent-return
+        onSubmit={(e) => {
+          e.preventDefault();
+
+          let status = false;
+
+          if (!userName) {
+            status = true;
+
+            setWarningUser(true);
+          }
+
+          if (!title) {
+            status = true;
+
+            setWarningTitle(true);
+          }
+
+          if (status) {
+            return;
+          }
+
+          const user = [...usersFromServer].find(
+            (userItem) => userItem.name === userName,
+          );
+
+          if (user) {
+            const todo = {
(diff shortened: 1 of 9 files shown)


SerhiiKirik (changes requested):
Well done 🔥 
Let's fix some bugs 🧑‍💻 
I can add todo just with spaces in the title
image

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,12 @@
-export const TodoInfo = () => {};
+import React from 'react';
+import { UserInfo } from '../UserInfo';
+import { TodoInfoProps } from './type';
+
+export const TodoInfo:React.FC<TodoInfoProps> = React.memo(({ todo, user }) => {
+  return (
+    <article data-id={todo.id} className={`TodoInfo ${todo.completed && 'TodoInfo--completed'}`}>
```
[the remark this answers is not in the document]
victor-buldenko: Only 1 class, does not make sense.

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+      (userItem) => userItem.name === userName,
+    );
+
+    if (!user) {
+      throw new Error('Submit error, user not defined');
+    }
+
+    if (user) {
+      const todo = {
+        id: todoList.length,
```
polosanya, line 46: Change 'completed: true' to 'completed: false' since new todos should initially be uncompleted.

src/App.scss
```diff
@@ -17,23 +28,80 @@ button {
...
+label {
+  display: block;
+  margin-bottom: 4px;
+  color: rgba(0, 0, 0, 0.87);
+  font-size: 14px;
+  font-weight: 500;
+}
+
+input[type="text"],
+select {
```
polosanya, line 68: It is better to use class selectors instead of tags

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [userName, setUserName] = useState('');
+  const [todoList, setTodoList] = useState(todosFromServer);
+  const [isWarningTitle, setIsWarningTitle] = useState(false);
+  const [isWarningUser, setIsWarningUser] = useState(false);
+  const handleSubmit = (e: React.FormEvent) => {
+    e.preventDefault();
+
+    let status = false;
```
polosanya, line 17: It would be great to specify what exactly this status it and name it like boolean
image

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+          <label>
+            {'Title: '}
+            <input
+              type="text"
+              data-cy="titleInput"
+              placeholder="Enter a title"
+              value={title}
+              onChange={(event) => {
+                setTitle(event.target.value);
+                setIsWarningTitle(false);
```
polosanya, line 80: Create a separate function to handle changes. Fix all cases

polosanya (changes requested):
A few more changes to fix

src/App.scss
victor-buldenko: image
tags - its default style for this task.

sTorba24 (approved):
Cool!

PR #1797, opened 2023-06-21
1st version

Grechana:
https://Grechana.github.io/react_add-todo-form/

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,1 +1,14 @@
-export const TodoInfo = () => {};
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+export const TodoInfo:React.FC<{ todo: Todo }> = ({ todo }) => {
+  const userInfo = todo.user && <UserInfo user={todo.user} />;
+  const todoInfoClasses = `TodoInfo ${todo.completed ? 'TodoInfo--completed' : ''}`;
+
+  return (
+    <article className={todoInfoClasses}>
+      <h2 className="TodoInfo__title">{todo.title}</h2>
+      {userInfo}
+    </article>
+  );
+};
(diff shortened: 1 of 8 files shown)


src/types/Todo.ts
```diff
@@ -0,0 +1,9 @@
+import { User } from './User';
+
+export interface Todo {
+  id: number,
+  userId: number,
+  title: string,
+  completed: boolean,
+  user: User | null,
```
lerastarynets, line 8: It's better to use [an optional property](https://www.typescriptlang.org/docs/handbook/2/objects.html#optional-properties) instead of using User | null.

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,14 @@
-export const TodoInfo = () => {};
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+export const TodoInfo:React.FC<{ todo: Todo }> = ({ todo }) => {
+  const userInfo = todo.user && <UserInfo user={todo.user} />;
+  const todoInfoClasses = `TodoInfo ${todo.completed ? 'TodoInfo--completed' : ''}`;
```
lerastarynets, line 6: Really well done here) It's best practise to keep such values in separate consts

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,14 @@
-export const TodoInfo = () => {};
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+export const TodoInfo:React.FC<{ todo: Todo }> = ({ todo }) => {
+  const userInfo = todo.user && <UserInfo user={todo.user} />;
```
lerastarynets, line 5: Actually, you can place `todo.user && <UserInfo user={todo.user} />` directly inside jsx

lerastarynets (approved):
Left few comments on how to improve your code) You're doing a fantastic job, keep going!

PR #1799, opened 2023-06-25
added task solution

holytimy4:
[DEMO LINK](https://holytimy4.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,121 @@
+import React, { useState } from 'react';
+import { TodoList } from './components/TodoList/TodoList';
+import { Todo } from './Types/Todo';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { getUserById } from './Helpers/getUser';
+import { getNewTodoId } from './Helpers/getNewTodoId';
 
-export const App = () => {
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { User } from './Types/User';
+
+const todosWithUsers: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId, usersFromServer),
+}));
+
+export const App: React.FC = () => {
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,121 @@
...
+
+    const newTodo = {
+      id: getNewTodoId(todos),
+      userId: +userId,
+      title,
+      completed: false,
+      user: getUserById(+userId, usersFromServer),
+    };
+
+    setTodos((current => [...current, newTodo]));
```
dffUqp, line 48: Use more descriptive name.
```suggestion
    setTodos((prevTodos=> [...prevTodos, newTodo]));
```

src/App.tsx
```diff
@@ -1,61 +1,121 @@
...
+      if (!title) {
+        setIsTitleValid(false);
+      }
+
+      if (!userId) {
+        setIsUserSelected(false);
+      }
+
+      return;
+    }
```
dffUqp, line 38: let's refactor this condition to avoid using a nested if statement

```suggestion
      if (!title) {
        setIsTitleValid(false);
      }

      if (!userId) {
        setIsUserSelected(false);
      }

     if (!title || !userId) {
        return;
      }
```

src/App.tsx
```diff
@@ -1,61 +1,121 @@
...
+    };
+
+    setTodos((current => [...current, newTodo]));
+
+    setTitle('');
+    setUserId('');
+  };
+
+  const handleUserChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setUserId(event.target.value);
```
dffUqp, line 55: i think it will be better to format user id here

```suggestion
    setUserId(Number(event.target.value));
```

dffUqp (changes requested):
LGTM:fire: but needs a few fixes:upside_down_face:

image

silchyn (changes requested):
Form with a title consisting of spaces can be added

PR #1801, opened 2023-06-28
solution

KseniyaGonchar:
https://KseniyaGonchar.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,159 @@
+import { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+
+function getUser(userId: number | null) {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,154 @@
...
+      <form
+        action="/api/todos"
+        method="POST"
+        onSubmit={handleSubmit}
+      >
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <label htmlFor="taskTitle">
+            {'Title: '}
```
VitaliyBodnar, line 91: please don't add the spaces this way, you can use the margins and padding instead

src/App.tsx
```diff
@@ -1,61 +1,154 @@
+import { useState } from 'react';
 import './App.scss';
```
VitaliyBodnar, line 2: according to the style guide, import styles should be at the bottom of the imports list

VitaliyBodnar (changes requested):
Good Job🔥, let’s make a few changes.

IvanFesenko (approved):
Well done

PR #1808, opened 2023-07-04
react_add-todo-form

vintwp:
https://vintwp.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -6,21 +6,52 @@ import todosFromServer from './api/todos';
 import { NewTodo } from './components/NewTodo';
 import { TodoList } from './components/TodoList';
 
+import { Todo } from './types/Todo';
+import { User } from './types/User';
+
 import './App.scss';
 
-import { Todo } from './types/Todo';
+const users: User[] = [...usersFromServer];
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  // if there is no user with a given userId
+  return foundUser || null;
+}
+
+const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+type IndexedUsers = {
+  [index: number]: User;
+};
+
+const indexedUsers = users.reduce((
+  total: IndexedUsers, current: User,
+): IndexedUsers => {
+  // eslint-disable-next-line no-param-reassign
+  total[current.id] = { ...current };
+
+  return total;
+}, {});
 
 export const App = () => {
-  const [listTodo, setListTodo] = useState([...todosFromServer]);
+  const [todosToRender, setTodosToRender] = useState(todos);
 
   const addTodo = (todo: Omit<Todo, 'id'>) => {
-    const lastTodoIndex = listTodo.reduce((maxI, curI) => {
+    const lastTodoIndex = todosToRender.reduce((maxI, curI) => {
       return curI.id > maxI ? curI.id : maxI;
     }, 0);
 
-    setListTodo(currrentList => [...currrentList, {
+    const user = indexedUsers[todo.userId];
+
+    setTodosToRender(currentList => [...currentList, {
       ...todo,
       id: lastTodoIndex + 1,
+      user,
     }]);
   };
 
@@ -29,14 +60,12 @@ export const App = () => {
       <h1>Add todo form</h1>
 
       <NewTodo
-        users={usersFromServer}
-        onAdd={(todo) => {
-          addTodo((todo));
-        }}
+        users={users}
+        onAdd={addTodo}
       />
 
       <TodoList
-        todos={listTodo}
+        todos={todosToRender}
       />
     </div>
   );
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,43 @@
+import { useState } from 'react';
+
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { NewTodo } from './components/NewTodo';
+import { TodoList } from './components/TodoList';
+
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { Todo } from './types/Todo';
 
 export const App = () => {
+  const [listTodo, setListTodo] = useState([...todosFromServer]);
+
+  const addTodo = (todo: Omit<Todo, 'id'>) => {
+    const lastTodoIndex = listTodo.reduce((maxI, curI) => {
+      return curI.id > maxI ? curI.id : maxI;
+    }, 0);
+
+    setListTodo(currrentList => [...currrentList, {
+      ...todo,
+      id: lastTodoIndex + 1,
+    }]);
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/todos" method="POST">
-        <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
-        </div>
+      <NewTodo
+        users={usersFromServer}
+        onAdd={(todo) => {
+          addTodo((todo));
+        }}
+      />
 
-        <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>Choose a user</option>
-          </select>
-
-          <span className="error">Please choose a user</span>
-        </div>
-
-        <button type="submit" data-cy="submitButton">
-          Add
-        </button>
-      </form>
-
-      <section className="TodoList">
-        <article data-id="1" className="TodoInfo TodoInfo--completed">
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,61 +1,43 @@
...
-
-          <a className="UserInfo" href="mailto:Julianne.OConner@kory.org">
-            Patricia Lebsack
-          </a>
-        </article>
-      </section>
+      <NewTodo
+        users={usersFromServer}
+        onAdd={(todo) => {
+          addTodo((todo));
```
volodymyr-soltys97, line 34: You don't need to wrap 'todo' in extra parentheses. You can directly pass it to the 'addTodo' function.

src/App.tsx
```diff
@@ -1,61 +1,43 @@
...
-            quis ut nam facilis et officia qui
-          </h2>
-
-          <a className="UserInfo" href="mailto:Julianne.OConner@kory.org">
-            Patricia Lebsack
-          </a>
-        </article>
-      </section>
+      <NewTodo
+        users={usersFromServer}
```
volodymyr-soltys97, line 32: Directly passing 'usersFromServer' may result in performance issues and re-renders as the component will re-render every time 'usersFromServer' changes. Consider using a memoized version of the data.

src/App.tsx
```diff
@@ -1,61 +1,43 @@
...
+import { TodoList } from './components/TodoList';
+
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { Todo } from './types/Todo';
 
 export const App = () => {
+  const [listTodo, setListTodo] = useState([...todosFromServer]);
```
volodymyr-soltys97, line 14: It's not secure to directly use imported data in state. Consider copying data to the local variable first and then use it in the state.

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,48 @@
...
+    title,
+    userId,
+  },
+}) => {
+  // const user = usersFromServer.find(usr => usr.id === userId);
+  const user = findUser(userId, usersFromServer);
+
+  return (
+    <article
+      data-id={id}
```
volodymyr-soltys97, line 39: Using data-* attributes is not typically the React way of doing things. If the id is needed for later manipulations, it should be stored in the state.

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,48 @@
...
+import { UserInfo } from '../UserInfo';
+
+import { Todo } from '../../types/Todo';
+import { User } from '../../types/User';
+
+function findUser(userId: number, users: User[]): User {
+  const findedUser = users.find(user => user.id === userId);
+
+  return !findedUser
+    ? {
```
volodymyr-soltys97, line 13: Returning a hardcoded user object when no user is found might not be the best approach since it obscures the fact that a user was not found. It may be better to return null and handle this case appropriately in the rendering logic.

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,48 @@
-export const TodoInfo = () => {};
+import cn from 'classnames';
+import usersFromServer from '../../api/users';
+
+import { UserInfo } from '../UserInfo';
+
+import { Todo } from '../../types/Todo';
+import { User } from '../../types/User';
+
+function findUser(userId: number, users: User[]): User {
```
volodymyr-soltys97, line 9: The function findUser uses a potentially expensive operation (Array.prototype.find) every time a user needs to be found. This could be optimized by transforming usersFromServer into an object where user IDs are keys, which would make lookups more efficient.

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,48 @@
...
+
+export const TodoInfo: React.FC<Props> = ({
+  todo: {
+    id,
+    completed,
+    title,
+    userId,
+  },
+}) => {
+  // const user = usersFromServer.find(usr => usr.id === userId);
```
volodymyr-soltys97, line 34: The commented-out line of code should be removed to maintain code cleanliness.

src/components/NewTodo/NewTodo.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+export const NewTodo: React.FC<Props> = ({ users, onAdd }) => {
+  const [title, setTitle] = useState('');
+  const [titleHasError, setTitleHasError] = useState(false);
+
+  const [userSelect, setUserSelect] = useState(0);
+  const [userSelectHasError, setUserSelectHasError] = useState(false);
+
+  const isAllowedSymbols = (input: string): boolean => {
+    const pattern = /^[a-zA-Z А-ЩЬЮЯҐЄІЇа-щьюяґєії0-9]+$/;
```
volodymyr-soltys97, line 18: The regular expression used in the 'isAllowedSymbols' function only accepts Cyrillic and Latin characters. If the application is intended to support more languages, this could be a problem. Consider creating a more inclusive regular expression or handle this validation differently.

src/components/NewTodo/NewTodo.tsx
```diff
@@ -0,0 +1,115 @@
...
+  };
+
+  const clearTodoFields = () => {
+    setTitle('');
+    setUserSelect(0);
+  };
+
+  const handlerTitleInput = (event: React.ChangeEvent<HTMLInputElement>) => {
+    // setTitle(event.target.value);
+    setTitle(currentTitle => {
```
volodymyr-soltys97, line 34: You have commented out a line of code. If this code is not needed, it is best to remove it to keep the code clean and easier to read.

src/components/NewTodo/NewTodo.tsx
```diff
@@ -0,0 +1,115 @@
...
+type Props = {
+  users: User[];
+  onAdd: (todo: Omit<Todo, 'id'>) => void;
+};
+
+export const NewTodo: React.FC<Props> = ({ users, onAdd }) => {
+  const [title, setTitle] = useState('');
+  const [titleHasError, setTitleHasError] = useState(false);
+
+  const [userSelect, setUserSelect] = useState(0);
```
volodymyr-soltys97, line 14: The initial state of 'userSelect' is set to 0, which is not a valid user. Consider setting the initial state to null or undefined to avoid confusion or potential bugs.

volodymyr-soltys97 (changes requested):
Good job

src/App.tsx
```diff
@@ -1,61 +1,72 @@
...
+  total: IndexedUsers, current: User,
+): IndexedUsers => {
+  // eslint-disable-next-line no-param-reassign
+  total[current.id] = { ...current };
+
+  return total;
+}, {});
 
 export const App = () => {
+  const [todosToRender, setTodosToRender] = useState(todos);
```
anastasiiavorobiova, line 42: Consider adding generic type here

src/components/NewTodo/NewTodo.tsx
```diff
@@ -0,0 +1,119 @@
...
+import { User } from '../../types/User';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  users: User[];
+  onAdd: (todo: Omit<Todo, 'id'>) => void;
+};
+
+export const NewTodo: React.FC<Props> = ({ users, onAdd }) => {
+  const [title, setTitle] = useState('');
```
anastasiiavorobiova, line 11: Consider adding generic type here

src/components/NewTodo/NewTodo.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const [titleHasError, setTitleHasError] = useState(false);
```
anastasiiavorobiova, line 12: Consider adding generic type here

src/components/NewTodo/NewTodo.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  const [userSelect, setUserSelect] = useState<number | null>(null);
+  const [userSelectHasError, setUserSelectHasError] = useState(false);
```
anastasiiavorobiova, line 15: Consider adding generic type here

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,33 @@
...
+
+export const TodoInfo: React.FC<Props> = ({
+  todo: {
+    id,
+    completed,
+    title,
+    user,
+  },
+}) => {
+  const [todoId] = useState({ 'data-id': id });
```
anastasiiavorobiova, line 20: It's redundant you can set data attributes in the jsx

anastasiiavorobiova (changes requested):
Great solution! Added comments to fix

volodymyr-soltys97 (approved):

PR #1812, opened 2023-07-07
todo-add-form

Ir-ra:
[DEMO LINK](https://Ir-ra.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,41 @@
 import './App.scss';
+import { useState } from 'react';
+import { Todo } from './types/Todo';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { getUserById } from './services/user';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { TodoForm } from './components/TodoForm';
+
+export const initialTodos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
+
+function getNewTodotId(todos: Todo[]) {
+  const maxId = Math.max(...todos.map(todo => todo.id));
+
+  return maxId + 1;
+}
 
 export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>(initialTodos);
+  const addTodo = ({ id, ...data }: Todo) => {
+    const newTodo = {
+      id: getNewTodotId(todos),
+      ...data,
+    };
+
+    setTodos(currentTodo => [...currentTodo, newTodo]);
+  };
+
   return (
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,61 +1,41 @@
...
 
 export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>(initialTodos);
+  const addTodo = ({ id, ...data }: Todo) => {
+    const newTodo = {
+      id: getNewTodotId(todos),
+      ...data,
+    };
+
+    setTodos(currentTodo => [...currentTodo, newTodo]);
```
agniya-i, line 29: Typo in variable name 'currentTodo'. Should be 'currentTodos' as it refers to multiple todos, not a single one.

src/App.tsx
```diff
@@ -1,61 +1,41 @@
...
+
+function getNewTodotId(todos: Todo[]) {
+  const maxId = Math.max(...todos.map(todo => todo.id));
+
+  return maxId + 1;
+}
 
 export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>(initialTodos);
+  const addTodo = ({ id, ...data }: Todo) => {
```
agniya-i, line 23: You're destructuring 'id' from the argument but not using it in the function. If you don't need 'id', you can remove it from the argument destructuring.

agniya-i (changes requested):
I was able to add todo with an empty title by entering space 
Screenshot 2023-07-07 at 20 23 40

src/App.tsx
```diff
@@ -1,61 +1,45 @@
...
+export const initialTodos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
+
+// function getNewTodotId(todos: Todo[]) {
+//   const maxId = Math.max(...todos.map(todo => todo.id));
+
+//   return maxId + 1;
+// }
```
sTorba24, line 19: remove comments

src/services/user.ts
```diff
@@ -0,0 +1,9 @@
+import usersFromServer from '../api/users';
+import { User } from '../types/User';
+
+export function getUserById(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  // if there is no user with a given userId
```
sTorba24, line 7: remove

sTorba24 (approved):
Well done!

To fix comment about spaces was enough to do .trim() on validation
User should be able to add smth like `    text`, just trim it before add, it's a little not comfortable to search spaces so :) When user wrote text - he has smth to add :)

PR #1813, opened 2023-07-08
solution

Spec0s:
https://spec0s.github.io/react_add-todo-form/

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2090,9 +2090,9 @@
       }
     },
     "@mate-academy/scripts": {
-      "version": "1.2.3",
-      "resolved": "https://registry.npmjs.org/@mate-academy/scripts/-/scripts-1.2.3.tgz",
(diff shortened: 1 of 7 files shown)


witflash (changes requested):
Good work!
But could you implement checking for an empty input (just with spaces only)? You can use `trim` method to do that.
image

lerastarynets (changes requested):
I can still add user with whitespaces only. Make sure you fixed the issue and redeployed the project)

lerastarynets (changes requested):
Now you prohibited adding whitespace at all, it's also incorrect. Pls make use of `trim() `method to show error only after Add button is clicked. If trimmed string length = 0 show an error and don't add user, if trimmed string length > 0 add user)

Spec0s:
> Now you prohibited adding whitespace at all, it's also incorrect. Pls make use of `trim() `method to show error only after Add button is clicked. If trimmed string length = 0 show an error and don't add user, if trimmed string length > 0 add user)

oh yes my bad , weekend 😒

PR #1817, opened 2023-07-09
task done with title validation

pos1bl:
https://pos1bl.github.io/react_add-todo-form/

### DIFF
--- a/src/components/NewTodo/NewTodo.tsx
+++ b/src/components/NewTodo/NewTodo.tsx
@@ -1,0 +1,131 @@
+import { useState } from 'react';
+
+import users from '../../api/users';
+import { Todo } from '../../types/Todo';
+import { getUserById } from '../../services/user';
+
+const pattern = /^[a-zA-Zа-яА-ЯіІїЇєЄёЁґҐ0-9\s]+$/;
+
+type Props = {
+  onSubmit: (post: Todo) => void;
+};
+
+export const NewTodo: React.FC<Props> = ({ onSubmit }) => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+
+  const [titleErrorMessage, setTitleErrorMessage] = useState('');
+  const [hasUserIdError, setHasUserIdError] = useState(false);
+
+  const reset = () => {
+    setTitle('');
+    setUserId(0);
+  };
+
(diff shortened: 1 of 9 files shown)


src/types/Todo.ts
```diff
@@ -0,0 +1,9 @@
+import { User } from "./User";
+
+export type Todo = {
+  id: number,
+  title: string,
+  completed: boolean,
+  userId: number,
+  user: User | null,
+};
\ No newline at end of file
```
volodymyr-soltys97, line 9: Missing newline at the end of file. It's a general convention to end a file with a newline because some Unix utilities don't behave as expected when this newline is missing.

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,11 @@
-export const UserInfo = () => {};
+import { User } from '../../types/User';
+
+type Props = {
+  user: User | null,
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => (
+  <a className="UserInfo" href={`mailto:${user?.email}`}>
+    {user?.name}
```
volodymyr-soltys97, line 9: The code can potentially render null as a text inside 'a' tag if user is null. Here, it should be handled properly.

src/components/NewTodo/NewTodo.tsx
```diff
@@ -0,0 +1,131 @@
...
+          data-cy="userSelect"
+          value={userId}
+          onChange={handleUserIdChange}
+        >
+          <option value="0" disabled>
+            Choose a user
+          </option>
+
+          {users.map(user => (
+            <option value={user.id}>
```
volodymyr-soltys97, line 113: Each child in a list should have a unique 'key' prop. Add a key prop to the option element.

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,19 @@
...
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="TodoList">
+      {todos.map(todo => (
+        <TodoInfo
+          todo={todo}
+          data-id={todo.id}
+        />
```
volodymyr-soltys97, line 15: Need to add key

volodymyr-soltys97 (changes requested):
Good job 👍 

anastasiiavorobiova (approved):
Awesome!

PR #1821, opened 2023-07-10
Task done

AntonBaranskyi:
https://antonbaranskyi.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,104 @@
-import './App.scss';
+/*eslint-disable*/
+import "./App.scss";
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from "./api/users";
+import todosFromServer from "./api/todos";
+import { TodoList } from "./components/TodoList";
+import React, { useState } from "react";
+import { Todo } from "./types/ToDo";
+
+const posts = todosFromServer.map((todo) => {
+  const user = usersFromServer.find((person) => person.id === todo.userId);
+  return { ...todo, user };
+});
 
 export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>(posts);
+  const [titleField, setTitleField] = useState("");
+  const [userName, setUserName] = useState("");
+
+  const [errorTitle, setErrorTitle] = useState(false);
+  const [errorName, setErrorName] = useState(false);
+
+  const nextId = Math.max(...todos.map((todo) => todo.id)) + 1;
+
+  // const isDisabled = !!userName;
+
+  const addPost = (todo: Todo) => {
+    setTodos((prev) => [...prev, todo]);
+  };
+
+  const onChangeTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitleField(event.target.value);
+    setErrorTitle(false);
+  };
+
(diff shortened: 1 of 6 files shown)


cebanoleksandr (approved):
Good job:)

src/App.tsx
```diff
@@ -1,61 +1,112 @@
-import './App.scss';
+/*eslint-disable*/
```
qoa11a, line 1: Don't disable `eslint`. This tool should force a good code style. Without it, keeping your code clean and readable would be more problematic.

src/App.tsx
```diff
@@ -1,61 +1,112 @@
...
+  const [todos, setTodos] = useState<Todo[]>(posts);
+  const [titleField, setTitleField] = useState("");
+  const [userName, setUserName] = useState("");
+
+  const [errorTitle, setErrorTitle] = useState(false);
+  const [errorName, setErrorName] = useState(false);
+
+  const nextId = Math.max(...todos.map((todo) => todo.id)) + 1;
+
+  // const isDisabled = !!userName;
```
qoa11a, line 25: Delete comments that are redundant. They make code less readable.

src/App.tsx
```diff
@@ -1,61 +1,112 @@
...
 
 export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>(posts);
+  const [titleField, setTitleField] = useState("");
+  const [userName, setUserName] = useState("");
+
+  const [errorTitle, setErrorTitle] = useState(false);
+  const [errorName, setErrorName] = useState(false);
+
+  const nextId = Math.max(...todos.map((todo) => todo.id)) + 1;
```
qoa11a, line 23: You only use `nextId` inside the `sendPost` function. So it'd be better to move this calculated value there. At the moment on every component re-render you recalculate it which is bad for the performance.

qoa11a (changes requested):
Good job! 👍 
A few suggestions.

src/App.tsx
```diff
@@ -1,61 +1,112 @@
...
+    const objToAdd = {
+      id: +nextId,
+      title: titleField,
+      completed: false,
+      userId: nextId,
+      user: usersFromServer.find((user) => user.name === userName) || null,
+    };
+
+    if (!titleField) {
+      setErrorTitle(true);
```
anstsot, line 59: just add return here
```suggestion
      setErrorTitle(true);
      return;
```

src/App.tsx
```diff
@@ -1,61 +1,112 @@
...
+    setErrorName(false);
+    setErrorTitle(false);
+  };
+
+  const sendPost = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    const nextId = Math.max(...todos.map((todo) => todo.id)) + 1;
+
+    const objToAdd = {
```
anstsot, line 50: better to create `objToAdd` after all checks

anstsot (changes requested):
Good work, let's add few improvements
Please add labels to the inputs
image

src/App.tsx
AntonBaranskyi: But if do this user will see only one error after submitting the form with empty fields :(

PR #1823, opened 2023-07-10
finished the task

artkas03:
[DEMO LINK](https://artkas03.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,126 @@
 import './App.scss';
+import { useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { Todo } from './services/types';
+
+function getUserById(id: number) {
+  return usersFromServer.find(user => user.id === id);
+}
+
+const todosWithUser = (): Todo[] => {
+  return todosFromServer.map(todo => {
+    return {
+      ...todo,
+      user: getUserById(todo.id),
+    };
+  });
+};
+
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
 export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>(todosWithUser);
+
+  const [title, setTitle] = useState('');
+  const [isTitleValid, setIsTitleValid] = useState(true);
+
+  const [user, setUser] = useState(0);
+  const [isUserValid, setIsUserValid] = useState(true);
+
+  const [completed] = useState(false);
```
anstsot, line 33: unnecessary state

src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+    setUser(0);
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+    let abortSubmit = false;
+
+    if (!title) {
+      setIsTitleValid(false);
+      abortSubmit = true;
```
anstsot, line 46: just use `return;` instead of creating an additional variable
```suggestion
      return;
```

src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+          <select
+            name="user"
+            data-cy="userSelect"
+            onChange={event => handleUserOnChange(+event.currentTarget.value)}
+            value={user}
+          >
+            <option value="0">Choose a user</option>
+
+            {usersFromServer.map(({ id, name }) => (
+              <option value={id}>{name}</option>
```
anstsot, line 117: don't forget about `key`

anstsot (changes requested):
Good start:four_leaf_clover:
Let’s add a few fixes to improve your code.

src/App.tsx
artkas03: I used a variable to return only after checking both fields

And678 (approved):
Approve :cherries:

PR #1824, opened 2023-07-11
solution

vlkzmn:
[DEMO LINK](https://vlkzmn.github.io/react_add-todo-form/)


witflash (changes requested):
Good job!
Could you check the input using the `trim` method? Because I can create an empty title if I type spaces only
image

sTorba24 (approved):
Fine!

PR #1825, opened 2023-07-11
solution

yurii-krymskiy:
[DEMO LINK](https://yurii-krymskiy.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,41 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import todosFromServer from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+import { PostForm } from './components/PostForm/PostForm';
+import { getUser } from './services/user';
 
-export const App = () => {
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+function getNewTodoId(todoList: Todo[]) {
+  const maxId = Math.max(...todoList.map(todo => todo.id));
+
+  return maxId + 1;
+}
+
+export const App: React.FC = () => {
+  const [todosList, setTodosList] = useState<Todo[]>(todos);
+
+  const addTodo = ({ id, ...data }: Todo) => {
+    const newTodo = {
+      id: id || getNewTodoId(todosList),
+      ...data,
+    };
+
+    setTodosList(currentList => [...currentList, newTodo]);
+  };
+
   return (
     <div className="App">
-      <h1>Add todo form</h1>
+      <h1 className="App__title">Add todo form</h1>
 
-      <form action="/api/todos" method="POST">
-        <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
-        </div>
-
-        <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>Choose a user</option>
-          </select>
-
-          <span className="error">Please choose a user</span>
-        </div>
-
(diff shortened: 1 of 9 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,18 @@
...
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => (
+  <article
+    className={(todo.completed
```
volodymyr-soltys97, line 10: The className 'TodoInfo' is repeated in both conditions. Consider refactoring this to avoid repetition.

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,17 @@
...
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <section className="TodoList">
+    {todos.map(item => (
```
volodymyr-soltys97, line 10: It's a good practice to handle cases where the todos array could be null or undefined, to prevent your application from crashing.

src/App.tsx
```diff
@@ -1,61 +1,41 @@
...
+export const App: React.FC = () => {
+  const [todosList, setTodosList] = useState<Todo[]>(todos);
+
+  const addTodo = ({ id, ...data }: Todo) => {
+    const newTodo = {
+      id: id || getNewTodoId(todosList),
+      ...data,
+    };
+
+    setTodosList(currentList => [...currentList, newTodo]);
```
volodymyr-soltys97, line 30: It's good that you're using functional updates to add a new todo to your state. This prevents any issues from state updates being asynchronous.

volodymyr-soltys97 (changes requested):
Good job 👍 

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,17 @@
...
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <section className="TodoList">
+    {todos ? todos.map(item => (
+      <TodoInfo
+        key={item.id}
+        todo={item}
+      />
+    )) : null}
```
anastasiiavorobiova, line 15: Consider using a more descriptive placeholder than just 'null' when there are no todos. It can be a user-friendly message informing that there are no TODO items.

src/App.tsx
```diff
@@ -1,61 +1,41 @@
...
+
+  return maxId + 1;
+}
+
+export const App: React.FC = () => {
+  const [todosList, setTodosList] = useState<Todo[]>(todos);
+
+  const addTodo = ({ id, ...data }: Todo) => {
+    const newTodo = {
+      id: id || getNewTodoId(todosList),
```
anastasiiavorobiova, line 26: You are using `getNewTodoId(todosList)` to create a new id for the added Todo. This approach might lead to problems in case if ids are not contiguous. A safer approach could be using a unique id generator library like `uuid`.

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,14 @@
-export const UserInfo = () => {};
+import { User } from '../../types/User';
+
+type Props = {
+  user: User | null;
+};
+
+export const UserInfo:React.FC<Props> = ({ user }) => (
+  <a
```
anastasiiavorobiova, line 8: You are not checking if user exists before rendering the component, which means this component will render an empty anchor tag if no user is supplied. It would be better to conditionally render this component based on whether there is a user or not.

anastasiiavorobiova (approved):
Great work!

PR #1828, opened 2023-07-12
add task solution

IvanVaverchak:
[DEMO LINK](https://IvanVaverchak.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,110 @@
+import { useState } from 'react';
+
 import './App.scss';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { Todo } from './types/Todo';
 
-export const App = () => {
+export const App: React.FC = () => {
+  const [curentTodos, setCurentTodos] = useState(todosFromServer);
+
+  const [newTitle, setNewTitle] = useState('');
+  const [hasTitleError, setHasTitleError] = useState(false);
+
+  const [choosenUser, setChoosenUser] = useState(0);
+  const [hasUserError, setHasUserError] = useState(false);
+
+  const maxId = (arr: Todo[]): number => {
+    const max = Math.max(...arr.map(object => object.id));
+
+    return max + 1;
+  };
+
+  const reset = () => {
+    setNewTitle('');
+    setChoosenUser(0);
+    setHasTitleError(false);
+    setHasUserError(false);
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+    setHasTitleError(!newTitle);
+    setHasUserError(choosenUser === 0);
+
+    if (!newTitle || choosenUser === 0) {
+      return;
+    }
+
+    const newTodo: Todo = {
+      id: maxId(curentTodos),
+      title: newTitle,
+      completed: false,
+      userId: choosenUser,
+    };
+
+    setCurentTodos([...curentTodos, newTodo]);
+
+    reset();
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/todos" method="POST">
+      <form
+        action="/api/todos"
+        method="POST"
+        onSubmit={handleSubmit}
+      >
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
+
+  const maxId = (arr: Todo[]): number => {
+    const max = Math.max(...arr.map(object => object.id));
+
+    return max + 1;
+  };
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+  const reset = () => {
```
Andriy-Shymkiv, line 25: let's give more meaningful naming:
```
const resetState = () => {
  ...
}
```

src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
+export const App: React.FC = () => {
+  const [curentTodos, setCurentTodos] = useState(todosFromServer);
+
+  const [newTitle, setNewTitle] = useState('');
+  const [hasTitleError, setHasTitleError] = useState(false);
+
+  const [choosenUser, setChoosenUser] = useState(0);
+  const [hasUserError, setHasUserError] = useState(false);
+
+  const maxId = (arr: Todo[]): number => {
```
Andriy-Shymkiv, line 19: please, give another name for this function. I think `generateNewId` would be great

src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
+            value={choosenUser}
+            required
+            onChange={(event) => {
+              setChoosenUser(+event.target.value);
+              setHasUserError(false);
+            }}
+          >
             <option value="0" disabled>Choose a user</option>
+            {usersFromServer.map(user => (
+              <option value={user.id}>{user.name}</option>
```
Andriy-Shymkiv, line 94: you forgot to pass key prop for each mapping value

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,28 @@
...
+        'TodoInfo--completed': todo.completed,
+      })}
+    >
+      <h2 className="TodoInfo__title">
+        {todo.title}
+      </h2>
+
+      {users.map((user) => (user.id === todo.userId && (
+        <UserInfo user={user} key={user.id} />
+      )))}
```
Andriy-Shymkiv, line 25: it's not a good approach to find user like this. better use `find` method. here you can just:
```ts
const todoOwner: User | undefined = users.find((user) => user.id === todo.userId)
```

Andriy-Shymkiv: and then:
```ts
{!!todoOwner && <UserInfo user={todoOwner} key={todoOwner.id} />}
```

Andriy-Shymkiv: I didn't test it localy, but it should work

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,43 @@
...
+
+      <article data-id="2" className="TodoInfo">
+        <h2 className="TodoInfo__title">
+          quis ut nam facilis et officia qui
+        </h2>
+
+        <a className="UserInfo" href="mailto:Julianne.OConner@kory.org">
+          Patricia Lebsack
+        </a>
+      </article> */}
```
Andriy-Shymkiv, line 40: please remove comments

src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
+          <input
+            id="title-input"
+            type="text"
+            data-cy="titleInput"
+            value={newTitle}
+            placeholder="Enter a title"
+            onChange={(event) => {
+              setNewTitle(event.target.value);
+              setHasTitleError(false);
+            }}
```
Andriy-Shymkiv, line 73: If you want, you can create a separate function for onChange, and pass it here for better readability:
```ts
const onInputChange = (event) => {
  setNewTitle(event.target.value);
  setHasTitleError(false);
}
```

Andriy-Shymkiv: and pass it like this:
```ts
onChange={onInputChange}
```

src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
+          <label htmlFor="select-user">User: </label>
+          <select
+            id="select-user"
+            data-cy="userSelect"
+            value={choosenUser}
+            required
+            onChange={(event) => {
+              setChoosenUser(+event.target.value);
+              setHasUserError(false);
+            }}
```
Andriy-Shymkiv, line 90: here the same as above

Andriy-Shymkiv (changes requested):
good job! left comments

PR #1829, opened 2023-07-12
solution

DanyloPasko:
[DEMO LINK](https://DanyloPasko.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,136 @@
 import './App.scss';
+import React, { useState } from 'react';
+import usersFromServer from './api/users';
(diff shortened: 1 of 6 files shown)


vlad-demchuk (changes requested):
trim the string
I can add a todo with an empty one
image

anastasiiavorobiova (changes requested):
Great solution! 
Screenshot 2023-07-13 at 16 24 41
The previous comment is not resolved, still can add todo with an empty title

vlad-demchuk (approved):

PR #1835, opened 2023-07-14
add task solution

VolodymyrPst:
https://volodymyrpst.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,88 @@
+import React, { useState } from 'react';
 import './App.scss';
+import classNames from 'classnames';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
 export const App = () => {
+  const [userValue, setUserValue] = useState(0);
+  const [todoTitle, setTodoTitle] = useState('');
+  const [titleError, setTitleError] = useState(false);
+  const [userError, setUserError] = useState(false);
+
+  const idCreator = () => todosFromServer.length + 1;
+
+  const createToDo = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (!todoTitle) {
+      setTitleError(true);
+
+      return;
+    }
+
+    if (!userValue) {
+      setUserError(true);
+
+      return;
+    }
+
+    todosFromServer.push({
+      id: idCreator(),
+      title: todoTitle,
+      completed: false,
+      userId: userValue,
+    });
+
+    setUserValue(0);
+    setTodoTitle('');
+    setTitleError(false);
+    setUserError(false);
+  };
+
   return (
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,25 +1,88 @@
...
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
 export const App = () => {
+  const [userValue, setUserValue] = useState(0);
+  const [todoTitle, setTodoTitle] = useState('');
+  const [titleError, setTitleError] = useState(false);
+  const [userError, setUserError] = useState(false);
+
+  const idCreator = () => todosFromServer.length + 1;
```
anastasiiavorobiova, line 14: The way you create new ids for todos could potentially cause issues as it doesn't guarantee unique ids if a todo is removed. Consider using some library to generate unique ids, like uuid.

src/App.tsx
```diff
@@ -1,25 +1,88 @@
...
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <label htmlFor="title">Title:</label>
+          <input
+            type="text"
+            id="title"
+            value={todoTitle}
+            data-cy="titleInput"
+            placeholder="Enter a title"
+            onClick={() => setTitleError(false)}
```
anastasiiavorobiova, line 61: Avoid using onClick to clear the error. It's better to clear it onChange, as the user starts typing.

src/App.tsx
```diff
@@ -1,25 +1,88 @@
...
         </div>
 
         <div className="field">
-          <select data-cy="userSelect">
+          <label htmlFor="select">User:</label>
+          <select
+            data-cy="userSelect"
+            id="select"
+            value={userValue}
+            onClick={() => setUserError(false)}
```
anastasiiavorobiova, line 73: Avoid using onClick to clear the error. It's better to clear it onChange, as the user starts making a selection.

src/App.tsx
```diff
@@ -1,25 +1,88 @@
...
+            onChange={e => setUserValue(+e.target.value)}
+          >
             <option value="0" disabled>Choose a user</option>
+            {
+              usersFromServer.map((user) => {
```
anastasiiavorobiova, line 78: Each child in a list should have a unique 'key' prop.

src/App.tsx
```diff
@@ -1,25 +1,88 @@
...
+      return;
+    }
+
+    if (!userValue) {
+      setUserError(true);
+
+      return;
+    }
+
+    todosFromServer.push({
```
anastasiiavorobiova, line 31: Directly modifying the data from the server is not a good practice. Instead, maintain a separate state for the todos in the component.

anastasiiavorobiova (changes requested):
Great start! Keep doing!

volodymyr-soltys97 (approved):
Good job 👍

PR #1838, opened 2023-07-15
done

VladyslavKucher:
https://VladyslavKucher.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,133 @@
 import './App.scss';
+import { useState, useEffect } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { User } from './components/types/User';
+import { Todo } from './components/types/Todo';
+
+import { TodoList } from './components/TodoList/TodoList';
 
 export const App = () => {
+  const [copyTodos, setCopyTodos] = useState([...todosFromServer]);
+  const [selectedUserName, setSelectedUserName] = useState('');
+  const [newTitle, setNewTitle] = useState('');
+  const [errorTitle, setErrorTitle] = useState(false);
+  const [errorUser, setErrorUser] = useState(false);
+
+  const maxId = Math.max(...todosFromServer.map(todo => +todo.id));
+  const findUser = (userName: string): User | null => {
+    return usersFromServer.find(user => user.name === userName) || null;
+  };
+
+  const checkError = () => {
+    if (!newTitle) {
+      setErrorTitle(true);
+    }
+
+    if (!selectedUserName) {
+      setErrorUser(true);
+    }
+  };
+
+  const createTodo = () => {
+    const user = findUser(selectedUserName);
+
+    if (user && newTitle) {
+      const newTodo = {
+        id: maxId,
+        title: newTitle,
+        completed: false,
+        userId: user.id,
+      };
+
+      setCopyTodos(todos => [...todos, newTodo]);
+      setNewTitle('');
+      setSelectedUserName('');
+    }
+
+    return null;
+  };
+
+  const addHandler
+  = (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
+    event.preventDefault();
+    createTodo();
+    checkError();
+  };
+
+  const todos: Todo[] = copyTodos.map(todo => {
+    return {
+      ...todo,
+      user: usersFromServer.find(user => todo.userId === user.id),
+    };
+  });
+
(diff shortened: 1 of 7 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,26 @@
...
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    completed, title, user, id,
+  } = todo;
+
+  return (
+    <article data-id={id + 1} className={`TodoInfo ${completed ? 'TodoInfo--completed' : ''}`}>
```
lerastarynets, line 16: You are manipulating the `id` of the todo item by adding 1 to it. This may result in incorrect data-id attribute and can cause issues in future if relied upon. Consider using the original `id` value unless there is a specific reason for this manipulation.

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+      setSelectedUserName('');
+    }
+
+    return null;
+  };
+
+  const addHandler
+  = (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
+    event.preventDefault();
+    createTodo();
```
lerastarynets, line 55: The order of 'createTodo()' and 'checkError()' functions should be swapped. Validation should be done before creating a new todo.

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+
+  const todos: Todo[] = copyTodos.map(todo => {
+    return {
+      ...todo,
+      user: usersFromServer.find(user => todo.userId === user.id),
+    };
+  });
+
+  useEffect(() => {
+    if (newTitle) {
```
lerastarynets, line 67: The 'useEffect' hook is missing its dependency array. This could lead to unnecessary renders or stale data. It should include 'newTitle' and 'selectedUserName'.

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+    if (!newTitle) {
+      setErrorTitle(true);
+    }
+
+    if (!selectedUserName) {
+      setErrorUser(true);
+    }
+  };
+
+  const createTodo = () => {
```
lerastarynets, line 33: There is no need for 'return null' in the 'createTodo' function as it doesn't return any values.

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+import { TodoList } from './components/TodoList/TodoList';
 
 export const App = () => {
+  const [copyTodos, setCopyTodos] = useState([...todosFromServer]);
+  const [selectedUserName, setSelectedUserName] = useState('');
+  const [newTitle, setNewTitle] = useState('');
+  const [errorTitle, setErrorTitle] = useState(false);
+  const [errorUser, setErrorUser] = useState(false);
+
+  const maxId = Math.max(...todosFromServer.map(todo => +todo.id));
```
lerastarynets, line 18: `id` is numeric in `todosFromServer`
```suggestion
  const maxId = Math.max(...todosFromServer.map(todo => todo.id));
```

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
-
-          <span className="error">Please choose a user</span>
+          <label>
+            User:
+            <select
+              data-cy="userSelect"
+              onChange={(e) => setSelectedUserName(e.target.value)}
+            >
+              {!selectedUserName
+               && (<option value="0" disabled selected>Choose a user</option>)}
```
lerastarynets, line 106: The 'selected' attribute in the 'option' element is not recommended in React. Use the 'value' attribute in 'select' tag instead.

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)
Also now a user with title consisting of whitespaces only can be added, pls fix it

src/components/TodoInfo/TodoInfo.tsx
VladyslavKucher: id is the largest id in the array + 1 (add data-id={todo.id} attribute to each .TodoInfo).

src/App.tsx
```diff
@@ -1,61 +1,132 @@
...
+          </label>
+          {errorUser
+           && (<span className="error">Please choose a user</span>)}
         </div>
 
-        <button type="submit" data-cy="submitButton">
+        <button
+          type="submit"
+          data-cy="submitButton"
+          onClick={(e) => addHandler(e)}
```
agniya-i, line 123:
```suggestion
          onClick={(e) => handleAddTodo(e)}
```
Usually, we use the prefix `handle` for handler functions

agniya-i (approved):

PR #1842, opened 2023-07-17
React__add-todo-form

Soi4An:
DEMO - https://Soi4An.github.io/react_add-todo-form/

Test's problems:
  UserInfo
    1) should show a user.name
    2) should have a link with mailto: user.email
    3) should work for another user

In my opinion it is because i am transferring user Id in UserInfo but not whole user

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,112 @@
+import { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList/TodoList';
 
 export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>(todosFromServer);
+
+  const [title, setTitle] = useState('');
+  const [hasTitleErr, setHasTitleErr] = useState(false);
+  const [selectUserId, setSelectUserId] = useState(0);
+  const [hasSelectErr, setHasSelectErr] = useState(false);
+
+  const getBiggerId = (arrayItemsWithId: Todo[]) => {
+    const maxId = Math.max(...arrayItemsWithId.map(item => item.id));
+
+    return maxId + 1;
+  };
+
+  const handleTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const text = event.target.value;
+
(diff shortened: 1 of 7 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 

anastasiiavorobiova (changes requested):
Great work! The tiny comment to fix: use the trim function to avoid creating todos with spaces only
Screenshot 2023-07-17 at 18 24 53

src/App.tsx
```diff
@@ -1,61 +1,121 @@
...
+
+  const getBiggerId = (arrayItemsWithId: Todo[]) => {
+    const maxId = Math.max(...arrayItemsWithId.map(item => item.id));
+
+    return maxId + 1;
+  };
+
+  const handleTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const value = event.target.value;
+    const text = value.length === 1 ? value.trim() : value;
```
anastasiiavorobiova, line 26: It doesn't fix the problem, I still can add just spaces here. Consider adding validation error if there are just spaces

anastasiiavorobiova (changes requested):
Almost done! Great solution

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,23 @@
...
+export const TodoInfo: React.FC<Props> = ({
+  completed, title, todoId, userId,
+}) => (
+  <article
+    data-id={todoId}
+    className={`TodoInfo ${completed && 'TodoInfo--completed'}`}
+  >
+    <h2 className="TodoInfo__title">{title}</h2>
+
+    <UserInfo userId={userId} />
```
etojeDenys, line 21: better to just pass the user object and check here if this object exist

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,27 @@
...
+  const userDate: User | undefined = usersFromServer
+    .find((user: User) => user.id === userId);
+
+  let userEmail = '';
+  let userName = '';
+
+  if (userDate) {
+    userEmail = userDate.email;
+    userName = userDate.name;
+  }
```
etojeDenys, line 20: in this way we will get rid of unnecessary logic

etojeDenys (changes requested):

src/App.tsx
Soi4An: it is working, you maybe checked when change was not downloaded yet

volodymyr-soltys97 (approved):

PR #1852, opened 2023-07-18
Develop

AnastasiiaKhemych:
[DEMO LINK](https://AnastasiiaKhemych.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,1 +1,26 @@
-export const TodoInfo = () => {};
+import React from 'react';
+import cn from 'classnames';
+import { Todos } from '../../types/Todos';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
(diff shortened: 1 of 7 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,26 @@
...
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todos,
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  return (
+    <article
+      key={todo.id}
```
vladmakohon, line 13: It seems like this property is not needed here

vladmakohon (approved):
Nice solution!

src/components/TodoInfo/TodoInfo.tsx
anstsot, line 13: redundant
```suggestion
```

anstsot (approved):

PR #1859, opened 2023-07-18
Develop

olena-ole:
*Separated TextField and SelectField to be further reused

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,105 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { Todo, TodoWithUser, User } from './types';
+import { TodoList } from './components/TodoList';
+import { TextField } from './components/TextField';
+import { SelectField } from './components/SelectField';
 
-export const App = () => {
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,61 +1,105 @@
...
+    || null;
+
+  return {
+    ...todo,
+    user,
+  };
+});
+
+export const App: React.FC = () => {
+  const [count, setCount] = useState(0);
```
anstsot, line 22: why state variable `count` is needed?

src/App.tsx
```diff
@@ -1,61 +1,105 @@
...
-          <span className="error">Please enter a title</span>
-        </div>
-
-        <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>Choose a user</option>
-          </select>
+      <form
+        key={count}
+        onSubmit={(event) => handleSubmit(event)}
```
anstsot, line 75:
```suggestion
        onSubmit={handleSubmit}
```

anstsot (changes requested):
Good start:four_leaf_clover:
Let’s add a few fixes to improve your code.

src/App.tsx
olena-ole: it's needed to rerender the whole form and update the internal state of the TextField component (to default values). Without it, errors are shown after the form submit. I actually learned this from the previous task, or is it an antipattern?

anstsot (approved):

PR #1876, opened 2023-07-18
Solution

perfectAcedia:
[DEMO LINK](https://perfectAcedia.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,60 +1,115 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { ArrayOfTodos, Todos } from './types/Todos';
+
+function getUserById(userId: number) {
+  return usersFromServer.find(user => user.id === userId)
+    || null;
+}
+
+export const todosWithUser: ArrayOfTodos = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
 
 export const App = () => {
+  const [todos, setTodos] = useState<ArrayOfTodos>(todosWithUser);
+  const [title, setTitle] = useState('');
+  const [userId, setUserID] = useState(0);
+  const [titleError, setTitleError] = useState(false);
+  const [userError, setUserError] = useState(false);
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+    setTitleError(false);
+  };
+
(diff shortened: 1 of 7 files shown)


AlyonaSarapina (approved):
Good job! There are some small suggestions 👩‍🔧

src/App.tsx
```diff
@@ -1,60 +1,115 @@
...
-      <form action="/api/todos" method="POST">
+      <form
+        action="/api/todos"
+        method="POST"
+        onSubmit={handleAdd}
+      >
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          {'Title: '}
```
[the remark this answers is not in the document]
And678: Yes, @AlyonaSarapina is right

src/types/Todos.ts
```diff
@@ -0,0 +1,11 @@
+import User from './User';
+
+export interface Todos {
+  id: number,
+  title: string,
+  completed: boolean,
+  userId: number,
+  user?: User | null
```
And678, line 8: Use either `?` or `| null`.
```suggestion
  user: User | null
```

And678: Currently, user can be `User`, `null` and `undefined`, looks like you are using nulls in this case

src/App.tsx
```diff
@@ -1,60 +1,115 @@
...
+          <select
+            data-cy="userSelect"
+            required
+            value={userId}
+            onChange={handleUserIdChange}
+          >
             <option value="0" disabled>Choose a user</option>
+
+            {usersFromServer.map(user => (
+              <option value={user.id}>{user.name}</option>
```
And678, line 94: Where is `key`?)

src/App.tsx
```diff
@@ -1,60 +1,115 @@
...
+    setTitleError(false);
+  };
+
+  const handleUserIdChange = (event:
+  React.ChangeEvent<HTMLSelectElement>) => {
+    setUserID(+event.target.value);
+    setUserError(false);
+  };
+
+  const addTodos = (newTodo: Todos) => {
```
And678, line 37: Only one todo is added by this function
```suggestion
  const addTodo = (newTodo: Todos) => {
```

src/types/Todos.ts
```diff
@@ -0,0 +1,11 @@
...
+}
+
+export type ArrayOfTodos = Todos[];
```
And678, line 11: This type is not needed. `Todos[]` is a known syntax for every typescript dev. Using a custom type ArrayOfTodos is redundant.
```suggestion
```

And678 (changes requested):
GJ, but some fixes are needed

And678 (approved):
:+1: :+1:

PR #1880, opened 2023-07-18
add task solution

maksymPtashnyk:
[DEMO LINK](https://maksymPtashnyk.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,126 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Todo, User } from './components/interfaces/interfaces';
+
+function getUserById(id: number) {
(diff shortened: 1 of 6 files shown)


src/components/interfaces/interfaces.ts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+export interface User {
+  id: number;
+  name: string;
+  username: string;
+  email: string;
+}
+
+export interface Todo extends TodosWithoutUser {
```
And678, line 15: Nice usage of extends :fire:

src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+
+function getFilteredTitleInput(titleInput: string) {
+  return titleInput
+    .split('')
+    .filter((char: string | number) => (char >= 'a' && char <= 'z')
+      || (char >= 'A' && char <= 'Z')
+      || (char >= 'а' && char <= 'я')
+      || (char >= 'А' && char <= 'Я')
+      || (char >= '0' && char <= '9')
+      || char === ' ')
```
And678, line 34: Would be simpler to use a regex here

And678 (approved):
:+1: :+1:

PR #1886, opened 2023-07-19
Solution

kylyvnyk:
https://kylyvnyk.github.io/react_add-todo-form/


src/Types/Todo.ts
```diff
@@ -0,0 +1,9 @@
+import { User } from './User';
+
+export type Todo = {
```
anstsot, line 3: better to create interface
```suggestion
export interface Todo {
```

anstsot (approved):

PR #1890, opened 2023-07-19
react component to add todo form was added

Olena-Shevchenko:
[DEMO LINK](https://Olena-Shevchenko.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,1 +1,36 @@
-export const TodoInfo = () => {};
+import classNames from 'classnames';
+import { Todo } from '../../api/types/Todo';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todo,
+};
+
+export const TodoInfo: React.FC<Props> = ({
+  todo: {
+    id,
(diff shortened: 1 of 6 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,36 @@
...
+      )}
+    >
+      <h2 className="TodoInfo__title">
+        {title}
+      </h2>
+      {user
+        && (
+          <UserInfo
+            user={user}
+          />
```
polosanya, line 32:
```suggestion
          <UserInfo user={user} />
```
You can make it in one line if the amount of props is less than 3

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,20 @@
...
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="TodoList">
+      {todos.map(todo => (
+        <TodoInfo
+          key={todo.id}
+          todo={todo}
+        />
```
polosanya, line 16: Same here

polosanya (approved):
Nice!

PR #1893, opened 2023-07-21
solution

ShamievDavid:
[DEMO LINK](https://ShamievDavid.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,141 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { User } from './types/User';
+import { Todo } from './types';
+
+const getUserById = (userId: number) => {
+  return usersFromServer.find(user => user.id === userId)
+  || usersFromServer;
+};
+
+const preparedTodos = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId) as User,
+}));
 
 export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>(preparedTodos);
+  const [hasError, setHasError] = useState({ title: false, userId: false });
+  const [values, setValues] = useState({
+    title: '',
+    userId: 0,
+  });
+
+  const { title, userId } = values;
+
+  const createNewId = () => {
+    const allId = todos.map(todo => todo.id);
+
+    return Math.max(...allId) + 1;
+  };
+
+  const handleTitleInput = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setValues({ ...values, title: e.target.value });
+
+    setHasError({ ...hasError, title: false });
+  };
+
+  const handleUserIdSelect = (e: React.ChangeEvent<HTMLSelectElement>) => {
+    setValues({ ...values, userId: +e.target.value });
+
+    setHasError({ ...hasError, userId: false });
+  };
+
+  const reset = () => {
+    setHasError({ title: false, userId: false });
+
+    setValues({
+      title: '',
+      userId: 0,
+    });
+  };
+
+  const handleSubmit = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+
+    if (title === '') {
+      setHasError({ ...hasError, title: true });
+
+      return;
+    }
+
+    if (userId === 0) {
+      setHasError({ ...hasError, userId: true });
+
+      return;
+    }
+
+    const newTodo = {
+      id: createNewId(),
+      title,
+      userId,
+      completed: false,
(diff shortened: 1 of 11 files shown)


etojeDenys (changes requested):
we must show two errors immediately like in the working example
image

ShamievDavid:
> we must show two errors immediately like in the working example image

ShamievDavid:
it isa even easier. and it was my first version. but then i saw this in task description.
Screenshot 2023-07-22 at 10 52 25

src/App.tsx
```diff
@@ -1,61 +1,147 @@
...
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { User } from './types/User';
+import { Todo } from './types';
+
+const getUserById = (userId: number) => {
+  return usersFromServer.find(user => user.id === userId)
+  || usersFromServer;
+};
```
```suggestion
const getUserById = (userId: number) => {
  return usersFromServer.find(user => user.id === userId)
  || null;
};
```
better to return null instead of all array If the user wasn’t found

src/App.tsx
```diff
@@ -1,61 +1,147 @@
...
+
+const preparedTodos = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId) as User,
```
etojeDenys, line 17:
```suggestion
  user: getUserById(todo.userId),
```

src/App.tsx
```diff
@@ -1,61 +1,147 @@
...
+  };
+
+  const handleSubmit = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+
+    if (title === '' && userId === 0) {
+      setHasError({ title: true, userId: true });
+
+      return;
+    }
```
etojeDenys, line 64: looks redundant because we check all fields in the next if

src/App.tsx
```diff
@@ -1,61 +1,147 @@
...
+
+    const newTodo = {
+      id: createNewId(),
+      title,
+      userId,
+      completed: false,
+      user: getUserById(userId) as User,
```
etojeDenys, line 83:
```suggestion
      user: getUserById(userId),
```

better to avoid assertions as much as you can

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,27 @@
...
+  title,
+  id,
+  completed,
+  user,
+}) => (
+  <article
+    data-id={id}
+    className={completed
+      ? 'TodoInfo TodoInfo--completed'
+      : 'TodoInfo'}
```
etojeDenys, line 16: better to use classNames package for such complex classes

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,27 @@
...
+  >
+    <h2 className="TodoInfo__title">
+      {title}
+    </h2>
+
+    <UserInfo
+      name={user.name}
+      email={user.email}
+    />
```
etojeDenys, line 25: user can be null we need to handle this case here

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,145 @@
...
+import { User } from './types/User';
+
+const getUserById = (userId: number) => {
+  return usersFromServer.find(user => user.id === userId)
+  || null;
+};
+
+const preparedTodos = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId) as User,
```
IvanFesenko, line 17: not fixed
```suggestion
  user: getUserById(todo.userId),
```

src/types/Todo.ts
```diff
@@ -0,0 +1,9 @@
+import { User } from './User';
+
+export type Todo = {
+  id: number;
+  title: string,
+  completed: boolean,
+  userId?: number,
+  user?: User,
```
IvanFesenko, line 8:
```suggestion
  user: User | null,
```

src/types/Todo.ts
```diff
@@ -0,0 +1,9 @@
+import { User } from './User';
+
+export type Todo = {
+  id: number;
+  title: string,
+  completed: boolean,
+  userId?: number,
```
IvanFesenko, line 7:
```suggestion
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,32 @@
...
+      })}
+    >
+      <h2 className="TodoInfo__title">
+        {title}
+      </h2>
+
+      <UserInfo
+        name={name || 'No user'}
+        email={email || 'No email'}
+      />
```
IvanFesenko, line 29: Add check - if user exist than you show UserInfo component

IvanFesenko (changes requested):
Only spaces is not valid Todo title
2023-07-24_15h30_38

IvanFesenko (approved):

PR #1911, opened 2023-07-30
solution

Roman-Kulchytskyi:
https://roman-kulchytskyi.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,128 @@
-import './App.scss';
+import React, { FormEvent, useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import './App.scss';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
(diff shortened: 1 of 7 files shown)


etojeDenys (changes requested):
i can add the empty todo, let's fix that using [trim
](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/String/trim)
image

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/todos';
+import { User } from './types/users';
+
+const initialTodos = todosFromServer.map(todo => {
+  const user = usersFromServer.find(
+    userFromServer => userFromServer.id === todo.userId,
+  );
```
```suggestion
  ) || null;
```

etojeDenys: better to set null instead of undefined

etojeDenys (changes requested):

PR #1913, opened 2023-07-31
Solution

Viktoriia6666:


Viktoriia6666:
[DEMO LINK](https://Viktoriia6666.github.io/react_add-todo-form/)

src/App.tsx
```diff
@@ -1,60 +1,112 @@
...
+  ...todo,
+  user: getUserById(todo.userId),
+}));
 
 export const App = () => {
+  const [todos, setTodos] = useState<ArrayOfTodos>(todosWithUser);
+  const [title, setTitle] = useState('');
+  const [userId, setUserID] = useState(0);
+  const [titleError, setTitleError] = useState(false);
+  const [userError, setUserError] = useState(false);
```
loralevitska, line 24:
```suggestion
  const [isTitleError, setIsTitleError] = useState(false);
  const [isUserError, setIsUserError] = useState(false);
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,14 @@
...
+type Props = {
+  todos: ArrayOfTodos
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <>
+    {todos.map(todo => (
+      <TodoInfo todo={todo} key={todo.id} />
+    ))}
+  </>
```
loralevitska, line 13: Use semantic tags for list `ul` `li`. Don't forget to remove default `li` dots and `ul` paddings

loralevitska (changes requested):
Keep up the good work! Some comments to fix:

- fix adding empty comment, use [string.trim()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim)
image

PR #1924, opened 2023-08-14
Solution

AnnaViolin23:
[DEMO LINK](https://AnnaViolin23.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,126 @@
 import './App.scss';
+import { useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/Todo';
+
+function getUser(userId: number) {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser;
+}
+
+const initialTodos: Todo[]
+  = todosFromServer.map(
+    todo => ({
+      ...todo, user: getUser(todo.userId),
+    }),
+  );
 
 export const App = () => {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,136 @@
...
+
+  return foundUser;
+}
+
+const initialTodos: Todo[]
+  = todosFromServer.map(
+    todo => ({
+      ...todo, user: getUser(todo.userId),
+    }),
+  );
```
TarasHoliuk, line 20:
```suggestion
const initialTodos: Todo[] = todosFromServer.map(todo => ({
  ...todo,
  user: getUser(todo.userId),
}));
```
Will look a bit better

src/App.tsx
```diff
@@ -1,61 +1,136 @@
...
+    setHasUserError(false);
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    setHasTitleError(!title);
+    setHasUserError(userId === 0);
+
+    if (title && userId > 0) {
```
TarasHoliuk, line 66: you should check the trimmed value to avoid such behavior:
image

src/App.tsx
```diff
@@ -1,61 +1,136 @@
...
         </div>
 
         <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>Choose a user</option>
+          <label htmlFor="user">User: </label>
+          <select
+            data-cy="userSelect"
+            value={userId}
+            defaultValue="0"
```
TarasHoliuk, line 101: no need to set `defaultValue` if you set `value` (with correct initial value)

TarasHoliuk (changes requested):
GJ :+1: 
Let's improve it a bit)

TarasHoliuk (approved):
:+1:

PR #1925, opened 2023-08-14
Solution

RodionSav:


src/App.tsx
```diff
@@ -1,61 +1,123 @@
...
+            Title:
+            <input
+              name="title"
+              type="text"
+              placeholder="Enter a title"
+              data-cy="titleInput"
+              value={title}
+              onChange={handleTitleChange}
+            />
+            {hasTitle && <span className="error">Please enter a title</span>}
```
TarasHoliuk, line 89: sounds weird: if `hasTitle` we show `Please enter a title` message.
To make it clearer let's rename the variable to `hasTitleError`, etc.

src/App.tsx
```diff
@@ -1,61 +1,123 @@
...
+
+    const newTodo: Todo = {
+      id: getMaxTodoId() + 1,
+      title,
+      userId,
+      user: getUserById(userId),
+      completed: false,
+    };
+
+    setTodos([...todos, newTodo]);
```
TarasHoliuk, line 61: see info: https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state

TarasHoliuk (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,123 @@
...
+  };
+  //
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    setHasTitle(!title);
+    setHasUserId(!userId);
+
+    if (!title || !userId) {
```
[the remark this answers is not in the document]
etojeDenys: still not fixed
image


you can just use [trim](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/String/trim)

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,24 @@
...
+      data-id={todo.id}
+      className={cn('TodoInfo', {
+        'TodoInfo--completed': todo.completed,
+      })}
+    >
+      <h2 className="TodoInfo__title">
+        {todo.title}
+      </h2>
+
+      <UserInfo user={todo.user} />
```
etojeDenys, line 21: let's handle case when user is null here. There is no point in loading UserInfo with logic as to whether the user exists, it should simply output it

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,128 @@
...
+            User:
+            <select
+              name="select"
+              data-cy="userSelect"
+              value={userId}
+              onChange={handleUserIdChange}
+            >
+              <option value="0" disabled>Choose a user</option>
+              {usersFromServer.map(user => (
+                <option
```
volodymyr-soltys97, line 108: Use a destructuring for `user`

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,28 @@
...
+import cn from 'classnames';
+import { UserInfo } from '../UserInfo';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todo: Todo,
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  return (
```
volodymyr-soltys97, line 10: The same as above

volodymyr-soltys97 (changes requested):
Let's improve your code

anastasiiavorobiova (approved):
Great work!

PR #1927, opened 2023-08-15
add task solution

Luk2asz:
[DEMO LINK](https://Luk2asz.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -53,7 +53,6 @@ export const App: React.FC = () => {
 
         <div className="field">
           <label htmlFor="title">
-            {/* Title: */}
             {`Title: `}
             <input
               id="title"


src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,16 @@
-export const TodoList = () => {};
```
darokrk, line 1: The task requirement to create a form for adding new todos is not implemented in this file. You need to create a form with validated inputs for the user and the title of the todo. The form should add a new todo to the list when submitted.

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,11 @@
-export const UserInfo = () => {};
+import { User } from '../../types/User';
+
+type Props = {
+  user: User;
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => (
+  <a className="UserInfo" href={`mailto:${user.email}`}>
```
darokrk, line 8: This component doesn't seem to handle the case when the user object is null or undefined. This can result in errors if user.email or user.name is not defined.

please add `{user?.name}`

src/App.tsx
```diff
@@ -1,61 +1,112 @@
...
+            <span className="error">Please enter a title</span>
+          )}
         </div>
 
         <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>Choose a user</option>
-          </select>
+          <label htmlFor="title">
+            User:
```
darokrk, line 78: The label attribute 'htmlFor' should be set to 'user', not 'title'. Change htmlFor='title' to htmlFor='user'.

src/App.tsx
```diff
@@ -1,61 +1,112 @@
...
+        title,
+        completed: false,
+        userId: +userId,
+        user: getUser(+userId),
+      };
+
+      setUpdatedTodos([...updatedTodos, newTodo]);
+      setFormSubmitted(false);
+      setTitle('');
+      setuserId('');
```
darokrk, line 44: User ID should be set to null when form is cleared. Change setuserId('') to setuserId(null).

src/App.tsx
```diff
@@ -1,61 +1,112 @@
...
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [userId, setuserId] = useState('');
```
darokrk, line 25: userId should be rather 0 or null at the start and type as number ` useState<number>(0)`

src/App.tsx
```diff
@@ -1,61 +1,112 @@
...
+
+  const handleSubmit: FormEventHandler = (event) => {
+    event.preventDefault();
+
+    if (title && userId) {
+      const newTodo = {
+        id: Math.max(...updatedTodos.map(todo => todo.id)) + 1,
+        title,
+        completed: false,
+        userId: +userId,
```
darokrk, line 37: don't use that implicit coercion of types, please use `Number()` constructor, do that for the rest of code

src/App.tsx
```diff
@@ -1,61 +1,112 @@
...
+            {' '}
+            <select
+              data-cy="userSelect"
+              id="user"
+              name="user"
+              value={userId}
+              onChange={(event) => setuserId(event.target.value)}
+            >
+              <option value="" disabled>Choose a user</option>
+              {usersFromServer.map(user => (
```
darokrk, line 88: you can destruct your props here `{usersFromServer.map(({ id, name }) => (`

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,16 @@
...
+import { TodoInfo } from '../TodoInfo/TodoInfo';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <section className="TodoList">
+    {todos.map(todo => {
+      return (
```
darokrk, line 11: return word here is redundant just us `()` brackets

src/types/Todo.ts
```diff
@@ -0,0 +1,9 @@
+import { User } from './User';
+
+export interface Todo {
```
darokrk, line 3: you can move both types Todo and User into one file here called `index.ts` and just import types from here, you will have one file instead of two so it reduces codebase :)

darokrk (changes requested):
Looks good 👍🏽 , left mentions for code improvements 🛠️ 👨🏽‍💻

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,68 @@
-export const TodoList = () => {};
+import { useEffect, useState } from 'react';
+import { Todo, User } from '../../types/index';
+import { TodoInfo } from '../TodoInfo/TodoInfo';
+import usersFromServer from '../../api/users';
+
+export function getUser(userId: number): User | null {
```
mykhalenych, line 6: why did you duplicate this function?
export getUser from App

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,68 @@
...
+  buttonClicked,
+  titleChange,
+  userIdChange,
+  formSubmittedChange,
+  buttonClickedChange,
+}) => {
+  const [updatedTodos, setUpdatedTodos] = useState(todos);
+
+  useEffect(() => {
+    if (buttonClicked) {
```
mykhalenych, line 37: redundant state

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,68 @@
...
+      if (title && userId) {
```
mykhalenych, line 38: move this logic to submit

src/App.tsx
```diff
@@ -1,61 +1,121 @@
...
+  const [title, setTitle] = useState('');
+  const [userId, setuserId] = useState<number>(0);
+  const [formSubmitted, setFormSubmitted] = useState(false);
+  const [buttonClicked, setbuttonClicked] = useState(false);
+
+  const titleChange = (x: string) => {
+    setTitle(x);
+  };
+
+  const userIdChange = (x: number) => {
```
mykhalenych, line 33: the same

mykhalenych (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,111 @@
...
 
-      <form action="/api/todos" method="POST">
+      <form action="/api/todos" onSubmit={handleSubmit} method="POST">
+
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <label htmlFor="title">
+            Title:
+            {' '}
```
OleziO, line 56: Remove this line and use template strings.

[More info.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

OleziO (changes requested):

darokrk (approved):
Good job 🔥  approved 🏆

PR #1928, opened 2023-08-15
add todo form

Misha408:
[DEMO LINK](https://misha408.github.io/react_add-todo-form/)


volodymyr-soltys97 (changes requested):
Good job :
1. Add placeholder text here
image

2. The first name is repeated twice everywhere
image

etojeDenys (changes requested):
i can create the todo only with spaces let's fix this using `trim`

volodymyr-soltys97 (changes requested):
The add button does not need to be disabled, it should be active all the time, but if any field is not filled in, then there should be a message about it, as in the task description
image

nazarmatsevych (changes requested):
Good work but consider not to add todo without  user

image

nazarmatsevych (approved):
Nice 👍

PR #1934, opened 2023-08-17
Solution

s3rg4sh:
https://s3rg4sh.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,148 @@
+import { useState } from 'react';
 import './App.scss';
+import { TodoList } from './components/TodoList';
+import { ToDo } from './types';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+function getUser(id: number) {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,148 @@
...
+      { ...todo, title: event.target.value },
+    );
+  };
+
+  const handleUserIdChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setTodo(
+      {
+        ...todo,
+        userId: +event.target.value,
+        user: getUser(+event.target.value),
```
maxim2310, line 67: why do you need it, because you todo already has user with his id

src/App.tsx
```diff
@@ -1,61 +1,148 @@
...
+    event.preventDefault();
+
+    if (!todo.title || !todo.userId) {
+      setIsFilled(false);
+
+      return;
+    }
+
+    onAdd(todo);
+
```
maxim2310, line 82:
```suggestion
```

maxim2310 (changes requested):
Great job overall! Just couple tiny things needs to be fixed before approval)

maxim2310 (approved):
Well done!!!

PR #1935, opened 2023-08-17
solution

vitalii-fedusov:
[DEMO LINK](https://vitalii-fedusov.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,156 @@
 import './App.scss';
+import { useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+
+interface Todo {
+  id: number,
+  title: string,
+  completed: boolean,
+  userId: number,
+}
+
+interface User {
+  id: number,
+  name: string,
+  username: string,
+  email: string,
+}
+
+function getUserIDByName(arr: User[], name: string) {
+  const user = arr.find(item => item.name === name);
+
+  if (user !== undefined) {
+    return user.id;
+  }
+
+  return 1;
+}
+
+function getRandomId(): number {
+  return Math.floor(Math.random() * 100000);
+}
 
 export const App = () => {
+  const users: User[] = usersFromServer;
+  const [todos, setTodos] = useState<Todo[]>(todosFromServer);
+
+  const [title, setTitle] = useState('');
+  const [name, setName] = useState('0');
+
+  const newTodo = {
+    id: getRandomId(),
+    title,
+    completed: false,
+    userId: getUserIDByName(users, name),
+  };
+
+  const [titleError, setTitleError] = useState(false);
+  const [nameError, setNameError] = useState(false);
+
+  const handleTitleError = () => {
+    if (!title) {
+      setTitleError(true);
+    } else {
+      setTitleError(false);
+    }
+  };
+
+  const handleNameError = () => {
+    if (name === '0') {
+      setNameError(true);
+    } else {
+      setNameError(false);
+    }
+  };
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+  };
+
+  const handleNameChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setName(event.target.value);
+  };
+
+  const reset = () => {
+    setTitle('');
+    setName('0');
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+    handleTitleError();
+    handleNameError();
+
+    if (name !== '0' && title) {
+      reset();
+    }
+
+    if (!title || name === '0') {
+      return;
+    }
+
+    setTodos(currentTodos => [...currentTodos, newTodo]);
+  };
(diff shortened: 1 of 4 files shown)


etojeDenys (changes requested):
.

src/App.tsx
```diff
@@ -1,61 +1,164 @@
...
+
+function getUserIDByName(arr: User[], name: string) {
+  const user = arr.find(item => item.name === name);
+
+  if (user !== undefined) {
+    return user.id;
+  }
+
+  return 1;
+}
```
etojeDenys, line 30: let's just return the null if user does not exist

src/App.tsx
```diff
@@ -1,61 +1,164 @@
...
+    if (name === '0') {
+      setNameError(true);
+    } else {
+      setNameError(false);
+    }
+  };
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+  };
```
etojeDenys, line 71:
```suggestion
  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
    setTitle(event.target.value);
    setTitleError(false)
  };
```
if the user enters something, let's hide the error

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,165 @@
...
+  id: number,
+  name: string,
+  username: string,
+  email: string,
+}
+
+function getUserIDByName(arr: User[], name: string) {
+  const user = arr.find(item => item.name === name) || 1;
+
+  return user === 1 ? 1 : user.id;
```
etojeDenys, line 26:
```suggestion
  const user = arr.find(item => item.name === name)?.id || 1;
```

src/App.tsx
```diff
@@ -1,61 +1,165 @@
...
+
+  return Math.max(...arr) + 1;
+}
 
 export const App = () => {
+  const users: User[] = usersFromServer;
+  const [todos, setTodos] = useState<Todo[]>(todosFromServer);
+
+  const [title, setTitle] = useState('');
+  const [name, setName] = useState('0');
```
etojeDenys, line 40: why do we use here the zero
better to just set the empty string

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,51 @@
...
+        data-id={id}
+        className={cn('TodoInfo', {
+          'TodoInfo--completed': completed,
+        })}
+      >
+        <h2 className="TodoInfo__title">
+          {title}
+        </h2>
+
+        <UserInfo email={email} name={name} />
```
etojeDenys, line 47: user can be undefined, so let's handle this case here

because UserInfo doesn't need to know if the user exists or not, it just needs to output it

etojeDenys, line 47:
```suggestion
        {user && <UserInfo email={email} name={name} />}
```

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,165 @@
...
+    }
+  };
+
+  const handleNameError = () => {
+    if (name === '0') {
+      setNameError(true);
+    } else {
+      setNameError(false);
+    }
+  };
```
sTorba24, line 66:
```suggestion
  const handleNameError = () => {
    if (!name) {
      setNameError(true);
    } else {
      setNameError(false);
    }
  };
```

src/App.tsx
```diff
@@ -1,61 +1,165 @@
...
+    setTitle('');
+    setName('0');
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+    handleTitleError();
+    handleNameError();
+
+    if (name !== '0' && title.trim()) {
```
sTorba24, line 88: why do you use this 0? What a hidden sense to not use just empty string? ''?

src/App.tsx
```diff
@@ -1,61 +1,165 @@
...
+      reset();
+    }
+
+    if (!title.trim() || name === '0') {
+      return;
+    }
```
sTorba24, line 94: it looks so weird

src/components/TodoInfo/TodoInfo.tsx
sTorba24: image
still not fixed

sTorba24, line 47:
```suggestion
       {user && <UserInfo email={email} name={name} />}
```

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,12 @@
...
+  email?: string,
+};
+
+export const UserInfo: React.FC<Props> = ({ name, email }) => {
+  return (
+    <a className="UserInfo" href={`mailto:${email}`}>
+      {name}
+    </a>
+  );
+};
```
sTorba24, line 12:
```suggestion
export const UserInfo: React.FC<Props> = ({ name, email }) =>(
    <a className="UserInfo" href={`mailto:${email}`}>
      {name}
    </a>
  );
```

sTorba24 (changes requested):
Good work!

fix previous comments and some new :)

lerastarynets (approved):
You're doing amazing, keep it up!

PR #1938, opened 2023-08-20
Develop

RomanHrytsiuk:
https://RomanHrytsiuk.github.io/react_add-todo-form/


src/App.tsx
```diff
@@ -1,61 +1,153 @@
...
+
+          <select
+            value={userId}
+            data-cy="userSelect"
+            id="select-id"
+            onChange={handleSelect}
+          >
             <option value="0" disabled>Choose a user</option>
+
+            {usersFromServer.map(consumer => (
```
SanyaBratashchuk, line 133:
```suggestion
            {usersFromServer.map(user => (
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,14 @@
...
+
+type Props = {
+  todos: Todos[],
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="TodoList">
+      {todos.map(todo => <TodoInfo key={todo.id} todo={todo} />)}
+    </section>
```
SanyaBratashchuk, line 12: use semantic tags like `ul` and `li` for list and its items

src/types/users.ts
```diff
@@ -0,0 +1,7 @@
+export interface Users {
```
SanyaBratashchuk, line 1:
```suggestion
export interface User {
```

src/types/todos.ts
```diff
@@ -0,0 +1,9 @@
+import { Users } from './users';
+
+export interface Todos {
```
SanyaBratashchuk, line 3:
```suggestion
export interface Todo {
```

src/types/users.ts
```diff
@@ -0,0 +1,7 @@
...
+  id: number,
+  name: string,
+  username: string,
+  email: string,
+  // user?: Users,
```
SanyaBratashchuk, line 6: do not leave commented code

SanyaBratashchuk (changes requested):

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,26 @@
...
+import { Todo } from '../../types/todos';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todo
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  return (
+    <>
```
VitaliyBodnar, line 11: don't see any sense to use `<>` here, because you have only one tag inside

src/App.tsx
```diff
@@ -1,61 +1,153 @@
...
+            id="title-id"
+            placeholder="Enter a title"
+            type="text"
+            data-cy="titleInput"
+            onChange={handleTitleInput}
+          />
+
+          {titleError && (
+            <span className="error">
+              {!title.trim()
```
VitaliyBodnar, line 113: please don't call a function from the template because it will be executed every time when the state change, so make sense to move it to the variable

VitaliyBodnar (changes requested):
Almost done!

SanyaBratashchuk (approved):

PR #1939, opened 2023-08-20
Develop

Ilya-Osoben:
[DEMO LINK](https://Ilya-Osoben.github.io/react_add-todo-form/)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,22 @@
-export const TodoInfo = () => {};
+import cn from 'classnames';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
```
maxim2310, line 9: Would be great to use destructuring objects for better readability

maxim2310 (changes requested):
Almost there) Just one more thing needs to be fixed

PR #1948, opened 2023-08-25
done

yaroslav5481:
https://yaroslav5481.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,102 @@
 import './App.scss';
+import { useState } from 'react';
+import { getUserById } from './services/user';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    setHasTitleError(true);
+    setHasUserIdError(true);
+
+    if (!title || !userId) {
+      // eslint-disable-next-line no-useless-return
+      return;
+    }
```
sTorba24, line 31: set errors separately
image

now error is only on select user, but error appears for both inputs

sTorba24 (changes requested):
Well done! Lets fix little bug

Esceype (changes requested):
GJ!
Make sure that user can`t add todo only with spaces:
image

etojeDenys (approved):
lgtm

PR #1949, opened 2023-08-27
Develop

domkab:
https://domkab.github.io/react_add-todo-form

please ignore unused console statements for now,

also regarding my Users map in jsx, I should move it in order to get better readability or this can stay?

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,1 +1,27 @@
-export const TodoInfo = () => {};
+import React from 'react';
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+interface Props {
+  todo: Todo;
+}
+
+export const TodoInfo: React.FC<Props> = (
+  {
+    todo:
+      { completed, title, user },
+  },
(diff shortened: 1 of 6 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,27 @@
...
+import { UserInfo } from '../UserInfo';
+
+interface Props {
+  todo: Todo;
+}
+
+export const TodoInfo: React.FC<Props> = (
+  {
+    todo:
+      { completed, title, user },
```
witflash, line 13: I prefer to avoid nested object destructuring.
It's better to follow the approach:
```
export const TodoInfo = ({ todo }) => {
  const { completed, title, user } = todo;
```
It's more readable and more apparent.

witflash:
> https://domkab.github.io/react_add-todo-form
> 
> please ignore unused console statements for now,
> 
> also regarding my Users map in jsx, I should move it in order to get better readability or this can stay?

About `Users` map. Your approach is entirely well. This is a standard case in React JSX with a map.

And any `consoles` you definitely don't need in the final version :)

witflash (approved):
Good work 👍🏼
Pay attention to my comments for your future work.

PR #1955, opened 2023-08-30
solution

IvanOlecsandrovych:
[DEMO LINK](https://IvanOlecsandrovych.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,103 @@
+import React, { useState } from 'react';
 import './App.scss';
+import users from './api/users';
+import todos from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Todo } from './interfaces/todo';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+const newId = (list: Todo[]) => {
+  const maxId = Math.max(...list.map(todo => todo.id));
 
-export const App = () => {
+  return maxId + 1;
+};
+
+export const App: React.FC = () => {
+  const [todoList, setTodoList] = useState(todos);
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,111 @@
...
+            <select
+              data-cy="userSelect"
+              value={currentUserId}
+              onChange={onUserChange}
+            >
+              <option value="0" disabled>Choose a user</option>
+              {users.map(user => (
+                <option value={user.id}>
+                  {user.name}
+                </option>
```
IvanFesenko, line 94: Do not forget about key prop

src/App.tsx
```diff
@@ -1,61 +1,111 @@
...
+      setInvalidTitle(true);
+    }
+
+    if (currentUserId === '0') {
+      setInvalidUser(true);
+    }
+
+    if (!title.trim() || currentUserId === '0') {
+      return;
+    }
```
IvanFesenko, line 44: Better add return in each if
```suggestion
    if (!title.trim()) {
      setInvalidTitle(true);
            return;
    }

    if (currentUserId === '0') {
      setInvalidUser(true);
            return;
    }
```

IvanFesenko (changes requested):
Good job, check my comments

src/App.tsx
IvanFesenko: Add check into change method. Any case if you have questions - ask them in fe_chat

IvanFesenko (changes requested):

IvanFesenko (approved):
Nice

PR #1957, opened 2023-09-02
Solution

antonina-klishch:
https://antonina-klishch.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,128 @@
 import './App.scss';
+import React, { useState } from 'react';
+import { TodoList } from './components/TodoList';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { Todo } from './types/Todo';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+function getUserById(userId: number) {
+  return usersFromServer.find(user => user.id === userId) || null;
+}
+
+const startTodo = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
 
 export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>(startTodo);
+  const [title, setTitle] = useState('');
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,128 @@
...
+    event.preventDefault();
+    setHasTitleError(!title);
+    setHasUserIdError(!userId);
+
+    if (!title || !userId) {
+      return;
+    }
+
+    const newTodo = ({
+      user: usersFromServer.find(user1 => user1.id === userId) || null,
```
OleziO, line 65: Rename user1 to more clear name, like: `userFromServer`

OleziO (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,129 @@
...
+import todosFromServer from './api/todos';
+import { Todo } from './types/Todo';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+function getUserById(userId: number) {
+  return usersFromServer.find(user => user.id === userId) || null;
+}
+
+const startTodo = todosFromServer.map(todo => ({
```
etojeDenys, line 12:
```suggestion
const initialTodos = todosFromServer.map(todo => ({
```

etojeDenys (changes requested):
let's fix this case(with trim)
image

src/App.tsx
```diff
@@ -1,61 +1,129 @@
...
+    setHasTitleError(!title.trim());
+    setHasUserIdError(!userId);
+
+    if (!title.trim() || !userId) {
+      return;
+    }
+
+    const newTodo = ({
+      user: usersFromServer.find(userFromServer => userFromServer.id === userId)
+       || null,
```
VitaliyBodnar, line 66:
```suggestion
      user: getUserById(userId)
```

VitaliyBodnar (approved):
Good job!

PR #1960, opened 2023-09-04
Solved

silcki:
https://silcki.github.io/react_add-todo-form/


IvanFesenko (changes requested):
Good job. Check this case:
You should add an error in case invalid input value
https://github.com/mate-academy/react_add-todo-form/assets/61904995/790d9797-6672-471b-9205-dca3a3ee5418

silcki:
@IvanFesenko  fixed

PR #1965, opened 2023-09-05
Solution

oksana-vas:
[DEMO LINK](https://oksana-vas.github.io/react_add-todo-form/)


etojeDenys (changes requested):
i should see the error when I'm trying to add the new todo with spaces
image

PR #1967, opened 2023-09-06
Added logic for adding new task to todo list

davinnchii:
[DEMO LINK](https://davinnchii.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,130 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList/TodoList';
+import { Todo } from './types/todo';
+import { User } from './types';
+
+function getUserById(userId: number) {
+  return usersFromServer.find(user => user.id === userId)
+    || null;
+}
+
+function getMaxUserId() {
+  const userIds = usersFromServer.map(({ id }) => id);
+
+  return Math.max(...userIds);
+}
+
+export const createTodos = () => (todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+})));
 
 export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>(createTodos());
+  const [touched, setTouched] = useState({
+    title: false,
+    user: false,
+  });
+  const [newTitle, setNewTitle] = useState('');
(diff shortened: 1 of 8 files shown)


Koliras (changes requested):
Good job, but I think you should fix something

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,19 @@
...
+import { Todo } from '../../types/todo';
+
+type Props = {
+  todo: Todo
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => (
+  <article
+    className={classNames('TodoInfo', {
+      'TodoInfo--completed': todo.completed === true,
```
alexander-ignatow, line 12: @Koliras comment makes sense

```suggestion
      'TodoInfo--completed': todo.completed,

File: src/components/UserInfo/UserInfo.tsx [hunk 2]
```diff
@@ -1 +1,11 @@
...
+import { User } from '../../types';
+
+type Props = {
+  user: User | null,
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => (
+  <a className="UserInfo" href={`mailto:${user?.email}`}>
+    {user?.name}
+  </a>
```
alexander-ignatow, line 10: Even if user is null, you will render empty `a` tag - why not just return `null`?

src/App.tsx
```diff
@@ -1,61 +1,130 @@
...
+      setTouched({ title: false, user: true });
+
+      return;
+    }
+
+    if (!newTitle && !newUserId) {
+      setTouched({ title: true, user: true });
+
+      return;
+    }
```
alexander-ignatow, line 57: You can combine this like:
```suggestion
    if (!newTitle || !newUserId) {
      setTouched([set correct value here]);

      return;
    }
```

src/App.tsx
```diff
@@ -1,61 +1,130 @@
...
+          <label htmlFor="#title">
+            {'Title: '}
+            <input
+              id="title"
+              type="text"
+              data-cy="titleInput"
+              value={newTitle}
+              placeholder="Enter a title"
+              onChange={(event) => setNewTitle(event.target.value)}
+              onBlur={() => setTouched(prev => ({ ...prev, title: true }))}
```
alexander-ignatow, line 90: That looks bit not correct, as per task:
> errors should appear only after clicking the Add button;

alexander-ignatow (changes requested):
Good 👍 

Still, there are some things to fix

roman-mirzoian (approved):
Well done 👍

PR #1969, opened 2023-09-06
Implement adding todos to todolist

Koliras:
[DEMO LINK](https://Koliras.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/NewTodo/NewTodo.tsx
+++ b/src/components/NewTodo/NewTodo.tsx
@@ -1,0 +1,107 @@
+import React, { useState } from 'react';
+import { User } from '../../types/User';
+import { Todo } from '../../types/Todo';
+import todos from '../../api/todos';
+
+type Props = {
+  users: User[],
+  onSubmit: (todo: Todo) => void;
+};
+
+export const NewTodo: React.FC<Props> = ({ users, onSubmit }) => {
+  const [title, setTitle] = useState('');
+  const [selectedUserId, setSelectedUserId] = useState(0);
+  const [isTitleEmpty, setIsTitleEmpty] = useState(false);
+  const [isUserChosen, setIsUserChosen] = useState(true);
+  const maxId = Math.max(...todos.map(todo => todo.id));
+
+  const handleAddOfTodo = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (!title) {
+      setIsTitleEmpty(true);
+    }
+
+    if (!selectedUserId) {
+      setIsUserChosen(false);
+    }
+
+    if (!title || !selectedUserId) {
+      return;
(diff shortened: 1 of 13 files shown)


yuliiaant (approved):
Great job!

src/components/NewTodo/NewTodo.tsx
```diff
@@ -0,0 +1,107 @@
...
+type Props = {
+  users: User[],
+  onSubmit: (todo: Todo) => void;
+};
+
+export const NewTodo: React.FC<Props> = ({ users, onSubmit }) => {
+  const [title, setTitle] = useState('');
+  const [selectedUserId, setSelectedUserId] = useState(0);
+  const [isTitleEmpty, setIsTitleEmpty] = useState(false);
+  const [isUserChosen, setIsUserChosen] = useState(true);
```
alexander-ignatow, line 15: In your case, you may use a single state like `isError` and  calculate the values from other states

src/components/NewTodo/NewTodo.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const maxId = Math.max(...todos.map(todo => todo.id));
```
alexander-ignatow, line 16: You calculate this value inside the component but use it only inside the function `handleAddOfTodo`

It makes sense to calculate it inside `handleAddOfTodo` - only when it is needed

alexander-ignatow (changes requested):
Good job!

Still, some changes are needed

src/components/NewTodo/NewTodo.tsx
Koliras: The problem with calculation is the fact that in the beginning of rendering I need `isTitleEmpty` and `isUserChosen` to be the opposite of their true meaning, so that errors wouldn't be shown at first render

BudnikOleksii: > The problem with calculation is the fact that in the beginning of rendering I need `isTitleEmpty` and `isUserChosen` to be the opposite of their true meaning, so that errors wouldn't be shown at first render

It doesn't make sense when we have an empty string as `title` but at the same time `isTitleEmpty` is false 🤷‍♂️ 
The same as I see `selectedUserId` is 0 which means that no user was selected but at the same time `isUserChosen` is true.

If you want to show errors in different places at least you need provide not confusing names for your flags, maybe something like `isTitleValid`, because the title cannot be empty and not empty at the same time.

BudnikOleksii (changes requested):
Good work!
Let's fix a bit to avoid misunderstanding.

alexander-ignatow (approved):
LGTM 👍

PR #1975, opened 2023-09-06
[SOLUTION]

grabynsky:
[DEMO LINK](https://grabynsky.github.io/react_add-todo-form/)


etojeDenys (changes requested):
image
let's fix that

IvanFesenko (approved):
Well done

PR #1980, opened 2023-09-07
add task solution

ogospadzipalina:
[DEMO LINK](https://ogospadzipalina.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -26,7 +26,6 @@ export const App = () => {
   const [titleError, setTitleError] = useState('');
   const [userError, setUserError] = useState('');
   const [allTodos, setAllTodos] = useState(todos);
-  const [addButtonClicked, setAddButtonClicked] = useState(false);
 
   const handleNewTitle: React.ChangeEventHandler<HTMLInputElement>
   = (event) => {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,60 +1,126 @@
...
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState('0');
+  const [titleError, setTitleError] = useState('');
+  const [userError, setUserError] = useState('');
+  const [allTodos, setAllTodos] = useState(todos);
+  const [addButtonClicked, setAddButtonClicked] = useState(false);
```
klaudia-klis, line 29: I think this state [addButtonClicked, setAddButtonClicked] is unnecessary, you don't have to check if the button was clicked, because this function handleSubmit will run only after clicking on a button.

klaudia-klis (commented):
Good job! I suggested one small change :)

klaudia-klis (approved):

src/App.tsx
```diff
@@ -1,60 +1,126 @@
...
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState('0');
```
asakevych, line 25: i think sometimes id can also be `0`, so probably it's better to use `undefined` or `null` as a default value?

src/App.tsx
asakevych: totally agree with Klaudia 😉

asakevych (changes requested):
Great job!! 🚀 just some small changes needed 😉

src/App.tsx
ogospadzipalina: @asakevych it is 0 when nothing is selected, so I think it should be left like this as users start from 1

choeqq (approved):

PR #1986, opened 2023-09-07
add task solution

yar0man:
[DEMO LINK](https://yar0man.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,138 @@
 import './App.scss';
+import { useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+
+function getUserById(userId: number) {
+  return usersFromServer.find(user => user.id === userId)
+    || null;
+}
+
+function getMaxTodoId(todos: Todo[]) {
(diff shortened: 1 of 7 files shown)


Denys14Samoilenko (approved):
Good job!

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,31 @@
...
+  return (
+    <article
+      data-id={id}
+      className={cn('TodoInfo', {
+        'TodoInfo--completed': completed,
+      })}
+    >
+      <h2 className="TodoInfo__title">{title}</h2>
+
+      {user && (
```
roman-mirzoian, line 26: This check should be done at the beginning of the return, so that there is no option to display the title with an empty user info.

src/App.tsx
```diff
@@ -1,61 +1,138 @@
...
+      setHasTitleError(true);
+
+      return;
+    }
+
+    if (!userId) {
+      setHasUserIdError(true);
+
+      return;
+    }
```
roman-mirzoian, line 64: Try to write this condition a little shorter:
```suggestion
    if (!title || !userId) {
      if (!title) {
        setHasTitleError(true);
      }
     
      if (!userId) {
        setHasUserIdError(true);
      }

      return;
    }
```

roman-mirzoian (changes requested):

roman-mirzoian (approved):
Well done 👍

PR #1991, opened 2023-09-07
add task solution

Orpheus29:
[DEMO LINK](https://Orpheus29.github.io/react_add-todo-form/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,4 +25,4 @@ Implement the ability to add TODOs to the `TodoList` implemented in the **Static
 - Implement a solution following the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline).
 - Use the [React TypeScript cheat sheet](https://mate-academy.github.io/fe-program/js/extra/react-typescript).
 - Open one more terminal and run tests with `npm test` to ensure your solution is correct.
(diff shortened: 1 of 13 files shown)


ivannaromanovych (approved):
Great job!! 🔥🔥🔥

Moroz-Dmytro (changes requested):

src/helpers/checkTitleSpelling.tsx
```diff
@@ -0,0 +1,23 @@
...
+      || charCode === 1110
+      || charCode === 1111
+      || (charCode > 47 && charCode < 58)
+      || (charCode > 64 && charCode < 91)
+      || (charCode > 96 && charCode < 123)
+      || (charCode > 1025 && charCode < 1066)
+      || (charCode > 1069 && charCode < 1098)
+      || (charCode > 1101 && charCode < 1105)) {
+      preparedTitle += enteredTitle[i];
+    }
```
BudnikOleksii, line 19: I'm not sure why are you doing it, but if you want to validate and get rid of some characters you can use regular expressions.
Also, maybe it's not necessary to do it on every change, you can do it when the user submits.
And I think if some invalid characters appear you should notify the user that it's not valid as the user doesn't know why they type and nothing happens

PR #1995, opened 2023-09-07
added logic for todo add form

irahos:
[DEMO LINK](https://irahos.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/UserInfo/UserInfo.tsx
+++ b/src/components/UserInfo/UserInfo.tsx
@@ -1,1 +1,12 @@
-export const UserInfo = () => {};
+// import React from 'react';
+import { User } from '../../types';
+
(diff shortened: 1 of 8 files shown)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,12 @@
...
+
+interface UserInfoProps {
+  user: User | null;
+}
+
+export const UserInfo: React.FC<UserInfoProps> = ({ user }) => (
+  <a className="UserInfo" href={`mailto:${user?.email}`}>
+    {user?.name}
+  </a>
+);
```
alexander-ignatow, line 12: When the user is `null` your component will render empty `a` tag - which is not ok.

PR #2000, opened 2023-09-08
add initial solution

desevoker:
[DEMO LINK](https://desevoker.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/NewTodo/NewTodo.tsx
+++ b/src/components/NewTodo/NewTodo.tsx
@@ -1,0 +1,123 @@
+import React, { useState } from 'react';
+import { Todo } from '../../types';
+import { getUsers } from '../../services/apiService';
+import './NewTodo.scss';
+
+type Props = {
+  onAdd: (newTodo: Omit<Todo, 'id'>) => void,
+};
+
+export const NewTodo: React.FC<Props> = ({ onAdd }) => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [hasTitleError, setHasTitleError] = useState(false);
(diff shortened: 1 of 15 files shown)


NazarVotkanych (approved):
Good Job!
It's perfect that you implement an optional task.

src/types/Todo.ts
```diff
@@ -0,0 +1,12 @@
+import { User } from './User';
+
+export interface Todo {
+  id: number,
+  title: string,
+  completed: boolean,
+  userId: number,
+}
+
+export interface PreparedTodo extends Todo {
```
alexander-ignatow, line 10: Minor, but PreparedTodo is not perfect name here

Something like `AssignedTodo` or `TodoWithUser` would be better

src/types/Todo.ts
```diff
@@ -0,0 +1,12 @@
...
+  user: User | null;
```
alexander-ignatow, line 11: Once again - your solution works, but probably making `user` optional (instead of explicitly requiring `null` for "no user") would reduce the number of types and make your components slightly simpler

src/components/NewTodo/NewTodo.tsx
```diff
@@ -0,0 +1,123 @@
...
+    const newTitle = title.trim();
+
+    if (!newTitle || !userId) {
+      setHasTitleError(!newTitle);
+      setHasUserSelectError(!userId);
+
+      return;
+    }
+
+    const newTodo: Omit<Todo, 'id'> = {
```
alexander-ignatow, line 60: Good , that you have managed utility types 👍

alexander-ignatow (approved):
Very good!

Approving it, but please see some additional comments below

PR #2002, opened 2023-09-08
implement components: userInfo, todoInfo and todoList

serkrops:
[demo link](https://serkrops.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,123 @@
+import { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoWithUser } from './types/Todo';
+import { TodoList } from './components/TodoList';
+
+export const initialTodos = todosFromServer.map(todo => ({
+  ...todo,
+  user: usersFromServer.find(({ id }) => id === todo.userId) || null,
+}));
 
 export const App = () => {
+  const [todoTitle, setTodoTitle] = useState('');
+  const [hasTitleError, setHasTitleError] = useState(false);
+  const [userSelectId, setUserSelectId] = useState(0);
+  const [hasUserSelectedId, setHasUserSelectedId] = useState(false);
+  const [preparedTodos, setPreparedTodos] = useState(initialTodos);
+
+  function getUserById(userId: number) {
+    return usersFromServer.find(({ id }) => id === userId) || null;
+  }
+
+  function onSubmit(todo: TodoWithUser) {
+    setPreparedTodos(currentTodo => [...currentTodo, todo]);
+  }
+
+  function formReset() {
+    setTodoTitle('');
+    setUserSelectId(0);
+  }
+
+  function handleInputChange(event: React.ChangeEvent<HTMLInputElement>) {
+    setTodoTitle(event.target.value);
+    setHasTitleError(false);
+  }
+
+  function handleUserSelect(event: React.ChangeEvent<HTMLSelectElement>) {
+    setUserSelectId(+event.target.value);
+    setHasUserSelectedId(false);
+  }
+
+  const handleOnSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (!todoTitle || !userSelectId) {
+      setHasTitleError(Boolean(!todoTitle));
+      setHasUserSelectedId(Boolean(!userSelectId));
+
+      return;
+    }
+
+    const todoIds = preparedTodos.map(({ id }) => id);
+    const maxTodoId = Math.max(...todoIds) + 1;
+
+    onSubmit({
+      id: maxTodoId,
+      title: todoTitle,
+      completed: false,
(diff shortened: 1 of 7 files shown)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,14 @@
...
+
+type Props = {
+  user: User | null,
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  return (
+    <a className="UserInfo" href={`mailto:${user?.email}`}>
+      {user?.name}
+    </a>
```
alexander-ignatow, line 12: With your implementation you render an empty `a` tag when no user is provided - please fix this

src/types/Todo.ts
```diff
@@ -0,0 +1,16 @@
...
+  userId: number;
+}
+
+export interface TodoWithUser {
+  id: number;
+  title: string;
+  completed: boolean;
+  userId: number;
+  user: User | null;
+}
```
alexander-ignatow, line 16: Why do you need two types like this?

I would suggest you make the user optional
Even if you have some reasons to make it null rather than undefined, you should extend interfaces one from another (remember - DRY)

alexander-ignatow (changes requested):
Good job!

Now lets make small refactor - please create a separate component for `AddTodo` form (inputs + button), as your `App` looks overloaded

Also, there are some more small issues below

src/App.tsx
```diff
@@ -1,61 +1,96 @@
...
+  user: usersFromServer.find(({ id }) => id === todo.userId) || null,
+}));
 
 export const App = () => {
-  return (
-    <div className="App">
-      <h1>Add todo form</h1>
+  const [todoTitle, setTodoTitle] = useState('');
+  const [hasTitleError, setHasTitleError] = useState(false);
+  const [userSelectId, setUserSelectId] = useState(0);
```
tiserett, line 20: selectedUserId

src/App.tsx
```diff
@@ -1,61 +1,96 @@
...
+  const [hasUserSelectedId, setHasUserSelectedId] = useState(false);
```
tiserett, line 21: I believe it can be replaces with hasSelectedUserError, what do you think?

src/App.tsx
```diff
@@ -1,61 +1,96 @@
...
+  const [preparedTodos, setPreparedTodos] = useState(initialTodos);
```
tiserett, line 22: why preparedTodos, and not just todos?

src/App.tsx
```diff
@@ -1,61 +1,96 @@
...
+  const handleUserSelect = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setUserSelectId(+event.target.value);
+    setHasUserSelectedId(false);
+  };
 
-        <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>Choose a user</option>
-          </select>
+  const handleOnSubmit = (event: React.FormEvent) => {
```
tiserett, line 47: handleSubmit

src/App.tsx
```diff
@@ -1,61 +1,96 @@
...
+  const [hasTitleError, setHasTitleError] = useState(false);
+  const [userSelectId, setUserSelectId] = useState(0);
+  const [hasUserSelectedId, setHasUserSelectedId] = useState(false);
+  const [preparedTodos, setPreparedTodos] = useState(initialTodos);
+
+  function getUserById(userId: number) {
+    return usersFromServer.find(({ id }) => id === userId) || null;
+  };
+
+  function onSubmit(todo: TodoWithUser) {
```
tiserett, line 28: addTodo

src/App.tsx
```diff
@@ -1,61 +1,96 @@
...
 
-        <button type="submit" data-cy="submitButton">
-          Add
-        </button>
+      return;
+    }
+
+    const todoIds = preparedTodos.map(({ id }) => id);
+    const maxTodoId = Math.max(...todoIds) + 1;
+
```
tiserett, line 59: create function to receive maxId based on provided array

src/App.tsx
```diff
@@ -1,61 +1,96 @@
...
+  function getUserById(userId: number) {
+    return usersFromServer.find(({ id }) => id === userId) || null;
+  };
+
+  function onSubmit(todo: TodoWithUser) {
+    setPreparedTodos(currentTodo => [...currentTodo, todo]);
+  };
+
+  function formReset() {
```
tiserett, line 32: resetForm

src/components/TitleInput/TitleInput.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+type Props = {
+  todoTitle: string;
+  handleInputChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
+  hasTitleError: boolean;
+};
+
+export const TitleInput: React.FC<Props> = ({
```
tiserett, line 9: better rename it to just FormInput or TextField or something like this, with such approach you could easier reuse it

BudnikOleksii (approved):
Good work 👍

PR #2005, opened 2023-09-08
Develop

ivannaromanovych:


src/App.tsx
```diff
@@ -1,61 +1,37 @@
...
+
+import { TodoList } from './components/TodoList';
+import { TodoForm } from './components/TodoForm';
+import { Todo } from './types/Todo';
+
+function getNewTodoId(todos: Todo[]) {
+  const maxId = Math.max(...todos.map(({ id }) => id));
+
+  return maxId + 1;
+}
```
dffUqp, line 15: I think will be better to move it to the separate `helpers` file

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+      userId: !inputs.userId,
+    }));
+
+    if (!inputs.title || !inputs.userId) {
+      return;
+    }
+
+    onAdd({
+      ...inputs,
+      id: 0,
```
dffUqp, line 59: Let's come up with the better implementation for this

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+    setInputs(prevFields => ({
+      ...prevFields,
+      [name]: name === 'userId' ? +value : value,
+    }));
+    setInputsErrors(prevErrors => ({
+      ...prevErrors,
+      [name]: !(name === 'userId' ? +value : value),
+    }));
+  };
```
dffUqp, line 39: It's hard to read, I think it would be better to create two separate handlers.

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  const reset = () => {
```
dffUqp, line 41:
```suggestion
  const resetForm = () => {
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,33 @@
...
+import { User } from '../../types/User';
+import { Todo } from '../../types/Todo';
+
+interface Props {
+  todo: Todo
+}
+
+function getUserById(id: number): User | null {
+  return usersFromServer.find((user: User) => user.id === id) || null;
+}
```
dffUqp, line 15: same about `helpers` file

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,14 @@
...
+interface Props {
+  user: User | null,
+}
+
+export const UserInfo: React.FC<Props> = ({
+  user,
+}) => (
+  <a className="UserInfo" href={`mailto:${user?.email}`}>
+    {user?.name}
+  </a>
```
dffUqp, line 13: All user values possibly `null`.
Can we show user only if we have it?

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,33 @@
...
+    className={cn(
+      'TodoInfo',
+      { 'TodoInfo--completed': todo.completed },
+    )}
+  >
+    <h2 className="TodoInfo__title">
+      {todo.title}
+    </h2>
+
+    <UserInfo user={getUserById(todo.userId)} />
```
dffUqp, line 31: It is not a good practice to use function inside JSX, let's create a variable for it

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+}
+const DEFAULT_INPUTS = {
+  title: '',
+  userId: 0,
+};
+
+const DEFAULT_INPUTS_ERRORS = {
+  title: false,
+  userId: false,
+};
```
dffUqp, line 17:
```suggestion
const DEFAULT_FORM_VALUES = {
  title: '',
  userId: 0,
};

const DEFAULT_FORM_ERRORS = {
  title: false,
  userId: false,
};

interface Props {
  onAdd: (newTodo: Todo) => void;
  users: User[],
}
```

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+    }));
+  };
+
+  const reset = () => {
+    setInputs(DEFAULT_INPUTS);
+    setInputsErrors(DEFAULT_INPUTS_ERRORS);
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event?.preventDefault();
```
dffUqp, line 47:
```suggestion
    event.preventDefault();
```

dffUqp (changes requested):
LGTM:fire: but needs a few minor fixes:smiley:

src/App.tsx
```diff
@@ -1,61 +1,36 @@
...
-        </article>
+export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>(todosFromServer);
 
-        <article data-id="15" className="TodoInfo TodoInfo--completed">
-          <h2 className="TodoInfo__title">delectus aut autem</h2>
+  const handleAddTodo = (todoInfo: {
+    title: string,
+    userId: number,
+  }) => {
```
BudnikOleksii, line 18: You can use some of the utility types, for example, `partial`, `pick`, or `omit`.
https://www.typescriptlang.org/docs/handbook/utility-types.html

`Pick<Todo, title | userId>`

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,129 @@
...
+const DEFAULT_INPUTS_ERRORS = {
+  title: false,
+  userId: false,
+};
+
+interface Props {
+  onAdd: (newTodoInfo: {
+    title: string,
+    userId: number,
+  }) => void;
```
BudnikOleksii, line 18: Should be the same type as for the `handleAddTodo`, so it's a good idea to create and use it. As I mentioned above

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,129 @@
...
+  users: User[],
+}
+
+export const TodoForm: React.FC<Props> = ({
+  onAdd = () => { },
```
BudnikOleksii, line 23: `onAdd` is a required prop, you don't need to assign default value

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,129 @@
...
+      title: prevInputs.title.trim(),
+    }));
+
+    if (!inputs.title.trim() || !inputs.userId) {
+      return;
+    }
+
+    onAdd({
+      ...inputs,
+    });
```
BudnikOleksii, line 70:
```suggestion
    onAdd(inputs);
```
You can just pass the object directly

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,18 @@
-export const UserInfo = () => {};
+import React from 'react';
+import { User } from '../../types/User';
+
+interface Props {
+  user: User | null,
+}
```
BudnikOleksii, line 6:
```suggestion
interface Props {
  user: User,
}
```
Don't render the `UserInfo` component if no user is present

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,18 @@
...
+
+export const UserInfo: React.FC<Props> = ({
+  user,
+}) => (
+  <>
+    {user && (
+      <a className="UserInfo" href={`mailto:${user?.email}`}>
+        {user?.name}
+      </a>
+    )}
```
BudnikOleksii, line 16: Do this check in the `TodoInfo` component

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,32 @@
...
+      className={cn(
+        'TodoInfo',
+        { 'TodoInfo--completed': todo.completed },
+      )}
+    >
+      <h2 className="TodoInfo__title">
+        {todo.title}
+      </h2>
+
+      <UserInfo user={user} />
```
BudnikOleksii, line 29:
```suggestion
      {user && <UserInfo user={user} />}
```

BudnikOleksii (changes requested):
Great work 👍 
But can be even better, check the comments.
Also, the preview link is broken

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,125 @@
...
+          value={inputs.userId}
+          onChange={handleInputChange}
+          name="userId"
+        >
+          <option value="0" defaultChecked>Choose a user</option>
+          {users.map(user => (
+            <option value={user.id} key={user.id}>
+              {user.name}
+            </option>
+          ))}
```
roman-mirzoian, line 112:
```suggestion
          {users.map(({ id, name })=> (
            <option value={id} key={id}>
              {name}
            </option>
          ))}
```

roman-mirzoian (approved):
Well done 👍

PR #2008, opened 2023-09-08
implement addTodoForm functionality

kaavka:
[DEMO LINK](https://kaavka.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,35 @@
+import { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { Todo, TodoWithUser } from './types/Todo';
+import { TodoList } from './components/TodoList/TodoList';
+import { FormAddTodo } from './components/FormAddTodo/FormAddTodo';
+import { findUserById } from './utils/findUserById';
+
+function preparedTodos(todos: Todo[]) {
+  return todos.map((todo: Todo) => ({
+    ...todo,
+    user: findUserById(usersFromServer, todo.userId),
+  }));
+}
 
 export const App = () => {
+  const [todos, setTodos] = useState(preparedTodos(todosFromServer));
+
+  const onSubmit = (todo: TodoWithUser) => {
+    setTodos((prevState) => [...prevState, todo]);
+  };
+
   return (
     <div className="App">
-      <h1>Add todo form</h1>
-
-      <form action="/api/todos" method="POST">
-        <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
-        </div>
-
-        <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>Choose a user</option>
-          </select>
-
-          <span className="error">Please choose a user</span>
-        </div>
-
-        <button type="submit" data-cy="submitButton">
-          Add
-        </button>
-      </form>
-
-      <section className="TodoList">
-        <article data-id="1" className="TodoInfo TodoInfo--completed">
-          <h2 className="TodoInfo__title">
-            delectus aut autem
-          </h2>
-
-          <a className="UserInfo" href="mailto:Sincere@april.biz">
-            Leanne Graham
-          </a>
-        </article>
-
-        <article data-id="15" className="TodoInfo TodoInfo--completed">
-          <h2 className="TodoInfo__title">delectus aut autem</h2>
-
-          <a className="UserInfo" href="mailto:Sincere@april.biz">
-            Leanne Graham
-          </a>
-        </article>
-
-        <article data-id="2" className="TodoInfo">
-          <h2 className="TodoInfo__title">
-            quis ut nam facilis et officia qui
-          </h2>
-
-          <a className="UserInfo" href="mailto:Julianne.OConner@kory.org">
-            Patricia Lebsack
-          </a>
-        </article>
-      </section>
+      <FormAddTodo
+        users={usersFromServer}
+        onSubmit={onSubmit}
+        todos={todos}
+      />
+      <TodoList todos={todos} />
     </div>
   );
 };
--- a/src/components/FormAddTodo/FormAddTodo.tsx
+++ b/src/components/FormAddTodo/FormAddTodo.tsx
@@ -1,0 +1,114 @@
+import { useState } from 'react';
+import { TodoWithUser, User } from '../../types';
+import { findUserById } from '../../utils/findUserById';
+import {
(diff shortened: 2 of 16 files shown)


voronine (approved):
Сool solution!

src/App.tsx
```diff
@@ -1,61 +1,35 @@
...
-          </select>
-
-          <span className="error">Please choose a user</span>
-        </div>
-
-        <button type="submit" data-cy="submitButton">
-          Add
-        </button>
-      </form>
+function preparedTodos(todos: Todo[]) {
```
Moroz-Dmytro, line 11: Better to make this constant, not function

src/App.tsx
```diff
@@ -1,61 +1,35 @@
...
-          </a>
-        </article>
+export const App = () => {
+  const [todos, setTodos] = useState(preparedTodos(todosFromServer));
 
-        <article data-id="2" className="TodoInfo">
-          <h2 className="TodoInfo__title">
-            quis ut nam facilis et officia qui
-          </h2>
+  const onSubmit = (todo: TodoWithUser) => {
```
Moroz-Dmytro, line 21:
```suggestion
  const handleSubmit = (todo: TodoWithUser) => {
```

src/App.tsx
```diff
@@ -1,61 +1,35 @@
...
 
-          <a className="UserInfo" href="mailto:Julianne.OConner@kory.org">
-            Patricia Lebsack
-          </a>
-        </article>
-      </section>
+  return (
+    <div className="App">
+      <FormAddTodo
+        users={usersFromServer}
```
Moroz-Dmytro, line 28: Can we just import this `usersFromServer` in this component instead of passing it through props?

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,32 @@
...
+      data-id={id}
+      className={classNames('TodoInfo', {
+        'TodoInfo--completed': completed,
+      })}
+    >
+      <h2 className="TodoInfo__title">{title}</h2>
+
+      {user
+        ? <UserInfo user={user} />
+        : <p> No user found</p>}
```
Moroz-Dmytro, line 29:
```suggestion
      {user
        ? <UserInfo user={user} />
        : <p> No user found</p>
      }
```

src/utils/generateUniqueId.ts
```diff
@@ -0,0 +1,5 @@
+import { TodoWithUser } from '../types';
+
+export function generateUniqueId(todos: TodoWithUser[]): number {
+  return Math.max(...todos.map(todo => todo.id)) + 1;
```
Moroz-Dmytro, line 4:
```suggestion
  return Math.max(...todos.map(({ id })=> id)) + 1;
```

Moroz-Dmytro (changes requested):

src/components/FormAddTodo/FormAddTodo.tsx
```diff
@@ -0,0 +1,117 @@
...
+            onChange={handleUserSelect}
+          >
+            <option value="0" disabled>
+              Choose a user
+            </option>
+            {users.map(user => (
+              <option value={user.id} key={user.id}>
+                {user.name}
+              </option>
+            ))}
```
roman-mirzoian, line 103: Use destructuring on the object.

src/components/TodoInfo/TodoInfo.tsx
roman-mirzoian: Not fixed 😢

src/components/FormAddTodo/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const ERROR_MESSAGE_FOR_INVALID_TITLE = 'Please enter a title';
+export const ERROR_MESSAGE_FOR_INVALID_USER = 'Please choose a user';
```
roman-mirzoian, line 2:
```suggestion
export const INVALID_TITLE_ERROR_MESSAGE = 'Please enter a title';
export const INVALID_USER_ERROR_MESSAGE = 'Please choose a user';
```

src/utils/findUserById.ts
```diff
@@ -0,0 +1,5 @@
+import { User } from '../types';
+
+export function findUserById(users: User[], id: number) {
+  return users.find(user => user.id === id) || null;
+}
```
roman-mirzoian, line 5:
```suggestion
export function findUserById(users: User[], userId: number) {
  return users.find(({ id }) => id === userId) || null;
}
```

roman-mirzoian (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,34 @@
...
+import { TodoList } from './components/TodoList/TodoList';
+import { FormAddTodo } from './components/FormAddTodo/FormAddTodo';
+import { findUserById } from './utils/findUserById';
+
+const preparedTodos = (todos: Todo[]) => {
+  return todos.map((todo: Todo) => ({
+    ...todo,
+    user: findUserById(usersFromServer, todo.userId),
+  }));
+};
```
alexander-ignatow, line 16: Please consider @Moroz-Dmytro comment
At the very least you need to rename this function, as current name is confusing

src/components/FormAddTodo/FormAddTodo.tsx
```diff
@@ -0,0 +1,121 @@
...
+  function resetForm() {
+    setSelectedUserId(0);
+    setTitle('');
+  }
+
+  function handleSubmit(event: React.FormEvent<HTMLFormElement>) {
+    event.preventDefault();
+
+    setIsTitleValid(() => Boolean(title));
+    setIsUserSelected(() => Boolean(selectedUserId));
```
alexander-ignatow, line 35: Not critical, but there is no much sense in using setter with function parameter if you don't use old state value

src/components/TodoInfo/TodoInfo.tsx
alexander-ignatow: Ok, let's keep it then
Just one thing (optional to fix) - I would rather handle an empty user with the component to avoid this external check every time it is used.

alexander-ignatow (changes requested):
One minor issue is still there

src/components/FormAddTodo/FormAddTodo.tsx
```diff
@@ -0,0 +1,121 @@
...
+            </option>
+            {users.map((user) => {
+              const { id, name } = user;
+
+              return (
+                <option value={id} key={id}>
+                  {name}
+                </option>
+              );
+            })}
```
roman-mirzoian, line 107: Destructuring can be done directly in the argument:
```suggestion
            {users.map(({ id, name }) => (
                <option value={id} key={id}>
                  {name}
                </option>
              ))}
```

roman-mirzoian (approved):
Well done 👍

PR #2011, opened 2023-09-08
react_add-todo-form

vatatan:
[DEMO LINK](https://vatatan.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,142 @@
+import { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoWithUser } from './types/Todo';
+import { TodoList } from './components/TodoList';
+
+export const firstTodos = todosFromServer.map(todo => ({
+  ...todo,
+  user: usersFromServer.find(({ id }) => id === todo.userId) || null,
+}));
 
 export const App = () => {
+  const [todoTitle, setTodoTitle] = useState('');
+  const [hasTitleError, setHasTitleError] = useState(false);
+  const [userSelectedId, setUserSelectedId] = useState(0);
+  const [hasUserSelectedId, setHasUserSelectedId] = useState(false);
+  const [preparedTodos, setPreparedTodos] = useState(firstTodos);
+
+  function getUserById(userId: number) {
+    return usersFromServer.find(({ id }) => id === userId) || null;
+  }
+
+  function onSubmit(todo: TodoWithUser) {
+    setPreparedTodos(currentTodo => [...currentTodo, todo]);
+  }
+
+  function formReset() {
+    setTodoTitle('');
+    setUserSelectedId(0);
+  }
+
+  function handleInputChange(event: React.ChangeEvent<HTMLInputElement>) {
+    setTodoTitle(event.target.value);
+    setHasTitleError(false);
+  }
+
+  function handleUserSelect(event: React.ChangeEvent<HTMLSelectElement>) {
+    setUserSelectedId(+event.target.value);
+    setHasUserSelectedId(false);
+  }
+
+  const handleOnSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (!todoTitle || !userSelectedId) {
+      setHasTitleError(Boolean(!todoTitle));
+      setHasUserSelectedId(Boolean(!userSelectedId));
+
+      return;
+    }
+
+    const todoId = preparedTodos.map(({ id }) => id);
+    const maxTodoId = Math.max(...todoId) + 1;
+
+    onSubmit({
+      id: maxTodoId,
+      title: todoTitle,
+      completed: false,
+      userId: userSelectedId,
+      user: getUserById(userSelectedId),
+    });
+
+    formReset();
+  };
+
   return (
     <div className="App">
-      <h1>Add todo form</h1>
+      <h1>
+        Add todo form
+      </h1>
 
-      <form action="/api/todos" method="POST">
+      <form
+        action="/api/todos"
+        method="POST"
+        onSubmit={handleOnSubmit}
+      >
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <label htmlFor="titleInput">
+            Title:
+          </label>
+
+          <input
+            id="titleInput"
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoWithUser } from './types/Todo';
+import { TodoList } from './components/TodoList';
+
+export const firstTodos = todosFromServer.map(todo => ({
```
roman-mirzoian, line 9: The name "first todos" is a bit confusing, it might be better to call it "initial todos" so that it is more clear what kind of information it is.

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
 export const App = () => {
+  const [todoTitle, setTodoTitle] = useState('');
+  const [hasTitleError, setHasTitleError] = useState(false);
+  const [userSelectedId, setUserSelectedId] = useState(0);
+  const [hasUserSelectedId, setHasUserSelectedId] = useState(false);
+  const [preparedTodos, setPreparedTodos] = useState(firstTodos);
+
+  function getUserById(userId: number) {
+    return usersFromServer.find(({ id }) => id === userId) || null;
+  }
```
roman-mirzoian, line 23: This function should be moved outside the component.

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
+
+  function onSubmit(todo: TodoWithUser) {
+    setPreparedTodos(currentTodo => [...currentTodo, todo]);
+  }
+
+  function formReset() {
```
roman-mirzoian, line 29: Function names must begin with a verb:
```suggestion
  function resetForm() {
```

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
+    event.preventDefault();
+
+    if (!todoTitle || !userSelectedId) {
+      setHasTitleError(Boolean(!todoTitle));
+      setHasUserSelectedId(Boolean(!userSelectedId));
+
+      return;
+    }
+
+    const todoId = preparedTodos.map(({ id }) => id);
```
roman-mirzoian, line 54:
```suggestion
    const todoIds = preparedTodos.map(({ id }) => id);
```

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,17 @@
...
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  return (
+    <a
+      className="UserInfo"
+      href={`mailto:${user?.email}`}
+    >
+      {user?.name}
+    </a>
```
roman-mirzoian, line 15: If there is no information about the user, then you should not return anything at all:
```js
return user && (...)
```

roman-mirzoian (changes requested):

src/App.tsx
BudnikOleksii: > This function should be moved outside the component.

This helper function is still in the component

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
+export const initialTodos = todosFromServer.map(todo => ({
+  ...todo,
+  user: usersFromServer.find(({ id }) => id === todo.userId) || null,
+}));
 
 export const App = () => {
+  const [todoTitle, setTodoTitle] = useState('');
+  const [hasTitleError, setHasTitleError] = useState(false);
+  const [userSelectedId, setUserSelectedId] = useState(0);
+  const [hasUserSelectedId, setHasUserSelectedId] = useState(false);
```
BudnikOleksii, line 18: Is it `hasUserSelectedId` or `hasUserSelectedIdError`?

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
+    setUserSelectedId(+event.target.value);
+    setHasUserSelectedId(false);
+  }
+
+  const handleOnSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (!todoTitle || !userSelectedId) {
+      setHasTitleError(Boolean(!todoTitle));
+      setHasUserSelectedId(Boolean(!userSelectedId));
```
BudnikOleksii, line 49:
```suggestion
      setHasTitleError(!todoTitle);
      setHasUserSelectedId(!userSelectedId);
```
Operator `not` - `!` already converted to boolean.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_NOT

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,17 @@
-export const UserInfo = () => {};
+import React from 'react';
+import { User } from '../../types/User';
+
+type Props = {
+  user: User | null,
+};
```
BudnikOleksii, line 6:
```suggestion
type Props = {
  user: User,
};
```
Don't render a `UserInfo` component at all if you don't have a user. As in your case you render an empty link with invalid values

BudnikOleksii (changes requested):
Nice work 👍 
Left some comments, check then and improve your solution

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,34 @@
...
+      data-id={id}
+      className={classname('TodoInfo', {
+        'TodoInfo--completed': completed,
+      })}
+    >
+      <h2 className="TodoInfo__title">
+        {title}
+      </h2>
+
+      <UserInfo user={user as User} />
```
BudnikOleksii, line 31:
```suggestion
      {user && <UserInfo user={user} />}
```
Don't do this, if the user is `null`, don't render the UserInfo component

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,17 @@
-export const UserInfo = () => {};
+import React from 'react';
+import { User } from '../../types/User';
+
+type Props = {
+  user: User,
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  return user && (
```
BudnikOleksii, line 9: This check should be in the `TodoInfo` component

BudnikOleksii (changes requested):
In typescript, use `as` only when it's necessary and you sure that it is the correct type, in your case it's still can be `null`. Fix it

BudnikOleksii (approved):
Good work 👍

PR #2012, opened 2023-09-08
test solution

cardinal1991:
[DEMO LINK](https://cardinal1991.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,112 @@
+import React, { FormEventHandler, useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
 
-export const App = () => {
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+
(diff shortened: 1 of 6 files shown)


choeqq (changes requested):
code looks fine, please make sure your GH pages URL works correctly

choeqq (changes requested):
styles are off, please make sure you're using correct elements and classes

Meg-Sowka (approved):
Looks good

src/App.tsx
```diff
@@ -1,61 +1,123 @@
...
+      setUserError('Please choose a user');
+
+      return;
+    }
+
+    const userObj = usersFromServer.find(user => user.id === userId);
+
+    if (userObj) {
+      const newTodo: Todo = {
+        id: todos.length + 1,
```
asakevych, line 57: tests failed: 1) should add a title with the max existing id + 1 
so you need to find largest id and add 1 to it's value
[1, 4, 3, 2, 5, 32, 1] -> due to requirements id should be 33, but you would have 7 😉

asakevych (changes requested):
Great job!! one small change left 😉

choeqq (approved):

PR #2014, opened 2023-09-09
add task solution

cty2802:
[DEMO LINK](https://cty2802.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,153 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { User } from './components/types/User';
+import { TodoWithUser } from './components/types/Todo';
+import { TodoList } from './components/TodoList/TodoList';
+
+const preparedTodos: TodoWithUser[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: usersFromServer.find(({ id }) => id === todo.userId) || null,
+} as TodoWithUser));
+
+function findUserById(userId: number): User {
+  return usersFromServer.find(({ id }) => id === userId) as User;
+}
 
 export const App = () => {
+  const [todos, setTodos] = useState<TodoWithUser[]>(preparedTodos);
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [titleError, setTitleError] = useState(false);
+  const [userIdError, setUserIdError] = useState(false);
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+    setTitleError(false);
+  };
+
+  const handleBtAdd = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    let isError = false;
+
+    if (title === '') {
+      setTitleError(true);
+      isError = true;
+    }
+
+    if (userId === 0) {
+      setUserIdError(true);
+      isError = true;
+    }
+
+    if (isError) {
+      return;
+    }
+
+    function resetForm() {
+      setTitle('');
+      setUserId(0);
+    }
+
+    const todoIds = todos.map(({ id }) => id);
+    const maxTodoId = Math.max(...todoIds);
+
+    const newTodo: TodoWithUser = {
+      id: maxTodoId + 1,
+      title,
+      completed: false,
+      userId,
+      user: findUserById(userId),
+    };
+
+    setTodos([...todos, newTodo]);
+
+    resetForm();
+  };
+
+  const handleSelectUserId = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setUserId(+event.target.value);
+
+    setUserIdError(false);
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/todos" method="POST">
+      <form
+        action="/api/todos"
+        method="POST"
+        noValidate
+        onSubmit={handleBtAdd}
+      >
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <label
+            htmlFor="title"
+          >
+            Title:
+          </label>
+
+          <input
+            id="title"
+            type="text"
+            data-cy="titleInput"
+            value={title}
+            onChange={handleTitleChange}
+            required
+            placeholder="Enter a title"
+          />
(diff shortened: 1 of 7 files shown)


kaavka (approved):
Good job!

src/App.tsx
```diff
@@ -1,61 +1,153 @@
...
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { User } from './components/types/User';
+import { TodoWithUser } from './components/types/Todo';
+import { TodoList } from './components/TodoList/TodoList';
+
+const preparedTodos: TodoWithUser[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: usersFromServer.find(({ id }) => id === todo.userId) || null,
+} as TodoWithUser));
```
BudnikOleksii, line 13: Avoid using `as`, in this case, you already declared that `preparedTodos` have `TodoWithUser[]` type

src/App.tsx
```diff
@@ -1,61 +1,153 @@
...
+
+function findUserById(userId: number): User {
+  return usersFromServer.find(({ id }) => id === userId) as User;
```
BudnikOleksii, line 16: The same, if we don't find a user we return `undefind` but you told typescript that it will be user and you can miss an error

src/App.tsx
```diff
@@ -1,61 +1,153 @@
...
+}
 
 export const App = () => {
+  const [todos, setTodos] = useState<TodoWithUser[]>(preparedTodos);
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [titleError, setTitleError] = useState(false);
+  const [userIdError, setUserIdError] = useState(false);
```
BudnikOleksii, line 24: Boolean variables should have the prefix `is` or `has`. Something like this `isTitleError`

src/App.tsx
```diff
@@ -1,61 +1,153 @@
...
+    setTitle(event.target.value);
+    setTitleError(false);
+  };
+
+  const handleBtAdd = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    let isError = false;
+
+    if (title === '') {
```
BudnikOleksii, line 36:
```suggestion
    if (!title) {
```
Don't compare with an empty string, just check if the value exists.
[Falsy values article](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)

The same for the userId.

src/App.tsx
```diff
@@ -1,61 +1,153 @@
...
+  const [userId, setUserId] = useState(0);
+  const [titleError, setTitleError] = useState(false);
+  const [userIdError, setUserIdError] = useState(false);
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+    setTitleError(false);
+  };
+
+  const handleBtAdd = (event: React.FormEvent<HTMLFormElement>) => {
```
BudnikOleksii, line 31:
```suggestion
  const handleAddTodo = (event: React.FormEvent<HTMLFormElement>) => {
```
`BtAdd` what is "Bt", what do we actually add? 😄

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,34 @@
...
+        { 'TodoInfo--completed': completed },
+      )}
+    >
+      <h2 className="TodoInfo__title">
+        {title}
+      </h2>
+
+      <UserInfo user={todo.user} />
+
+    </article>
```
BudnikOleksii, line 32: We don't want to render `UserInfo` if we don't have a user.
Also, remove blank line

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,16 @@
-export const UserInfo = () => {};
+import { User } from '../types';
+
+export type Props = {
+  user: User | null,
+};
```
BudnikOleksii, line 5:
```suggestion
export type Props = {
  user: User,
};
```
We don't want to render UserInfo if we don't have a user. As in your solution we render empty link with invalid `href`

BudnikOleksii (changes requested):
Good start 👍 
Check the comments and improve your solution

src/App.tsx
```diff
@@ -1,61 +1,153 @@
...
+
+    const newTodo: TodoWithUser = {
+      id: maxTodoId + 1,
+      title,
+      completed: false,
+      userId,
+      user: findUserById(userId),
+    };
+
+    setTodos([...todos, newTodo]);
```
roman-mirzoian, line 66: In order to have an up-to-date state, we must take it from the callback:
```suggestion
    setTodos((currentTodos) => [...currentTodos, newTodo]);
```

src/App.tsx
```diff
@@ -1,61 +1,153 @@
...
+        onSubmit={handleAddTodo}
+      >
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <label
+            htmlFor="title"
+          >
+            Title:
+          </label>
```
roman-mirzoian, line 92:
```suggestion
          <label htmlFor="title">
            Title:
          </label>
```

src/App.tsx
```diff
@@ -1,61 +1,153 @@
...
         </div>
 
         <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>Choose a user</option>
+          <label
+            htmlFor="user-id"
+          >
+            User:
+          </label>
```
roman-mirzoian, line 112:
```suggestion
          <label htmlFor="user-id">
            User:
          </label>
```

src/App.tsx
```diff
@@ -1,61 +1,153 @@
...
+              Choose a user
+            </option>
+            {usersFromServer.map((user: User) => (
+              <option
+                key={`${user.name}_user.name`}
+                value={user.id}
+              >
+                {user.name}
+              </option>
+            ))}
```
roman-mirzoian, line 134: It can be written shorter:
```suggestion
            {usersFromServer.map(({ id, name }) => (
              <option key={id} value={id}>
                {name}
              </option>
            ))}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,22 @@
...
+    <section className="TodoList">
+      {todos.map((todo: TodoWithUser) => {
+        return (
+          <TodoInfo
+            key={todo.id}
+            todo={todo}
+          />
+        );
+      })}
+
```
roman-mirzoian, line 19: Don't forget about good code formatting.
```suggestion
```

roman-mirzoian (approved):
Well done, but pay attention to the comments.

PR #2022, opened 2023-09-10
add solution

dimarogkov:
[DEMO LINK](https://dimarogkov.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,30 +4,22 @@ import usersFromServer from './api/users';
 import todosFromServer from './api/todos';
 import { TodoList } from './components/TodoList';
 import { Todo } from './interfaces/todo';
-import { User } from './interfaces/user';
-import { Form } from './components/Form/Form';
+import { Form } from './components/Form';
+import { getUserById } from './helpers/getUserById';
+
+const todosWithUser = todosFromServer
+  .map(todo => ({ ...todo, todoUser: getUserById(todo.userId) }));
 
 export const App = () => {
-  const [todosArr, setTodosArr] = useState<Todo[]>(todosFromServer);
-
+  const [todosArr, setTodosArr] = useState<Todo[]>(todosWithUser);
   const addTodo = (todo: Todo) => setTodosArr([...todosArr, todo]);
-
-  const createTodosArr = (todos: Todo[], users: User[]) => {
-    const getUserById = (userId: number) => {
-      return users.find(user => user && user.id === userId);
-    };
-
-    return todos.map(todo => ({ ...todo, todoUser: getUserById(todo.userId) }));
-  };
-
-  const finalTodos = createTodosArr(todosArr, usersFromServer);
 
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
       <Form users={usersFromServer} addTodo={addTodo} />
-      <TodoList todos={finalTodos} />
+      <TodoList todos={todosArr} />
     </div>
   );
 };
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,33 @@
 import './App.scss';
-
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { useState } from 'react';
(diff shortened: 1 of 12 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,21 @@
...
+import { TodoInfo } from '../TodoInfo';
+
+interface Props {
+  todos: {
+    todoUser: User | undefined;
+    id: number;
+    title: string;
+    completed: boolean;
+    userId: number;
+  }[];
```
etojeDenys, line 12:
```suggestion
  todos: (Todo & {todoUser?: User})[];
```
we can simplify it

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,105 @@
...
+      setUserId(0);
+      setUserIdInvalid(false);
+    }
+  };
+
+  return (
+    <form
+      action="/api/todos"
+      method="POST"
+      onSubmit={(event) => submitForm(event)}
```
etojeDenys, line 65:
```suggestion
      onSubmit={submitForm}
```
we can simplify it

etojeDenys (changes requested):
i can create the todo only with spaces, let's fix this case

src/interfaces/todo.ts
```diff
@@ -0,0 +1,6 @@
+export interface Todo {
+  id: number,
+  title: string,
+  completed: boolean,
+  userId: number,
```
VitaliyBondarenko1982, line 5:
```suggestion
  userId: number,
  todoUser?: User,
```
and use this type everwhere

src/App.tsx
```diff
@@ -1,61 +1,33 @@
...
-            delectus aut autem
-          </h2>
+    return todos.map(todo => ({ ...todo, todoUser: getUserById(todo.userId) }));
+  };
 
-          <a className="UserInfo" href="mailto:Sincere@april.biz">
-            Leanne Graham
-          </a>
-        </article>
+  const finalTodos = createTodosArr(todosArr, usersFromServer);
```
VitaliyBondarenko1982, line 23: better to make this function outside of component for initial todos state and after that change just state of todos.
in your implementation you make this operation during add each new todo

```
const getUserById = (userId: number) => {
  return usersFromServer.find(user => user && user.id === userId) || null;
};

const todosWithUser = todosFromServer
  .map(todo => ({ ...todo, todoUser: getUserById(todo.userId) }));
```

src/App.tsx
```diff
@@ -1,61 +1,33 @@
...
-        <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
-        </div>
-
-        <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>Choose a user</option>
-          </select>
+  const [todosArr, setTodosArr] = useState<Todo[]>(todosFromServer);
```
VitaliyBondarenko1982, line 11:
```suggestion
  const [todosArr, setTodosArr] = useState<Todo[]>(todosWithUser);
```

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    if (title.length > 0 && userId > 0) {
+      setId(prev => prev + 1);
+
+      const newTodo: Todo = {
+        id,
+        title,
+        completed: false,
+        userId,
```
VitaliyBondarenko1982, line 50:
```suggestion
      const newTodo: Todo = {
        id,
        title,
        completed: false,
        userId,
        todoUser: getUserById(userId)
```
and here you can add user immediately during make new todo

VitaliyBondarenko1982 (changes requested):
Good job.
Needs some improvements.

etojeDenys (changes requested):
still can add the todo only with spaces
image

etojeDenys (approved):
good job

PR #2035, opened 2023-09-11
add task solution

szymonpachucki:
[DEMO LINK](https://szymonpachucki.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,1 +1,31 @@
-export const TodoList = () => {};
+import React from 'react';
+import { TodoInfo } from '../TodoInfo';
+
+type Todo = {
+  id: number,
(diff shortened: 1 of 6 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,31 @@
...
+  userId: number,
+  user: User | null,
+};
+
+type User = {
+  id: number,
+  name: string,
+  username: string,
+  email: string,
+};
```
Matylda-Lewandowska, line 17: I'm not sure, if this should be in other file precisely for each special type file :)

Matylda-Lewandowska (approved):
Only thinking about one part, but all over it looks good to me. Good Job! 👍

choeqq (approved):

PR #2037, opened 2023-09-11
add solution

prostonadobobb:
[DEMO LINK](https://prostonadobobb.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,112 @@
 import './App.scss';
+import { useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Todo } from './components/types';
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,120 @@
...
+  const [titleError, setTitleError] = useState('');
+  const [userError, setUserError] = useState('');
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (!title) {
+      setTitleError('Please enter title');
+
+      return;
```
nazarmatsevych, line 30: Because of this return, we can only see one error message at a time. We can try to combine this checks or create boolean flag to allow or disallow submit

nazarmatsevych (changes requested):

sTorba24 (approved):
GJ 🚀

PR #2039, opened 2023-09-11
Add task solution

ArtemBehter:


src/App.tsx
```diff
@@ -1,61 +1,144 @@
...
+      id: 0,
+      title: '',
+      completed: false,
+      user: null,
+    },
+  );
+
+  const [isFilled, setIsFilled] = useState(true);
+
+  const onAdd = (newTodo: ToDo) => {
```
darokrk, line 31: this function name should be rather `addTodo` to have more descriptive name instead of `onAdd`

src/App.tsx
```diff
@@ -1,61 +1,144 @@
...
+import { TodoList } from './components/TodoList';
+import { ToDo } from './types/type';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+function getUser(id: number) {
+  return usersFromServer.find(usr => (id) === usr.id) || null;
```
darokrk, line 10: brackets around id are redundant here, just fix it to be:

`.find(usr => id === usr.id) || null`

src/App.tsx
```diff
@@ -1,61 +1,144 @@
...
+  );
+
+  const [isFilled, setIsFilled] = useState(true);
+
+  const onAdd = (newTodo: ToDo) => {
+    setTodos(currentTodos => [
+      ...currentTodos,
+      {
+        ...newTodo,
+        id: Math.max(...(todosFromServer.map(toDo => toDo.id))) + 1,
```
darokrk, line 36: why not todo instead of toDo inside function :) ?

src/App.tsx
```diff
@@ -1,61 +1,144 @@
...
+    setTodo(
+      { ...todo, title: event.target.value },
+    );
+  };
+
+  const handleUserIdChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setTodo(
+      {
+        ...todo,
+        user: getUser(+event.target.value),
```
darokrk, line 64: easier for readability will be use explicit Number() constructor to convert value to number, e.g.:

`user: getUser(Number(event.target.value))`

src/App.tsx
```diff
@@ -1,61 +1,144 @@
...
+            Title:
+            <input
+              type="text"
+              value={todo.title}
+              data-cy="titleInput"
+              name="title"
+              placeholder="Enter a title"
+              onChange={handleTitleChange}
+            />
+            {(!todo.title.trim() && !isFilled)
```
darokrk, line 102: you can move this check to some variable, to simplify this code, e.g. :

`const hasTitleError = !todo.title.trim() && !isFilled;`

and then just use it here for check:

`{hasTitleError && <span className="error">Please enter a title</span>}`

src/App.tsx
```diff
@@ -1,61 +1,144 @@
...
+          <label htmlFor="select">
+            User:
+            <select
+              name="select"
+              data-cy="userSelect"
+              onChange={handleUserIdChange}
+              value={!todo.user ? 0 : todo.user.id}
+            >
+              <option value="0" disabled>Choose a user</option>
+              {usersFromServer.map(usr => (
```
darokrk, line 117: `usr` is not such descriptive name as `user`, please rename argument in function

src/App.tsx
```diff
@@ -1,61 +1,144 @@
...
+                  value={usr.id}
+                  key={usr.id}
+                >
+                  {usr.name}
+                </option>
+              ))}
+            </select>
 
-          <span className="error">Please choose a user</span>
+            {(!todo.user && !isFilled)
```
darokrk, line 127: same as before you can move this check to some variable instead of doing this inline

src/App.tsx
```diff
@@ -1,61 +1,144 @@
...
+    onAdd(todo);
+    resetForm();
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/todos" method="POST">
+      <form
```
darokrk, line 86: You can even move this whole form logic into a separate component, like TodoForm.tsx, and use it here as:

`<TodoList todos={todos} />`

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,26 @@
-export const TodoInfo = () => {};
+import React from 'react';
+import cn from 'classnames';
+import { UserInfo } from '../UserInfo';
+import { ToDo } from '../../types';
```
darokrk, line 4: please rename it to `Todo` instead of `ToDo`

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,26 @@
...
+
+type Props = {
+  todo: ToDo,
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  return (
```
darokrk, line 11: return word is redundant here, just use brackets:

```
({ todo }) => (
  <article
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,21 @@
...
+import React from 'react';
+import { TodoInfo } from '../TodoInfo';
+import { ToDo } from '../../types';
+
+type Props = {
+  todos: ToDo[],
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
```
darokrk, line 10: same as above, the `return` word is redundant, please use just brackets

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,21 @@
...
+    <section className="TodoList">
+      {todos.map(todo => (
+        <TodoInfo
+          key={todo.id}
+          todo={todo}
+        />
+      ))}
+      ;
```
darokrk, line 18: There is an unnecessary semicolon after the TodoInfo component mapping. Removing this will clean up the code.

image

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,17 @@
-export const UserInfo = () => {};
+import React from 'react';
+import { User } from '../../types';
+
+type UsrInfo = {
+  user: User
+};
+
+export const UserInfo: React.FC<UsrInfo> = ({ user }) => {
+  return (
```
darokrk, line 9: same as before remove redundant `return` word just use `()`

src/types/type.ts
darokrk, line 1: this file should be rather named `types.ts` instead of `type.ts`, bcs it contains multiple types instead of one

darokrk (changes requested):
UI looks good 👏🏽  left comments for code improvement 👨🏽‍💻

plodnikate (approved):
Good job ! :fire:

PR #2047, opened 2023-09-13
add task solution

UlianaChorna:
[DEMO LINK](https:UlianaChorna.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,40 @@
 import './App.scss';
+import { useState } from 'react';
+import { Todo } from './components/TodoInfo';
+import { TodoList } from './components/TodoList';
+import { getUserById } from './services/getUser';
+import todosFromServer from './api/todos';
+import { TodoForm } from './components/TodoForm/TodoForm';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+const initialTodo = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
+
+function getNewTodoId(todos: Todo[]) {
+  const maxId = Math.max(...todos.map(todo => todo.id));
+
+  return maxId + 1;
+}
 
 export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>(initialTodo);
+
+  const addTodo = (todo: Todo) => {
+    const newTodo = {
+      ...todo,
+      id: getNewTodoId(todos),
+    };
+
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,40 @@
...
+const initialTodo = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
+
+function getNewTodoId(todos: Todo[]) {
+  const maxId = Math.max(...todos.map(todo => todo.id));
+
+  return maxId + 1;
+}
```
DarkMistyRoom, line 18: looks like service. Move to the `services` folder

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+    setHasTitleError(!title);
+    setUserNameError(!userId);
+
+    if (!title || !userId) {
+      return;
+    }
+
+    onSubmit({
+      id: 0,
```
DarkMistyRoom, line 27: why do you need to pass `id: 0` and then change it? Just set the correct `id` here or don't pass it

DarkMistyRoom (changes requested):
I can add a todo with white spaces only. Handle this case
image

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+    setHasTitleError(!checkTitle);
+    setUserNameError(!userId);
+
+    if (!checkTitle || !userId) {
+      return;
+    }
+
+    onSubmit({
+      id: 1256,
```
mykhalenych, line 28: generate uniq id on each submit

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+type Props = {
+  onSubmit: (todo: Todo) => void;
+};
+
+export const TodoForm: React.FC<Props> = ({ onSubmit }) => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [userNameError, setUserNameError] = useState(false);
+  const [hasTitleError, setHasTitleError] = useState(false);
```
mykhalenych, line 14:
```suggestion
  const [hasTitleError, setHasTitleError] = useState(false);
  
```

mykhalenych (changes requested):

mykhalenych (approved):

PR #2050, opened 2023-09-14
task solution was added

Liliya-Kalinichenko:
[DEMO LINK](https://Liliya-Kalinichenko.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,124 @@
-import './App.scss';
+import React, { useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import './App.scss';
+import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/Todo';
+import { getNewTodoId, getUserById } from './helper';
+
+const initialTodos = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
 
 export const App = () => {
+  const [todos, setTodos] = useState(initialTodos);
+  const [title, setTitle] = useState('');
+  const [titleError, setTitleError] = useState(false);
+  const [userId, setUserId] = useState(0);
+  const [userIdError, setUserIdError] = useState(false);
+
+  const addTodo = (newTodo: Todo) => {
+    setTodos(currentTodos => [...currentTodos, newTodo]);
+  };
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const invalidSymbols = /[^a-zA-Za-åa-ö-w-я 0-9/ .,]/gi;
+    const normalizedValue = event.target.value.replaceAll(invalidSymbols, '');
+
+    setTitle(normalizedValue);
+    setTitleError(false);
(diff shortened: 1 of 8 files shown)


Anastasiia-Svintsova (changes requested):
The user shouldn't be able to add a todo with an empty title. Use the trim() method to check it 
image

src/App.tsx
```diff
@@ -1,61 +1,124 @@
...
+    addTodo({
+      id: getNewTodoId(todos),
+      title,
+      userId,
+      completed: false,
+      user: getUserById(userId),
+    });
+
+    setTitle('');
+    setUserId(0);
```
Viktor-Kost, line 63: move to resetForm function

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,24 @@
...
+import cn from 'classnames';
+import { UserInfo } from '../UserInfo';
+import { Todo } from '../../types/Todo';
+
+interface Props {
+  todo: Todo
+}
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  return (
```
Viktor-Kost, line 10: it's a good practice to use object destructing
```suggestion
export const TodoInfo: React.FC<Props> = ({ todo }) => {
  const {...} = todo;
  
  return (
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,24 @@
...
+    <article
+      data-id={todo.id}
+      className={cn('TodoInfo', {
+        'TodoInfo--completed': todo.completed === true,
```
Viktor-Kost, line 14:
```suggestion
        'TodoInfo--completed': completed,
```

src/App.tsx
```diff
@@ -1,61 +1,124 @@
...
+      setTitleError(true);
+    }
+
+    if (!userId) {
+      setUserIdError(true);
+    }
+
+    if (!title.trim() || !userId) {
+      return;
+    }
```
Viktor-Kost, line 52: no needs to add the same condition as you did earlier 
```suggestion
    if (!title.trim()) {
      setTitleError(true);
      return;
    }

    if (!userId) {
      setUserIdError(true);
      return;
    }
```

src/App.tsx
```diff
@@ -1,61 +1,124 @@
...
+  const [userId, setUserId] = useState(0);
+  const [userIdError, setUserIdError] = useState(false);
+
+  const addTodo = (newTodo: Todo) => {
+    setTodos(currentTodos => [...currentTodos, newTodo]);
+  };
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const invalidSymbols = /[^a-zA-Za-åa-ö-w-я 0-9/ .,]/gi;
+    const normalizedValue = event.target.value.replaceAll(invalidSymbols, '');
```
Viktor-Kost, line 28: it's a good practice to move such functions to helpers (outside handleTitleChange and App)

Viktor-Kost (changes requested):
Nice!
Some minor fixes needed

sTorba24 (approved):
Cool!

PR #2053, opened 2023-09-14
Develop

NikitaOrliak:
[DEMO LINK](https://NikitaOrliak.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,132 @@
 import './App.scss';
+import { useState } from 'react';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { Todo } from './helpers/Todo';
+
+function getUserById(userId: number) {
+  return usersFromServer.find(user => user.id === userId)
+      || null;
+}
+
+const defaultTodos = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
+
+function getMaxId(todos: Todo[]) {
+  const maxId = Math.max(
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,140 @@
...
+          <select
+            data-cy="userSelect"
+            value={userId}
+            onChange={handleUserIdChange}
+          >
             <option value="0" disabled>Choose a user</option>
-          </select>
+            {
+              usersFromServer.map(user => (
+                <option
```
volodymyr-soltys97, line 115: Use a destructuring for `user`

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,37 @@
...
+        'TodoInfo--completed': completed,
+      })}
+    >
+      <h2 className="TodoInfo__title">
+        {title}
+      </h2>
+
+      {
+        user && (
+          <UserInfo user={user} key={userId} />
```
etojeDenys, line 32: why do we need to pass key here?
we usually pass key when looping the array, so in this case, you can remove this prop

src/App.tsx
```diff
@@ -1,61 +1,144 @@
...
+          <select
+            data-cy="userSelect"
+            value={userId}
+            onChange={handleUserIdChange}
+          >
             <option value="0" disabled>Choose a user</option>
-          </select>
+            {
+              usersFromServer.map(user => {
+                const { id, name } = user;
```
etojeDenys, line 115:
```suggestion
              usersFromServer.map(({id, name}) => {
```

etojeDenys (changes requested):

plodnikate (approved):
Good job !

PR #2056, opened 2023-09-18
Done

yurii-shkrobut-m:
[DEMO LINK](https://yurii-shkrobut-m.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -29,6 +29,8 @@ export const App = () => {
 
   const [todos, setTodos] = useState<Todo[]>(initialTodos);
 
+  const isEmpty = !title.trim().length;
+
   const handleTitleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     setTitle(e.target.value);
     setHasTitleError(false);
@@ -42,10 +44,10 @@ export const App = () => {
   const handleSubmit = (e: React.FormEvent) => {
     e.preventDefault();
 
-    setHasTitleError(!title);
+    setHasTitleError(!title || isEmpty);
     setHasUserIdError(!userId);
 
-    if (!title || !userId) {
+    if (!title || !userId || isEmpty) {
       return;
     }
 
@@ -102,12 +104,12 @@ export const App = () => {
               Choose a user
             </option>
 
-            {usersFromServer.map(user => (
+            {usersFromServer.map(({ id, name }) => (
               <option
-                key={user.id}
-                value={user.id}
+                key={id}
+                value={id}
               >
-                {user.name}
+                {name}
               </option>
             ))}
           </select>
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -6,19 +6,25 @@ interface Props {
   todo: Todo,
 }
 
-export const TodoInfo: React.FC<Props> = ({ todo }) => {
+export const TodoInfo: React.FC<Props> = ({
+  todo: {
+    id, title, completed, user,
+  },
+}) => {
   return (
-    <article
-      data-id={todo.id}
+    <li
+      data-id={id}
       className={cn('TodoInfo', {
-        'TodoInfo--completed': todo.completed,
+        'TodoInfo--completed': completed,
       })}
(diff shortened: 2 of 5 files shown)


src/App.tsx
```diff
@@ -1,61 +1,130 @@
...
+import todosFromServer from './api/todos';
+import { Todo } from './types/Todo';
+
+function getUserById(userId: number) {
+  return usersFromServer.find(user => user.id === userId)
+    || null;
+}
+
+function getNewPostId(posts: Todo[]) {
+  return Math.max(...posts.map(post => post.id)) + 1;
```
VitaliyBodnar, line 15: There is some redundant logic, you can make random ID with library uuidv4, or just with new Date()

VitaliyBodnar (changes requested):
Almost done!

src/App.tsx
yurii-shkrobut-m: image

yurii-shkrobut-m: image

yurii-shkrobut-m: image

src/App.tsx
```diff
@@ -1,61 +1,130 @@
...
+            onChange={handleUserIdChange}
+          >
+            <option value="0">
+              Choose a user
+            </option>
+
+            {usersFromServer.map(user => (
+              <option
+                key={user.id}
+                value={user.id}
```
loralevitska, line 108: Destructure `user`

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,24 @@
...
+interface Props {
+  todo: Todo,
+}
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  return (
+    <article
+      data-id={todo.id}
+      className={cn('TodoInfo', {
+        'TodoInfo--completed': todo.completed,
```
loralevitska, line 14: Destructure `todo`

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,19 @@
...
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo';
+
+interface Props {
+  todos: Todo[],
+}
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="TodoList">
```
loralevitska, line 10: Use semantic tags for list `ul` `li`. Don't forget to remove default `li` dots and `ul` paddings

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,24 @@
...
+      })}
+    >
+      <h2 className="TodoInfo__title">
+        {todo.title}
+      </h2>
+
+      <UserInfo user={todo.user} />
```
loralevitska, line 21: Don't render `UserInfo` component if there is no user. It's redundant 
```suggestion
     {todo.user && <UserInfo user={todo.user} />}
```

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,16 @@
...
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  return (
+    <a
+      className="UserInfo"
+      href={`mailto:${user?.email}`}
+    >
+      {user?.name}
+    </a>
+  );
+};
```
loralevitska, line 16: According comment about this component after fixing code there always will be user object, so we can destructure it 🤭 

```suggestion
export const UserInfo: React.FC<Props> = ({ user }) => {
  const { email, name } = user;

  return (
    <a
      className="UserInfo"
      href={`mailto:${email}`}
    >
      {name}
    </a>
  );
};
```

loralevitska (changes requested):
Let's improve your string validation)

- fix adding empty comments, use [`string.trim()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim)
image

lerastarynets (approved):
You've got this, keep pushing forward!

PR #2057, opened 2023-09-20
solution

Capoverde:
[DEMO LINK](https://Capoverde.github.io/react_add-todo-form/)
 
 I've been combining as much as I can, but I don't know how to avoid the errors that occur.


src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,26 @@
...
+import { User } from '../../type/user';
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: Array<Todo>;
+  users: Array<User>;
+};
+
+export const TodoList: React.FC<Props> = ({ todos, users }) => {
+  const maxTodoId = Math.max(...todos.map((todo) => todo.id));
```
choeqq, line 12: why do you need `maxTodoId` ?

choeqq (approved):

PR #2059, opened 2023-09-21
done

Aleksey-Aleksey:
[DEMO LINK](https://Aleksey-Aleksey.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,118 @@
 import './App.scss';
+import { useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+import { getUserById } from './services/user';
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [titleError, setTitleError] = useState(false);
+
+  const [userId, setUserId] = useState(0);
+  const [userIdError, setUserIdError] = useState(false);
+
+  const [todoList, setTodoList] = useState<Todo[]>(todos);
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+    setTitleError(false);
+  };
+
+  const handleSelectChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setUserId(+event.target.value);
+    setUserIdError(false);
+  };
+
+  const reset = () => {
+    setTitle('');
+    setUserId(0);
+  };
+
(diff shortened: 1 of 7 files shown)


etojeDenys (changes requested):
I can create the todo with spaces
image
fix this case and then re-request me

Aleksey-Aleksey:
> I can create the todo with spaces image fix this case and then re-request me

image

src/App.tsx
```diff
@@ -1,61 +1,126 @@
...
+        onSubmit={handleSubmit}
+      >
+        <label className="field" htmlFor="title">
+          {'Title: '}
+          <input
+            id="title"
+            type="text"
+            data-cy="titleInput"
+            placeholder="Enter a title"
+            value={title.trimStart()}
```
lerastarynets, line 86: u trim value here, which means if value has only white spaces, a todo with empty title will be added. user should see validation error and todo shouldn't be created if title consists of white spaces

lerastarynets (changes requested):
Pls fix prev comments

VitaliyBodnar (changes requested):
- I still can create todos without name
- you need to use method trim for validation 
Screenshot 2023-09-22 at 13 56 33

Aleksey-Aleksey:
> * I still can create todos without name
> * you need to use method trim for validation
> 
> Screenshot 2023-09-22 at 13 56 33

Please check again, because i fixed it in vs code and it works correct during npm start.

Aleksey-Aleksey:
> > * I still can create todos without name
> > * you need to use method trim for validation
> > 
> > Screenshot 2023-09-22 at 13 56 33
> 
> Please check again, because i fixed it in vs code and it works correct during npm start.

Aleksey-Aleksey:
i dont know why but the link didint update. i even changed my code but the link still works not correct

src/App.tsx
```diff
@@ -1,61 +1,126 @@
...
+
+    if (!title.trim()) {
+      setTitleError(true);
+    }
+
+    if (!userId) {
+      setUserIdError(true);
+    }
+
+    if (!userId || !title.trim()) {
```
VitaliyBodnar, line 61: You need to use && here, because || will be completed when one of conditions will be true
```suggestion
    if (!userId && !title.trim()) {
```

VitaliyBodnar (changes requested):

src/App.tsx
Aleksey-Aleksey: not vise versa? if i do like you told me  - i can add only one field . but if i use my aproach i cant add one field

IvanFesenko (changes requested):
Please re-deploy your work
https://github.com/mate-academy/react_add-todo-form/assets/61904995/7e047700-9a02-49b5-96d7-ff9b97a38b77

Aleksey-Aleksey:
> Please re-deploy your work https://github.com/mate-academy/react_add-todo-form/assets/61904995/7e047700-9a02-49b5-96d7-ff9b97a38b77
done

VitaliyBodnar (approved):

PR #2067, opened 2023-09-25
Develop

Sirius9312:
LINK: https://Sirius9312.github.io/react_add-todo-form/
COMMIT: Add Todo Form 2

### DIFF
--- a/src/components/TodoForm/TodoForm.tsx
+++ b/src/components/TodoForm/TodoForm.tsx
@@ -1,0 +1,154 @@
+import { useState } from 'react';
+import usersFromServer from '../../api/users';
+
+type User = {
+  id: number;
+  name: string;
+  username: string;
+  email: string;
+};
+
+type Todos = {
+  user: User | null;
+  id: number;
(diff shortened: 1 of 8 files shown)


src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,141 @@
...
+          hasErrorUserId && (
+            <span className="error">{hasErrorUserId}</span>
+          )
+        }
+
+        {
+          userId && hasErrorUserId && (
+            setHasErrorUserId('')
+          )
+        }
```
etojeDenys, line 130: better to move such conditions in useEffect or handle function

etojeDenys (changes requested):
I can create the empty todo, let's fix this case
image

Sirius9312:
LINK: https://sirius9312.github.io/react_add-todo-form/
COMMIT: Fixed

volodymyr-soltys97 (approved):
Great work 🔥

PR #2072, opened 2023-09-27
Develop

IraPolchanin:
https://irapolchanin.github.io/react_add-todo-form/

### DIFF
--- a/src/services/user.ts
+++ b/src/services/user.ts
@@ -1,0 +1,7 @@
+import usersFromServer from '../api/users';
+import { User } from '../types/User';
+
+export function getUserById(userId: number):User | null {
+  return usersFromServer.find(user => user.id === userId)
+    || null;
+}
--- a/src/types/Todo.ts
+++ b/src/types/Todo.ts
@@ -1,0 +1,9 @@
+import { User } from "./User";
+
+export interface Todo {
+  user: User | null;
+  id: number;
+  title: string;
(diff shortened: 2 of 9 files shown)


src/services/user.ts
```diff
@@ -0,0 +1,7 @@
+import usersFromServer from '../api/users';
+import { User } from '../types/User';
+
+export function getUserById(userId: number):User | null {
+  return usersFromServer.find(user => user.id === userId)
+    || null;
+}
```
VitaliyBodnar, line 7: null is redundant here
```suggestion
export function getUserById(userId: number):User {
  return usersFromServer.find(user => user.id === userId);
}
```

VitaliyBodnar (changes requested):
- there is no validation for empty text
Screenshot 2023-09-27 at 14 40 52

src/types/Todo.ts
```diff
@@ -0,0 +1,9 @@
+import { User } from './User';
+
+export interface Todo {
+  user: User | null;
```
IvanFesenko, line 4: To fix this issue you should also update type
2023-09-27_15h53_46

```suggestion
  user?: User;
```

IvanFesenko (approved):

IraPolchanin:
Thanks!

ср, 27 вер. 2023 р. о 15:54 Ivan Fesenko ***@***.***> пише:

PR #2076, opened 2023-09-29
finally

Alessandranitti15:
[DEMO LINK](https://Alessandranitti15.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,95 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  // if there is no user with a given userId
+  return foundUser || null;
+}
+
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,103 @@
...
+}
+
+export const createTodos = () => todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState('0');
```
lerastarynets, line 24: why do u set id to string here if u transform it back to number afterwards?

src/App.tsx
```diff
@@ -1,61 +1,103 @@
...
+            name="user"
+            id="user"
+            value={userId}
+            onChange={(e) => {
+              setUserId(e.target.value);
+            }}
+          >
             <option value="0" disabled>Choose a user</option>
+            {usersFromServer.map((user) => (
+              <option value={user.id}>{user.name}</option>
```
lerastarynets, line 88: where is `key` prop?

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

src/App.tsx
anastasiia-tilikina: didn't fixed.

anastasiia-tilikina (changes requested):
please fix one more issue

lerastarynets (approved):
You've got this, keep pushing forward!

PR #2078, opened 2023-10-01
Solution Add Todo form

HannieA:
https://HannieA.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,33 @@
+import React, { useState } from 'react';
+
 import './App.scss';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+import { TodoForm } from './components/TodoForm';
+import { getUser } from './utils';
+import todosFromServer from './api/todos';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+const todos: Todo[] = todosFromServer.map((todo) => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
-export const App = () => {
+export const App: React.FC = () => {
+  const [updatedTodo, setUpdatedTodo] = useState(todos);
+
+  const addNewTodo = (todo: Todo) => {
+    const newTodo = {
+      ...todo,
+    };
+
+    setUpdatedTodo(prevTodo => [...prevTodo, newTodo]);
+  };
+
   return (
-    <div className="App">
-      <h1>Add todo form</h1>
-
-      <form action="/api/todos" method="POST">
-        <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
-        </div>
-
-        <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>Choose a user</option>
-          </select>
-
-          <span className="error">Please choose a user</span>
-        </div>
-
-        <button type="submit" data-cy="submitButton">
-          Add
-        </button>
-      </form>
-
-      <section className="TodoList">
-        <article data-id="1" className="TodoInfo TodoInfo--completed">
-          <h2 className="TodoInfo__title">
-            delectus aut autem
-          </h2>
-
-          <a className="UserInfo" href="mailto:Sincere@april.biz">
-            Leanne Graham
-          </a>
-        </article>
-
-        <article data-id="15" className="TodoInfo TodoInfo--completed">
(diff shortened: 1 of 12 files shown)


anastasiia-tilikina (changes requested):
1. Great styling 🔥 
2. Please read this [article](https://javascript.plainenglish.io/handy-naming-conventions-for-event-handler-functions-props-in-react-fc1cbb791364) and use proper event handlers' names in the future
3. Read my comments
4. Watch the [video](https://www.loom.com/share/132a672b65af4568a88487e8cda09cf7?sid=a5749f37-9102-4592-ae71-f3b59f896653)

HannieA:
Thank you, your comments are very useful!

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,113 @@
...
+    setHasUserError(false);
+    setHasTitleError(false);
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (userId === 0) {
+      setHasUserError(true);
+    }
```
anastasiia-tilikina, line 40:
```suggestion
    if (userId === 0) {
      setHasUserError(true);
      
      return;
    }
```

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+    if (!title || !title.trim()) {
+      setHasTitleError(true);
+    }
```
anastasiia-tilikina, line 44:
```suggestion
    if (!title.trim()) {
      setHasTitleError(true);
      
      return;
    }
```

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,113 @@
...
+      addTodo({
+        id: 0,
+        title: title.trim(),
+        completed: false,
+        userId,
+        user: getUser(userId),
+      });
+
+      restart();
+    }
```
anastasiia-tilikina, line 56:
```suggestion
      addTodo({
        title: title.trim(),
        completed: false,
        userId,
        user: getUser(userId),
      });

      restart();
```

src/components/TodoForm/TodoForm.tsx
anastasiia-tilikina: we add return in this if brunch to make our method not submit if there is userError

src/components/TodoForm/TodoForm.tsx
anastasiia-tilikina: same here

src/components/TodoForm/TodoForm.tsx
anastasiia-tilikina: now we can avoid additional if statement since we have already checked title and user

anastasiia-tilikina: we don't need to add some id here since we'll do it in app component

src/components/TodoForm/TodoForm.tsx
anastasiia-tilikina: also check for `title.trim()` will be enough here

src/App.tsx
```diff
@@ -1,61 +1,41 @@
...
+
+const todos: Todo[] = todosFromServer.map((todo) => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [updatedTodo, setUpdatedTodo] = useState(todos);
+
+  const addNewTodo = (todo: Todo) => {
```
anastasiia-tilikina, line 19:
```suggestion
  const addNewTodo = (todo: Omit<Todo, 'id'>) => {
```

anastasiia-tilikina: and here we need to change parameter type since our parameter won't contain id
It is called [utility type](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys)

anastasiia-tilikina (changes requested):
You did really great 🔥 
Just a couple more changes to make your code even more clear)

src/components/TodoForm/TodoForm.tsx
HannieA: I did like you suggested, but in this case the error message is shown only for field in the first conditional statement even if two fields are empty / not selected. maybe I'm missing something

lerastarynets (approved):
You're on the right track!

PR #2084, opened 2023-10-03
Solution

YuraMisura:


src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,115 @@
...
+  todos: Todo[],
+  onAdd: (newTodo: Todo) => void;
+};
+
+export const TodoForm: React.FC<Props> = ({ todos, onAdd }) => {
+  const [title, setTitle] = useState('');
+  const [hasTitleError, setHasTitleError] = useState(false);
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const textValue = event.target.value.replace(/[^a-zA-ZіІїЇєЄґҐ а-я]+/gu, '');
```
VitaliyBondarenko1982, line 16:
```suggestion
    const textValue = event.target.value
      .replace(/[^a-zA-ZіІїЇєЄґҐ а-я]+/gu, '');
```
move to other line should fix such linter issue

PR #2091, opened 2023-10-10
added task solution

koros-rk:
[DEMO LINK](https://koros-rk.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,48 @@
+import { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { TodoForm } from './components/TodoForm';
+import { PreparedTodo, Todo } from './types/Todo';
+import { User } from './types/User';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+type Prepare = (todos: Todo[], users: User[]) => PreparedTodo[];
+
+const prepareTodos: Prepare = (todos, users) => {
+  const prepared = [...todos];
+
+  prepared.forEach((todo) => {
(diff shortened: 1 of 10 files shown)


yulabakun (approved):
Well Done 😉

src/App.tsx
```diff
@@ -1,61 +1,48 @@
...
+import todosFromServer from './api/todos';
+
+type Prepare = (todos: Todo[], users: User[]) => PreparedTodo[];
+
+const prepareTodos: Prepare = (todos, users) => {
+  const prepared = [...todos];
+
+  prepared.forEach((todo) => {
+    Object.defineProperty(todo, 'user', {
+      value: users.find((user) => user.id === todo.userId),
```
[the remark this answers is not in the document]
polosanya: ➕ , also here it is better to use map method instead of the forEach. The current solution is a bit overcomplicated

src/App.tsx
```diff
@@ -1,61 +1,48 @@
...
+    });
+  });
+
+  return prepared as PreparedTodo[];
+};
 
 export const App = () => {
+  const [todos, setTodos]
+    = useState(prepareTodos(todosFromServer, usersFromServer));
```
polosanya, line 27: image

src/App.tsx
```diff
@@ -1,61 +1,48 @@
...
-      </section>
+      <TodoForm
+        todos={todos}
+        createNewTodo={(newTodo: PreparedTodo) => {
+          setTodos(prevState => {
+            return [
+              ...prevState,
+              newTodo,
+            ];
+          });
```
polosanya, line 41: Would be great to move such logic into a separate function

polosanya (changes requested):
Good job! Let's improve your code a bit)

StasSokolov (approved):
LGTM🔥

PR #2099, opened 2023-10-14
finished all the tests

EvhenIvchenko:
[DEMO LINK](https://EvhenIvchenko.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,107 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
 
 export const App = () => {
+  const initialTodo = {
+    id: 0,
+    title: '',
+    completed: false,
+    userId: 0,
+  };
+
+  const [wasSubmitted, setWasSubmitted] = useState(false);
+  const [todos, setTodos] = useState(todosFromServer);
+  const [newTodo, setNewTodo] = useState(initialTodo);
+
+  const hasTitleError = wasSubmitted && newTodo.title === '';
+  const hasUserError = wasSubmitted && newTodo.userId === 0;
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    if (hasTitleError || hasUserError) {
+      return;
+    }
+
+    setTodos([...todos, newTodo]);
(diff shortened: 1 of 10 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,33 @@
...
+  const nextId = maxId + 1;
+  const user = findUsedById(todo);
+
+  return (
+    <article
+      data-id={nextId}
+      className={cn('TodoInfo', {
+        'TodoInfo--completed': todo.completed,
+      })}
+      key={todo.id}
```
TarasHoliuk, line 24: you've added `key` to TodoInfo component (right place) so no need to do it here

src/services/findUsedById.js
```diff
@@ -0,0 +1,5 @@
+import users from '../api/users';
+
+export const findUsedById = (todo) => {
```
TarasHoliuk, line 3:
```suggestion
export const findUsedById = (userId) => {
```
such functions should accept `id` as an argument

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,33 @@
...
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+import todosFromServer from '../../api/todos';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const maxId = Math.max(...todosFromServer
```
TarasHoliuk, line 12: `todosFromServer` is a static variable, so there will be calculated the same id for each new todo
you should calculate it when you add a new one, based on the actual todos array, here just set `data-id={todo.id}`

src/App.tsx
```diff
@@ -1,61 +1,107 @@
...
+    completed: false,
+    userId: 0,
+  };
+
+  const [wasSubmitted, setWasSubmitted] = useState(false);
+  const [todos, setTodos] = useState(todosFromServer);
+  const [newTodo, setNewTodo] = useState(initialTodo);
+
+  const hasTitleError = wasSubmitted && newTodo.title === '';
+  const hasUserError = wasSubmitted && newTodo.userId === 0;
```
TarasHoliuk, line 21: just check it inside `handleSubmit` and get rid of `wasSubmitted` state
also, trim `title` during validation to prevent adding title with spaces only

TarasHoliuk (changes requested):

VitaliyBondarenko1982 (approved):
Looks good

PR #2107, opened 2023-10-21
Develop

xnquaint:
[DEMO LINK](https://xnquaint.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,1 +1,32 @@
-export const TodoInfo = () => {};
+import cn from 'classnames';
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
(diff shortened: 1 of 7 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,32 @@
...
+      data-id={id}
+      className={cn('TodoInfo', {
+        'TodoInfo--completed': completed,
+      })}
+    >
+      <h2 className="TodoInfo__title">
+        {title}
+      </h2>
+
+      <UserInfo user={user} />
```
etojeDenys, line 29: let's check if user exists here
```suggestion
     {user && <UserInfo user={user} />}
```
that component doesn't need to know that user can be null, its goal is to just output the user

etojeDenys (changes requested):

VitaliyBondarenko1982 (approved):
Well done

PR #2108, opened 2023-10-22
`add task solution`

mpohorenyi:
[DEMO LINK](https://mpohorenyi.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,39 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { Todo } from './types/Todo';
+import { getUserById } from './services/user';
+import todosFromServer from './api/todos';
 
-export const App = () => {
+import { TodoForm } from './components/TodoForm';
+import { TodoList } from './components/TodoList';
+
+const initialTodos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
+
+function getNewTodoId(todos: Todo[]) {
+  return Math.max(...todos.map(todo => todo.id)) + 1;
+}
+
+export const App: React.FC = () => {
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,61 +1,39 @@
...
+
+const initialTodos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
+
+function getNewTodoId(todos: Todo[]) {
+  return Math.max(...todos.map(todo => todo.id)) + 1;
+}
+
```
sintax1s, line 19: In my opinion, it would be better if this function was in the file “TodoForm.tsx”, so as not to put a zero id, but yes, you will have to change a little the logic of your array with todo. All this is subjective, so I approve.

sintax1s (approved):
Good job.I like your idea about a new file for form.

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+    setHasTitleError(!title);
+    setHasUserIdError(!userId);
+
+    if (!title || !userId) {
+      return;
+    }
+
+    onSubmit({
+      id: 0,
```
polosanya, line 38: Why id is hardcoded here?

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,13 @@
...
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <section className="TodoList">
+    {todos.map(todo => <TodoInfo todo={todo} key={todo.id} />)}
```
polosanya, line 11:
```suggestion
    {todos.map(todo => (
      <TodoInfo todo={todo} key={todo.id} />
    ))}
```

polosanya (approved):
Good job!

PR #2112, opened 2023-10-22
Solution + validation

YaroslavYarynych:
[DEMO LINK](https://YaroslavYarynych.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,123 @@
+/* eslint-disable no-useless-return */
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import todosFromServer from './api/todos';
 
-export const App = () => {
+import { getUserById } from './services/user';
+
+import usersFromServer from './api/users';
+
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/Todo';
+
+export const preparedTodos: Todo[] = todosFromServer.map((todo) => (
+  {
+    ...todo,
+    user: getUserById(todo.userId),
+  }
+));
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [hasTitleError, setHasTitleError] = useState(false);
+  const [hasUserError, setHasUserError] = useState(false);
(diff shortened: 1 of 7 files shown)


ElenaGrasovskaya (approved):
Excellent!

superpooperxxx (changes requested):
Let's outsource the logic of the form into separate component. App.tsx should only be responsible for rendering Form, TodoList and managing the todos state

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,115 @@
+import React, { useState } from 'react';
+import usersFromServer from '../../api/users';
+import { getUserById } from '../../services/user';
+import { Todo } from '../../types/Todo';
+import { User } from '../../types/User';
+
+type Props = {
+  setNewTodos: (val: any) => void;
```
vasylmiskiv, line 8: Should add type instead of `any`

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+    setHasTitleError(!title);
+    setHasUserError(!userId);
+
+    if (!title || !userId) {
+      return;
+    }
+
+    const addTodo = (newPreparedTodo: Todo) => {
+      setNewTodos((currentTodos: any) => [
```
vasylmiskiv, line 36: Should add type instead of `any`

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,115 @@
...
+        ...currentTodos,
+        { ...newPreparedTodo },
```
vasylmiskiv, line 38:
```suggestion
        newPreparedTodo,
```

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,115 @@
...
+      ]);
+    };
+
+    const newPreparedTodo: Todo = {
+      id: Math.max(...newTodos.map(t => t.id)) + 1,
```
vasylmiskiv, line 43: Would be better to move it to separate function instead of writing it in the object directly

vasylmiskiv (approved):
Well done but pay attention to the comments

PR #2113, opened 2023-10-23
add solution todo form

Kuper18:
[DEMO](https://kuper18.github.io/react_add-todo-form/)
Hi, here is my solution

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,131 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import React, { useState } from 'react';
 
-export const App = () => {
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/Todo';
+
+function getUser(userId: number) {
+  return usersFromServer.find(user => user.id === userId) || null;
+}
+
+function getNewTodoId(todos: Todo[]) {
+  const maxId = Math.max(...todos.map(todo => todo.id));
+
+  return maxId + 1;
+}
+
+const todos: Todo[] = todosFromServer.map(todo => {
+  return {
+    ...todo,
+    user: getUser(todo.userId),
+  };
+});
+
+export const App: React.FC = () => {
+  const [initialTodo, setInitialTodo] = useState<Todo[]>(todos);
+
+  const [userId, setUserId] = useState(0);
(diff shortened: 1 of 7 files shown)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,12 @@
...
+import { User } from '../../types/User';
+
+type Props = {
+  user: User | null;
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => (
+  <a className="UserInfo" href={`mailto:${user?.email}`}>
+    {user?.name}
+  </a>
```
YevhenKharko, line 11: Seems OK, but I in my opinion it's better to check **user** before, in one place

YevhenKharko (approved):
Good job!

src/App.tsx
```diff
@@ -1,25 +1,131 @@
...
+
+const todos: Todo[] = todosFromServer.map(todo => {
+  return {
+    ...todo,
+    user: getUser(todo.userId),
+  };
+});
+
+export const App: React.FC = () => {
+  const [initialTodo, setInitialTodo] = useState<Todo[]>(todos);
```
polosanya, line 28:
```suggestion
  const [todos, setTodos] = useState<Todo[]>(todos);
```
The name is a bit confusing as it's an array of todos, not singular. Also, they are not initial, new todos will be added to this state too

src/App.tsx
```diff
@@ -1,25 +1,131 @@
...
+
+  const [userId, setUserId] = useState(0);
+  const [selectErrorMessage, setSelectErrorMessage] = useState(false);
```
polosanya, line 31: Bad naming here too. It's not clear that the value is boolean. Change the name or use empty string as the value

src/App.tsx
```diff
@@ -1,25 +1,131 @@
...
+            type="text"
+            value={title}
+            placeholder="Enter a title"
+            data-cy="titleInput"
+            onChange={handleTitleChange}
+          />
+
+          {titleErrorMessage && (
+            <span className="error">
+              Please enter a title
```
polosanya, line 103: You can save the error text in state and use here instead of the hardcoded value

src/components/UserInfo/UserInfo.tsx
polosanya: Just make the destructuring of the user

polosanya (changes requested):
Nice! Let's improve your code a bit)

superpooperxxx (approved):
🔥

PR #2116, opened 2023-10-23
Todo form with user selection

ElenaGrasovskaya:
[DEMO LINK](https://ElenaGrasovskaya.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,97 @@
-import './App.scss';
+import "./App.scss";
+import { useState } from "react";
+import usersFromServer from "./api/users";
+import todosFromServer from "./api/todos";
+import { TodoList } from "./components/TodoList";
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+let isEmptyTitle = false;
+let isEmptySelectedUser = false;
+const users = [...usersFromServer];
+let todos = [...todosFromServer].map((todo) => ({
+  ...todo,
+  user: users.find((user) => user.id === todo.userId),
+}));
+const maxId = todos.map((todo) => todo.id).sort((a, b) => b - a)[0];
 
 export const App = () => {
+  const [newTitle, setNewTitle] = useState("");
+  const [selectedUser, setSelectedUser] = useState(0);
+  const [latstTaskId, setLastTaskId] = useState(maxId);
+  const [afterSubmit, setAfterSubmit] = useState(true);
+
+  if (afterSubmit) {
+    isEmptyTitle = false;
+    isEmptySelectedUser = false;
+  } else {
+    isEmptyTitle = !newTitle;
+    isEmptySelectedUser = !selectedUser;
+  }
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+    if (!selectedUser || !newTitle) {
+      setAfterSubmit(false);
+      return;
+    }
+
+    todos = [
+      ...todos,
+      {
+        id: latstTaskId + 1,
+        title: newTitle,
+        completed: false,
+        userId: selectedUser,
+        user: users.find((user) => user.id === selectedUser),
+      },
+    ];
+    setAfterSubmit(true);
+    setLastTaskId(latstTaskId + 1);
+    setSelectedUser(0);
+    setNewTitle("");
+  };
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/todos" method="POST">
+      <form action="/api/todos" method="POST" onSubmit={handleSubmit}>
         <div className="field">
-          <input type="text" data-cy="titleInput" />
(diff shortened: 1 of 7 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,24 @@
...
+interface Props {
+  todo: Todo;
+}
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  return (
+    <article
+      data-id={todo.id}
+      className={
+        'TodoInfo ' +
```
ivaZaiets, line 14: You can use smth like that: 
```js
className={cn('TodoInfo', {
        })}
```

src/App.tsx
```diff
@@ -1,61 +1,97 @@
...
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+let isEmptyTitle = false;
+let isEmptySelectedUser = false;
+const users = [...usersFromServer];
+let todos = [...todosFromServer].map((todo) => ({
+  ...todo,
+  user: users.find((user) => user.id === todo.userId),
+}));
+const maxId = todos.map((todo) => todo.id).sort((a, b) => b - a)[0];
```
polosanya, line 14: Just use Math.max() instead
https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Math/max

src/App.tsx
```diff
@@ -1,61 +1,97 @@
...
+  const [selectedUser, setSelectedUser] = useState(0);
+  const [latstTaskId, setLastTaskId] = useState(maxId);
+  const [afterSubmit, setAfterSubmit] = useState(true);
+
+  if (afterSubmit) {
+    isEmptyTitle = false;
+    isEmptySelectedUser = false;
+  } else {
+    isEmptyTitle = !newTitle;
+    isEmptySelectedUser = !selectedUser;
```
polosanya, line 27: It should be inside some function, not just inside component

src/App.tsx
```diff
@@ -1,61 +1,97 @@
...
+let todos = [...todosFromServer].map((todo) => ({
+  ...todo,
+  user: users.find((user) => user.id === todo.userId),
+}));
+const maxId = todos.map((todo) => todo.id).sort((a, b) => b - a)[0];
 
 export const App = () => {
+  const [newTitle, setNewTitle] = useState("");
+  const [selectedUser, setSelectedUser] = useState(0);
+  const [latstTaskId, setLastTaskId] = useState(maxId);
```
polosanya, line 19: There is no need to store this value in state

src/App.tsx
```diff
@@ -1,61 +1,97 @@
...
+  const [afterSubmit, setAfterSubmit] = useState(true);
```
polosanya, line 20: Also, the usage of this state is also not clear. The name is confusing and doesn't give any info

src/App.tsx
```diff
@@ -1,61 +1,97 @@
...
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+    if (!selectedUser || !newTitle) {
+      setAfterSubmit(false);
+      return;
+    }
+
+    todos = [
+      ...todos,
```
polosanya, line 38: Better keep todos and state and update them correctly, not just reassign

src/components/TodoInfo/TodoInfo.tsx
polosanya: ➕

src/types/Todo.ts
```diff
@@ -0,0 +1,9 @@
+import User from "./User";
+
+export default interface Todo {
+    id: number,
+    title: string,
+    completed: boolean,
+    userId: number,
+    user: User | undefined,  
+}
```
polosanya, line 9: Newline is expected at the end

polosanya (changes requested):
Nice! Let's try to simplify your solution a bit)

src/App.tsx
ElenaGrasovskaya: Done

src/App.tsx
ElenaGrasovskaya: rebuilt this logic entirely

src/App.tsx
ElenaGrasovskaya: don't store the maxId anymore

src/App.tsx
ElenaGrasovskaya: Changed it slightly

src/App.tsx
ElenaGrasovskaya: Store todos in state

src/types/Todo.ts
ElenaGrasovskaya: fixed that

polosanya (approved):
Nice!

PR #2118, opened 2023-10-23
implemented task solution

YevhenKharko:
[DEMO LINK](https://yevhenkharko.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -18,7 +18,7 @@ const preparedTodos = todosFromServer.map(todo => {
 
   return {
     ...todo,
-    user: user,
+    user,
   };
 });
 
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -7,21 +7,18 @@ interface Props {
   todo: Todo,
 }
 
-export const TodoInfo: FC<Props> = ({ todo }) => {
-  console.log(todo);
-  return (
-    <article
-      data-id={todo.id}
-      className={cn('TodoInfo', {
-        'TodoInfo--completed': todo.completed,
-      })}
-    >
+export const TodoInfo: FC<Props> = ({ todo }) => (
+  <article
+    data-id={todo.id}
+    className={cn('TodoInfo', {
+      'TodoInfo--completed': todo.completed,
+    })}
+  >
 
-      <h2 className="TodoInfo__title">
-        {todo.title}
-      </h2>
+    <h2 className="TodoInfo__title">
+      {todo.title}
+    </h2>
 
-      <UserInfo user={todo.user} />
-    </article>
-  );
-};
+    <UserInfo user={todo.user} />
+  </article>
+);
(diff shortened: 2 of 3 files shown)


src/App.tsx
```diff
@@ -1,61 +1,137 @@
...
+    }
+
+    if (!currentUserId) {
+      setHasCurrentUserIdError(!hasCurrentUserIdError);
+
+      return;
+    }
+
+    const newPost = {
+      id: maxId + 1,
```
OlegPopovych, line 69: You can calculate a new `id` on the same time when you're creating new object, according to the checklist. 
```suggestion
      id:  (todos.map((todo) => todo.id).sort((a, b) => b - a)[0] + 1);
```

OlegPopovych (approved):
Good job!

src/App.tsx
```diff
@@ -1,61 +1,137 @@
...
+export const App: FC = () => {
+  const [todos, setTodos] = useState(preparedTodos);
+
+  const [title, setTitle] = useState('');
+  const [hasTitleError, setHasTitleError] = useState(false);
+
+  const [currentUserId, setCurrentUserId] = useState(0);
+  const [hasCurrentUserIdError, setHasCurrentUserIdError] = useState(false);
+
+  const maxId = todos.map((todo) => todo.id).sort((a, b) => b - a)[0];
```
vasylmiskiv, line 34: Instead of it you could use `reduce` method or` Math.max` to find a max id of the todo cause it more clear and readable

src/App.tsx
```diff
@@ -1,61 +1,137 @@
...
+
+  const handleUserSelectChange = (e: React.ChangeEvent<HTMLSelectElement>) => {
+    setCurrentUserId(+e.target.value);
+    setHasCurrentUserIdError(false);
+  };
+
+  const handleOnInput = ((e: React.ChangeEvent<HTMLInputElement>) => {
```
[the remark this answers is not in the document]
vasylmiskiv: Actually this is not really necessary as it is clearly an event. Both options are correct and depend on personal preference

src/App.tsx
```diff
@@ -1,61 +1,137 @@
...
-      <form action="/api/todos" method="POST">
+      <form
+        action="/api/todos"
+        method="POST"
+        onSubmit={handleOnSubmit}
+      >
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <input
```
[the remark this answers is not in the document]
vasylmiskiv: In this case, the label is not needed because there is no additional text before the input to wrap the input or add the htmlFor attribute

src/App.tsx
```diff
@@ -1,61 +1,137 @@
...
+                value={user.id}
+                key={user.id}
+              >
+                {user.name}
+              </option>
+            ))}
           </select>
 
-          <span className="error">Please choose a user</span>
+          {hasCurrentUserIdError
```
vasylmiskiv, line 125:
```suggestion
          { hasCurrentUserIdError
```

vasylmiskiv (approved):
Well done

PR #2124, opened 2023-10-23
task solution

dolphinnq:
[DEMO LINK](https://dolphinnq.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,143 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { TodoWithUser } from './types/Todo';
+
+const todosWithUsers: TodoWithUser[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: usersFromServer.find(user => user.id === todo.userId) || null,
+}));
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [titleError, setTitleError] = useState(false);
+  const [titlePatternError, setTitlePatternError] = useState(false);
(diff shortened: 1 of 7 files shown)


dsfreedom (approved):
Great job. You only need to add labels)

src/App.tsx
```diff
@@ -1,61 +1,151 @@
...
+
+const todosWithUsers: TodoWithUser[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: usersFromServer.find(user => user.id === todo.userId) || null,
+}));
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [titleError, setTitleError] = useState(false);
+  const [titlePatternError, setTitlePatternError] = useState(false);
```
polosanya, line 17: image

src/App.tsx
```diff
@@ -1,61 +1,151 @@
...
+    }
+
+    if (!selectedUserId) {
+      setSelectedUserIdError(true);
+
+      return;
+    }
+
+    const newPost = {
+      id: (Math.max(...posts.map(todo => todo.id)) + 1),
```
polosanya, line 76: Would be great to create a variable\function for that value

src/App.tsx
```diff
@@ -1,61 +1,151 @@
...
+      title,
+      completed: false,
+      userId: selectedUserId,
+      user: usersFromServer.find(user => user.id === selectedUserId) || null,
```
polosanya, line 80: Same here

src/App.tsx
```diff
@@ -1,61 +1,151 @@
...
+
+          {titleError && (
+            <span className="error">
+              Please enter a title
+            </span>
+          )}
+
+          {titlePatternError && (
+            <span className="error">
+              Only ua and en letters, digits, and spaces are allowed
```
polosanya, line 114: You could save the error text in the state and use here instead of the boolean value

PR #2127, opened 2023-10-23
first solution

Clover247:
[DEMO LINK](https://clover247.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,110 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-export const App = () => {
+import { Todo } from './types/Todo';
+
+import { TodoList } from './components/TodoList';
+
+const findUserById = (id: number) => {
+  return usersFromServer.find(user => user.id === id) || null;
+};
+
+const todosWithUsers = todosFromServer.map(todo => ({
+  ...todo,
+  user: findUserById(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+
+  const [hasTitleError, setHasTitleError] = useState(false);
+  const [hasUserError, setHasUserError] = useState(false);
+
+  const [newTodos, setNewTodos] = useState<Todo[]>(todosWithUsers);
+
+  const handleTitleInput = ((e: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(e.target.value);
+    setHasTitleError(false);
+  });
+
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { Todo } from './types/Todo';
+
+import { TodoList } from './components/TodoList';
+
+const findUserById = (id: number) => {
+  return usersFromServer.find(user => user.id === id) || null;
+};
```
BetterSol, line 13: It's better to relocate this function to a separate file.

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
           </select>
 
-          <span className="error">Please choose a user</span>
+          {hasUserError && <span className="error">Please choose a user</span>}
         </div>
 
         <button type="submit" data-cy="submitButton">
           Add
         </button>
       </form>
```
BetterSol, line 117: Consider creating a separate file for a ToDoForm and it's hooks, it will keep your App file more readable

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+      setHasTitleError(!hasTitleError);
+
+      return;
+    }
+
+    if (!userId) {
+      setHasUserError(!hasUserError);
+
+      return;
+    }
```
BetterSol, line 64: you actually don't need these

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+
+  const resetForm = () => {
+    setTitle('');
+    setUserId(0);
+  };
+
+  const handleSubmit = ((e: React.ChangeEvent<HTMLFormElement>) => {
+    e.preventDefault();
+
+    if (!title && !userId) {
```
BetterSol, line 47: better to use `||` operator,  then you won't need separate conditions below

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,25 @@
...
+
+interface Props {
+  todo: Todo,
+}
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => (
+  <article
+    data-id={todo.id}
+    className={cn('TodoInfo', {
+      'TodoInfo--completed': todo.completed,
```
BetterSol, line 15: Guess it's better to move this part to a new variable

BetterSol (changes requested):
Good job! Made a few suggestions about your code.

src/App.tsx
Clover247: Then I will always have both mistakes

src/App.tsx
Clover247: I need this to prevent changes and raise individual errors if there are any

polosanya (approved):
Good job!

PR #2136, opened 2023-10-25
add task

VladyslavKolisnyk:
DEMO LINK: https://vladyslavkolisnyk.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,29 +1,29 @@
 import './App.scss';
 import { useState } from 'react';
-import todosFromServer from './api/todos';
+import todos from './api/todos';
 import { TodoList } from './components/TodoList';
 import { Todo } from './interface/Todo';
 import { Form } from './components/Form/Form';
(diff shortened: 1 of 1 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,23 @@
...
+    <article
+      data-id={todo.id}
+      className={cn('TodoInfo',
+        { ' TodoInfo--completed': todo.completed })}
+    >
+      <h2 className="TodoInfo__title">
+        {todo.title}
+      </h2>
+
+      <UserInfo user={todo.user} />
```
etojeDenys, line 20: let's check here if user exists, UserInfo goal is to render the user information, that component doesn't need to know if the user exists or not

etojeDenys (changes requested):
I can create the todo only with spaces
image
let's fix that

lerastarynets (approved):
lgtm 🚀

PR #2140, opened 2023-10-25
Develop

atrofymovych:
[DEMO LINK](https://atrofymovych.github.io/react_add-todo-form/)


etojeDenys (changes requested):
image
let's fix

etojeDenys (approved):
good job

PR #2141, opened 2023-10-25
react add todo form

whiovt:
[DEMO LINK](https://whiovt.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,123 @@
+import React, { useState } from 'react';
 import './App.scss';
 
 // import usersFromServer from './api/users';
 // import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-export const App = () => {
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/Todo';
+
+function getUser(userId: number) {
+  const findUser = usersFromServer.find(user => user.id === userId);
+
+  return findUser || null;
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [titleErrorMessage, setTitleErrorMessage] = useState('');
+  const [userErrorMessage, setUserErrorMessage] = useState('');
+  const [allTodos, setAllTodos] = useState(todos);
+
+  const handleTitleChange: React.ChangeEventHandler<HTMLInputElement>
+    = (event) => {
+      event.preventDefault();
+
+      setTitleErrorMessage('');
+
+      return setTitle(event.target.value);
+    };
+
+  const handleUserChange: React.ChangeEventHandler<HTMLSelectElement>
+    = (event) => {
+      event.preventDefault();
+
+      setUserErrorMessage('');
+
+      return setUserId(+event.target.value);
+    };
+
+  const handleSubmit: React.FormEventHandler = (event) => {
+    event.preventDefault();
+
+    if (!title) {
+      setTitleErrorMessage('Please enter a title');
+    }
+
+    if (userId === 0) {
+      setUserErrorMessage('Please choose a user');
+    }
+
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,131 @@
+import React, { useState } from 'react';
 import './App.scss';
 
 // import usersFromServer from './api/users';
 // import todosFromServer from './api/todos';
```
anastasiia-tilikina, line 5: delete comments

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/Todo';
+
+function getUser(userId: number) {
+  const findUser = usersFromServer.find(user => user.id === userId);
+
+  return findUser || null;
```
anastasiia-tilikina, line 15:
```suggestion
  return usersFromServer.find(user => user.id === userId) || null;
```

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+  const [userErrorMessage, setUserErrorMessage] = useState('');
+  const [allTodos, setAllTodos] = useState(todos);
+
+  const handleTitleChange: React.ChangeEventHandler<HTMLInputElement>
+    = (event) => {
+      event.preventDefault();
+
+      setTitleErrorMessage('');
+
+      return setTitle(event.target.value);
```
anastasiia-tilikina, line 36:
```suggestion
      setTitleErrorMessage('');
      setTitle(event.target.value);
```
you don't need prevent default for ChangeEvent. as well as return statement

anastasiia-tilikina: check everywhere

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+    if (userId === 0) {
+      setUserErrorMessage('Please choose a user');
+    }
+
+    const findId = Math.max(...allTodos.map(todo => todo.id));
+
+    const todo: Todo = {
+      id: (findId + 1),
+      title,
+      userId: +userId,
```
anastasiia-tilikina, line 64: userId is already a number, no need to transform it

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+      completed: false,
+      user: getUser(+userId),
+    };
+
+    if (title && userId !== 0) {
+      setAllTodos([...todos, todo]);
```
anastasiia-tilikina, line 70: if you're setting new state value based on the previous one use callback  inside setState

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+
+    const todo: Todo = {
+      id: (findId + 1),
+      title,
+      userId: +userId,
+      completed: false,
+      user: getUser(+userId),
+    };
+
+    if (title && userId !== 0) {
```
anastasiia-tilikina, line 69:
```suggestion
    if (title && userId) {
```

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+            type="text"
+            data-cy="titleInput"
+            value={title}
+            onChange={handleTitleChange}
+            placeholder="Enter a title"
+          />
+
+          <span className="error">
+            {titleErrorMessage}
+          </span>
```
anastasiia-tilikina, line 98: error message should be rendered conditionally

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,24 @@
...
+import React from 'react';
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+interface Props {
+  todo: Todo
+}
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
```
anastasiia-tilikina, line 10: consider using destructuring here

anastasiia-tilikina (changes requested):
1. Check my comments
2. Now I can add todo with only spaces in the title. fix this.
3. When I try to add second todo it replaces the previous one instead of keep all

PR #2150, opened 2023-10-28
Develop

Vovchukkul:
https://Vovchukkul.github.io/react_add-todo-form/

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,1 +1,18 @@
-export const TodoInfo = () => {};
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo:React.FC<Props> = ({ todo }) => {
+  return (
+    <article
+      className={`TodoInfo ${todo.completed ? 'TodoInfo--completed' : ''}`}
+      data-id={todo.id}
+    >
+      <h2 className="TodoInfo__title">{todo.title}</h2>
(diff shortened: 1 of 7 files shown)


etojeDenys (changes requested):
image

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,18 @@
...
+};
+
+export const TodoInfo:React.FC<Props> = ({ todo }) => {
+  return (
+    <article
+      className={`TodoInfo ${todo.completed ? 'TodoInfo--completed' : ''}`}
+      data-id={todo.id}
+    >
+      <h2 className="TodoInfo__title">{todo.title}</h2>
+      <UserInfo user={todo.user} />
```
etojeDenys, line 15: let's handle the case if user is null here, UserInfo goal is to render the user info, that component doesn't need to know that the user can be null

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,18 @@
...
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo:React.FC<Props> = ({ todo }) => {
+  return (
+    <article
+      className={`TodoInfo ${todo.completed ? 'TodoInfo--completed' : ''}`}
```
etojeDenys, line 11: let's use classnames package here

etojeDenys (changes requested):

etojeDenys (changes requested):

etojeDenys (approved):
good job

PR #2153, opened 2023-11-01
solution

luka4ok:
Please, check my solution.
[DEMO LINK](https://luka4ok.github.io/react_add-todo-form/)


etojeDenys (changes requested):
image
let's fix that

etojeDenys (approved):
good job

PR #2154, opened 2023-11-02
add solution

purpleFade:
[DEMO LINK](https://purplefade.github.io/react_add-todo-form/)

### DIFF
--- a/src/context/AppContext.tsx
+++ b/src/context/AppContext.tsx
@@ -1,0 +1,80 @@
+import React, {
+  ReactNode, createContext, useEffect, useState,
+} from 'react';
+
+import usersFromServer from '../api/users';
+import todosFromServer from '../api/todos';
+import { Todo } from '../types/Todo';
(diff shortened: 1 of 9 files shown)


src/context/AppContext.tsx
```diff
@@ -0,0 +1,80 @@
...
+    setVisibleTodos,
+    inputError,
+    setInputError,
+    users,
+    userError,
+    setUserError,
+  };
+
+  return <appContext.Provider value={state}>{children}</appContext.Provider>;
+};
```
DarkMistyRoom, line 80: consider using useState and props instead of Context. Use Context only when it's appropriate. Please read this article
https://blog.logrocket.com/pitfalls-of-overusing-react-context/

DarkMistyRoom: It is a mistake and the app works fine. But on bigger projects, it may harm the performance

DarkMistyRoom (approved):

PR #2158, opened 2023-11-03
add task solution

MishaLehotskyi:
[DEMO LINK](https://MishaLehotskyi.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -92,7 +92,7 @@ export const App = () => {
             >
               <option value="0" disabled>Choose a user</option>
               {usersFromServer.map(user => (
-                <option value={user.id}>{user.name}</option>
+                <option value={user.id} key={user.id}>{user.name}</option>
               ))}
             </select>
           </label>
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,111 @@
...
+              value={chosenUser}
+              onChange={event => {
+                setChosenUser(+event.target.value);
+                setUserErrorMessage('');
+              }}
+            >
+              <option value="0" disabled>Choose a user</option>
+              {usersFromServer.map(user => (
+                <option value={user.id}>{user.name}</option>
+              ))}
```
etojeDenys, line 96: do not forget to add key prop here

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,35 @@
...
+      data-id={id}
+      className={classNames('TodoInfo', {
+        'TodoInfo--completed': completed,
+      })}
+    >
+      <h2 className="TodoInfo__title">
+        {title}
+      </h2>
+
+      <UserInfo user={user} />
```
etojeDenys, line 32: user can be undefined, we need to handle this case here

etojeDenys (changes requested):

PR #2169, opened 2023-11-10
Added solution

FurArt:
[Visit site](https://furart.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/function/getUserNameByUserId.ts
+++ b/src/components/function/getUserNameByUserId.ts
@@ -1,0 +1,21 @@
+import { Todo, User } from '../types/types';
+
+export function getUserNameByUserId(
+  userId: number,
+  todos: Todo[],
+  users: User[],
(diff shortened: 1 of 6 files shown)


src/components/function/getUserNameByUserId.ts
```diff
@@ -0,0 +1,21 @@
...
+  userId: number,
+  todos: Todo[],
+  users: User[],
+): User | undefined {
+  const userTodo: Todo | undefined = todos.find(
+    (todo) => todo.userId === userId,
+  );
+
+  if (!userTodo) {
+    return undefined;
```
VitaliyBondarenko1982, line 13:
```suggestion
    return null;
```
when we need to render nothing, it should be returned null.
undifined also render nothing, but general practice return null.

VitaliyBondarenko1982 (approved):
Well done

PR #2174, opened 2023-11-16
react todo adder was realised

KirillMaslov:
[DEMO LINK](https://kirillmaslov.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -2,8 +2,12 @@ iframe {
   display: none;
 }
 
+* {
+  box-sizing: border-box;
+}
+
 html {
-  font-family: Arial, "Helvetica Neue", Helvetica, sans-serif;
+  font-family: "Roboto", sans-serif;
 }
 
 label,
@@ -11,8 +15,72 @@ button {
   cursor: pointer;
 }
 
+h1 {
+  margin: 0;
+}
+
+.App {
+  display: block;
+  width: fit-content;
+  margin: 0 auto;
+
+  &__title {
+    font-size: 35px;
+    letter-spacing: 1.2px;
+    text-align: center;
+
+    margin-bottom: 40px;
+    color: #000;
+  }
+
+  &__form {
+    margin-bottom: 50px;
+  }
+}
+
+.field {
+  &:not(:first-child) {
+    margin-top: 10px;
+  }
+}
+
 .error {
   color: red;
+  display: block;
+  margin-top: 1px;
+
+  &--disabled {
+    display: none;
+  }
+}
+
+.userSelect {
+  width: 50%;
+  height: 25px;
+  display: block;
+  border: 1px solid #ccc;
+  border-radius: 5px;
+  // background-color: rgba(173, 216, 230, 0.8);
+  background-color: rgb(135, 237, 140);
+  color: #333;
+  font-size: 16px;
+
+  &:focus {
+    outline: none;
+    border-color: #51e24a; // Light blue color when focused
+    box-shadow: 0 0 5px rgba(74, 144, 226, 0.7);
+  }
+}
+
+.titleInput {
+  display: block;
+  border: 1px solid black;
+  background-color: #fff;
+  border-radius: 20px;
+  height: 40px;
+  width: 100%;
+  font-size: 17px;
+  padding: 10px;
 }
 
 .TodoInfo {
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,61 +1,101 @@
...
+            {
+              id: maxId + 1,
+              title: todoTitle,
+              completed: false,
+              userId,
+            },
+          ]);
+          setTodoTitle('');
+          setUserId(0);
+        }}
```
etojeDenys, line 56: move this logic to separate function to improve readability

src/App.tsx
```diff
@@ -1,61 +1,101 @@
...
+          <select
+            data-cy="userSelect"
+            className="userSelect"
+            value={userId}
+            onChange={handleUserIdChange}
+          >
             <option value="0" disabled>Choose a user</option>
+            {usersFromServer.map((user) => (
+              <option value={user.id}>{user.name}</option>
+            ))}
```
etojeDenys, line 82: do not forget about key prop

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,22 @@
...
+  todos,
+}) => {
+  return (
+    <section className="TodoList">
+      {todos.map((todo: Todo) => {
+        return (
+          <TodoInfo
+            todo={todo}
+          />
+        );
```
etojeDenys, line 18: same here

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,36 @@
...
+      data-id={id}
+      className={cn('TodoInfo', {
+        'TodoInfo--completed': completed,
+      })}
+    >
+      <h2 className="TodoInfo__title">
+        {title}
+      </h2>
+
+      <UserInfo user={user} />
```
etojeDenys, line 33: user can be null, let's handle it here

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,36 @@
...
+export const TodoInfo: React.FC<Props> = ({
+  todo,
+}) => {
+  const {
+    id, title, completed, userId,
+  } = todo;
+
+  const user: User
+    = usersFromServer.find((us: User) => us.id === userId)
+    ?? usersFromServer[0];
```
etojeDenys, line 20:
```suggestion
    = usersFromServer.find((us: User) => us.id === userId)
   || null;
```
it's strange to set the first user if we cannot find the real one in this array

etojeDenys (changes requested):
image
i can create the empty todo, let's fix that

src/App.scss
```diff
@@ -2,17 +2,85 @@ iframe {
...
+  }
+}
+
+.userSelect {
+  width: 50%;
+  height: 25px;
+  display: block;
+  border: 1px solid #ccc;
+  border-radius: 5px;
+  // background-color: rgba(173, 216, 230, 0.8);
```
VitaliyBodnar, line 63: remove all comments

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,53 @@
...
+interface Props {
+  todo: Todo
+}
+
+export const TodoInfo: React.FC<Props> = ({
+  todo,
+}) => {
+  const {
+    id, title, completed, userId,
+  } = todo;
```
VitaliyBodnar, line 16: can be less
```suggestion
export const TodoInfo: React.FC<Props> = ({
  todo: {
    id, title, completed, userId,
  }
}) => {
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,53 @@
...
+
+  const user: User | null
+    = usersFromServer.find((us: User) => us.id === userId)
+    || null;
```
VitaliyBodnar, line 20: null is redundant cause method filter always return a array
```suggestion
  const user: User = usersFromServer.find((us: User) => us.id === userId)
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,53 @@
...
+
+  if (user !== null) {
```
VitaliyBodnar, line 22:
```suggestion
  if (user) {
```

VitaliyBodnar (changes requested):

VitaliyBodnar (approved):

PR #2181, opened 2023-11-20
React. Add todo form

khviktoriia:
[DEMO LINK](https://khviktoriia.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,1 +1,28 @@
-export const TodoInfo = () => {};
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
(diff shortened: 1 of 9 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,28 @@
...
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    completed, title, user, id,
+  } = todo;
+
+  return (
+    <article
+      className={`TodoInfo ${
+        completed ? 'TodoInfo--completed' : ''
```
etojeDenys, line 17: we usually use classnames package for conditional classes

etojeDenys (changes requested):
image

etojeDenys (approved):
awesome

PR #2186, opened 2023-11-22
Develop

ArtemSydun:
[DEMO](https://artemsydun.github.io/react_add-todo-form/)


src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,91 @@
...
+      return;
+    }
+
+    onSubmit({
+      title,
+      completed: false,
+      userId: UserId,
+    });
+    setTitle('');
+    setUserId(0);
```
polosanya, line 32: Would be great to create a separate function for reset

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,91 @@
...
+type Props = {
+  onSubmit: (todo: Todo) => void;
+  users: User[];
+};
+
+export const TodoForm: React.FC<Props> = ({ onSubmit, users }) => {
+  const [title, setTitle] = useState('');
+  const [UserId, setUserId] = useState(0);
+  const [hasTitleError, setHasTitleError] = useState(false);
+  const [hasUserError, setHasUserError] = useState(false);
```
polosanya, line 14: You can also create a state for an error message instead of the boolean and check if it's not empty

polosanya (approved):
Good job!

src/components/TodoForm/TodoForm.tsx
ArtemSydun: smart

PR #2189, opened 2023-11-25
solution

yevtok:
[DEMO LINK](https://yevtok.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,41 @@
+import { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import todosFromServer from './api/todos';
+import { getUserById } from './services/user';
+import { Todo } from './types/Todo';
+import { TodoForm } from './components/TodoForm/TodoForm';
+import { TodoList } from './components/TodoList';
+
+const initialTodos: Todo[] = todosFromServer.map((todo) => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
+
+function getNewTodoId(todos: Todo[]) {
+  const maxId = Math.max(...todos.map((todo) => todo.id));
+
+  return maxId + 1;
+}
 
 export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>(initialTodos);
+
+  const addTodo = (todo: Todo) => {
+    const newTodo = {
+      ...todo,
+      id: getNewTodoId(todos),
+    };
+
+    setTodos((currentTodos) => [...currentTodos, newTodo]);
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/todos" method="POST">
-        <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
-        </div>
-
-        <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>Choose a user</option>
-          </select>
-
-          <span className="error">Please choose a user</span>
-        </div>
-
-        <button type="submit" data-cy="submitButton">
-          Add
-        </button>
-      </form>
-
-      <section className="TodoList">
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,61 +1,41 @@
...
+import { Todo } from './types/Todo';
+import { TodoForm } from './components/TodoForm/TodoForm';
+import { TodoList } from './components/TodoList';
+
+const initialTodos: Todo[] = todosFromServer.map((todo) => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
+
+function getNewTodoId(todos: Todo[]) {
```
artemforsoff, line 15: the code uses two types of function declaration, it is better to choose one so that the code is consistent

src/App.tsx
```diff
@@ -1,61 +1,41 @@
+import { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import todosFromServer from './api/todos';
+import { getUserById } from './services/user';
+import { Todo } from './types/Todo';
```
artemforsoff, line 6: in order to immediately clearly see that a type or component is being imported,
we can explicitly indicate that this is a type
`import type { Todo } from './types/Todo';`

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,99 @@
+import React, { useState } from 'react';
+import usersFromServer from '../../api/users';
+import { User } from '../../types/User';
+import { Todo } from '../../types/Todo';
+import { getUserById } from '../../services/user';
+
+type Props = {
```
artemforsoff, line 7: if this type is needed for export, then it’s better to let us know whose props they are
like this `TodoFormProps`

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const reset = () => {
+    setTitle('');
+    setUserId(0);
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    setHasTitleErrors(!title.trim());
+    setHasUserIdErrors(!userId);
```
artemforsoff, line 37: It is better to move the validation logic into a separate function

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+};
+
+export const TodoForm: React.FC<Props> = ({ onSubmit }) => {
+  const [title, setTitle] = useState('');
+  const [hasTitleErrors, setHasTitleErrors] = useState(false);
+
+  const [userId, setUserId] = useState(0);
+  const [hasUserIdErrors, setHasUserIdErrors] = useState(false);
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
artemforsoff, line 18: the prefix `handle` doesn't say anything about the function,
so the prefix is not required

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,26 @@
...
+  return (
+    <article
+      className={classNames('TodoInfo', {
+        'TodoInfo--completed': completed,
+      })}
+      data-id={id}
+    >
+      <h2 className="TodoInfo__title">{title}</h2>
+
+      {!!user && <UserInfo user={user} />}
```
artemforsoff, line 23: [Operator Logical AND](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND) convert to boolean value
so you can write shorter `{user && <UserInfo user={user} />}`

src/services/user.ts
artemforsoff, line 1: I think that one function is not a service and a more appropriate name for the directory would be `shared/lib`

artemforsoff (approved):
overall great job 🔥
I left some minor comments, please pay attention to them

PR #2196, opened 2023-11-29
Develop

Daniil-Lilin:
[demo link](https://Daniil-Lilin.github.io/react_add-todo-form/)


nazarmatsevych (changes requested):
Almost done, please check comments:

I can add todo with empty title, use trim() method to fix it
image

etojeDenys (commented):
i can create the empty todo, let's fix that
image
still not fixed(

etojeDenys (changes requested):
.

PR #2197, opened 2023-11-29
Develop

vanya-kalyenichenko:
[DEMO LINK](https://vanya-kalyenichenko.github.io/react_add-todo-form/)


src/types/TodosUsers.tsx
```diff
@@ -0,0 +1,12 @@
+export type Todos = {
```
etojeDenys, line 1:
```suggestion
export type Todo = {
```
it's one todo

src/types/TodosUsers.tsx
```diff
@@ -0,0 +1,12 @@
...
+  id: number;
+  title: string;
+  completed: boolean;
+  userId: number;
+  user: {
+    id: number;
+    name: string;
+    username: string;
+    email: string;
```
etojeDenys, line 10: let's create the user type also, and reuse here

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #2202, opened 2023-11-30
Develop

MrHunterBoi:
[DEMO LINK](https://MrHunterBoi.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,7 +3,7 @@ import './App.scss';
 import { useState } from 'react';
 import todosFromServer from './api/todos';
 import usersFromServer from './api/users';
-import { TodoForm } from './components/TodoForm';
+import { TodoForm } from './components/TodoForm/TodoForm';
 import { TodoList } from './components/TodoList';
 
 export const App = () => {
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,99 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { useState } from 'react';
+import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import { TodoForm } from './components/TodoForm';
+import { TodoList } from './components/TodoList';
 
 export const App = () => {
+  const [todos, setTodos] = useState(todosFromServer);
+  const [firstOptionDisabled, setFirstOptionDisabled] = useState(false);
+  const [formInputs, setFormInputs] = useState({
+    title: '',
+    userId: 0,
+  });
+  const [titleErrorMessage, setTitleErrorMessage] = useState('');
+  const [userErrorMessage, setUserErrorMessage] = useState('');
+
+  const todosWithUsers = todos.map(todo => ({
+    ...todo,
+    user: usersFromServer.find(u => u.id === todo.userId),
+  }));
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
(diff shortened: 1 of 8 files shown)


MrHunterBoi:
image
👍

src/App.tsx
```diff
@@ -1,61 +1,99 @@
...
+  const [formInputs, setFormInputs] = useState({
+    title: '',
+    userId: 0,
+  });
+  const [titleErrorMessage, setTitleErrorMessage] = useState('');
+  const [userErrorMessage, setUserErrorMessage] = useState('');
+
+  const todosWithUsers = todos.map(todo => ({
+    ...todo,
+    user: usersFromServer.find(u => u.id === todo.userId),
```
ArtemTeslenko, line 21: [checklist](https://github.com/mate-academy/react_add-todo-form/blob/master/checklist.md) p. 1

src/App.tsx
```diff
@@ -1,61 +1,99 @@
...
 
 export const App = () => {
+  const [todos, setTodos] = useState(todosFromServer);
+  const [firstOptionDisabled, setFirstOptionDisabled] = useState(false);
+  const [formInputs, setFormInputs] = useState({
+    title: '',
+    userId: 0,
+  });
+  const [titleErrorMessage, setTitleErrorMessage] = useState('');
+  const [userErrorMessage, setUserErrorMessage] = useState('');
```
ArtemTeslenko, line 17: imagine if you have form with 20 fields, which you need to validate, will you initiate 20 states for them? maybe better create one state which will return boolean depending on any error existing and then if this state returns `true` check where is this error?

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,64 @@
...
+    userId: number;
+  };
+  titleErrorMessage: string;
+  firstOptionDisabled: boolean;
+  userErrorMessage: string;
+}
+
+export function TodoForm(props: Props) {
+  return (
+    <form onSubmit={props.handleSubmit}>
```
ArtemTeslenko, line 19: can use destructuring for props here

ArtemTeslenko (changes requested):
Good job 👍 See comments please)

src/components/TodoForm/TodoForm.tsx
MrHunterBoi: Lots of destructuring to do, but fair

ksyvvkaa (approved):
good job🤙

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  formInputs: {
+    title: string;
+    userId: number;
+  };
+  firstOptionDisabled: boolean;
+  errorStatuses: {
+    title: boolean,
+    user: boolean,
+  };
```
BudnikOleksii, line 19: For `formInputs` and `errorStatuses` it's better to create interfaces as you set them to state in the `App` component, so they should be the same

src/interfaces/Todo.ts
```diff
@@ -0,0 +1,9 @@
+import User from './User';
+
+export default interface Todo {
+  id: number;
+  title: string;
+  completed: boolean;
+  userId: number;
+  user: User | undefined;
```
BudnikOleksii, line 8:
```suggestion
  user?: User;
```

BudnikOleksii (approved):
Good work 👍

PR #2203, opened 2023-11-30
solution

katya-sn:
https://katya-sn.github.io/react_add-todo-form/

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,1 +1,26 @@
-export const TodoInfo = () => {};
+import React from 'react';
+import { UserInfo } from '../UserInfo';
+import { Todo } from '../../types/Types';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    title, completed, id, user,
+  } = todo;
+
+  return (
+    <article
+      data-id={id}
+      className={`TodoInfo ${completed && 'TodoInfo--completed'}`}
+    >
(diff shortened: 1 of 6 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,26 @@
...
+  return (
+    <article
+      data-id={id}
+      className={`TodoInfo ${completed && 'TodoInfo--completed'}`}
+    >
+      <h2 className="TodoInfo__title">
+        {title}
+      </h2>
+
+      {user && <UserInfo user={user} />}
```
etojeDenys, line 23: if you added the check here to render component only when the user exists

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,12 @@
...
+import React from 'react';
+import { User } from '../../types/Types';
+
+type Props = {
+  user: User | null;
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => (
+  <a className="UserInfo" href={`mailto:${user?.email}`}>
+    {user?.name}
```
etojeDenys, line 10: so you can remove it
```suggestion
  user: User;
};

export const UserInfo: React.FC<Props> = ({ user }) => (
  <a className="UserInfo" href={`mailto:${user.email}`}>
    {user.name}
```

etojeDenys (changes requested):
nit:

src/components/UserInfo/UserInfo.tsx
katya-sn: exactly! I remowed that operator

etojeDenys (approved):
awesome

PR #2208, opened 2023-12-03
solution

YuriiHlushenko:
[DEMO LINK](https://YuriiHlushenko.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,1 +1,17 @@
-export const TodoInfo = () => {};
+import { UserInfo } from '../UserInfo';
+import { Todo } from '../types/Todo';
+
+type Props = { todo: Todo };
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => (
+  <article
+    data-id={todo.id}
+    className={todo.completed
+      ? 'TodoInfo TodoInfo--completed'
+      : 'TodoInfo'}
+  >
+    <h2 className="TodoInfo__title">{todo.title}</h2>
+
+    <UserInfo user={todo.user} />
+  </article>
+);
--- a/src/components/services/services.ts
+++ b/src/components/services/services.ts
@@ -1,0 +1,12 @@
+import usersFromServer from '../../api/users';
+import todosFromServer from '../../api/todos';
+
+export function getUserById(userId: number) {
+  return usersFromServer.find(user => user.id === userId)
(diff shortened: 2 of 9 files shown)


nazarmatsevych (changes requested):
I can add todo with empty title, use trim() method to fix it
image

src/components/services/services.ts
```diff
@@ -0,0 +1,12 @@
+import usersFromServer from '../../api/users';
+import todosFromServer from '../../api/todos';
+
+export function getUserById(userId: number) {
+  return usersFromServer.find(user => user.id === userId)
+    || usersFromServer[0];
```
etojeDenys, line 6: let's return null in this case

src/components/types/Todo.ts
```diff
@@ -0,0 +1,9 @@
+import { User } from './User';
+
+export interface Todo {
+  id: number,
+  title: string,
+  completed: boolean,
+  userId: number
+  user: User
```
etojeDenys, line 8:
```suggestion
  user: User | null
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,17 @@
...
+export const TodoInfo: React.FC<Props> = ({ todo }) => (
+  <article
+    data-id={todo.id}
+    className={todo.completed
+      ? 'TodoInfo TodoInfo--completed'
+      : 'TodoInfo'}
+  >
+    <h2 className="TodoInfo__title">{todo.title}</h2>
+
+    <UserInfo user={todo.user} />
```
etojeDenys, line 15: user can be null, let's handle it here

etojeDenys (changes requested):
feel free to ask me about it in the chat, if you need some help

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,30 @@
...
+      todo.user
+        ? <UserInfo user={todo.user} />
+        : (
+          <UserInfo user={{
+            id: 999,
+            name: 'Undefined User',
+            username: 'UndefinedUsername',
+            email: 'Undefined@Email.com',
+          }}
+          />
```
SanyaBratashchuk, line 25:
```suggestion
      todo.user
        && <UserInfo user={todo.user} />
```

SanyaBratashchuk (approved):

PR #2210, opened 2023-12-04
solution

OstinM:
- [DEMO LINK](https://OstinM.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -29,7 +29,7 @@ export const App: React.FC = () => {
     setHasUserIdError(false);
   };
 
-  const reset = () => {
+  const resetForm = () => {
     setTitle('');
     setHasTitleError(false);
 
@@ -61,7 +61,7 @@ export const App: React.FC = () => {
       [...currentTodos, newTodo]
     ));
 
-    reset();
+    resetForm();
   };
 
   return (
@@ -91,12 +91,15 @@ export const App: React.FC = () => {
             required
           >
             <option value="0" disabled>Choose a user</option>
-            {usersFromServer.map((user) => (
+            {usersFromServer.map(({
+              id,
+              name,
+            }) => (
               <option
-                value={user.id}
-                key={user.id}
+                value={id}
+                key={id}
               >
-                {user.name}
+                {name}
               </option>
             ))}
           </select>
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -7,8 +7,7 @@ type Props = {
   todo: Todo,
 };
 
-export const TodoInfo: React.FC<Props> = (props) => {
-  const { todo } = props;
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
   const {
     id,
     title,
--- a/src/components/UserInfo/UserInfo.tsx
+++ b/src/components/UserInfo/UserInfo.tsx
@@ -4,8 +4,7 @@ type Props = {
   user: User,
 };
 
-export const UserInfo: React.FC<Props> = (props) => {
-  const { user } = props;
+export const UserInfo: React.FC<Props> = ({ user }) => {
   const { name, email } = user;
 
   return (
--- a/src/helpers.tsx
+++ b/src/helpers.tsx
@@ -15,11 +15,7 @@ export const getMaxId = (todos: Todo[]) => {
 };
 
 export const getUser = (userId: number) => {
-  const requiredUser = usersFromServer.find(user => (
-    user.id === userId
-  ));
-
-  return requiredUser;
+  return usersFromServer.find(({ id }) => id === userId);
 };
 
 export const preparedTodoList: Todo[] = todosFromServer.map(todo => ({


src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,19 @@
-export const UserInfo = () => {};
+import { User } from '../../types';
+
+type Props = {
+  user: User,
+};
+
+export const UserInfo: React.FC<Props> = ({
+  user: {
```
mykhailoyurkovskyi, line 8: I'm not sure that it's a good idea to do it here. I think you should do destructuring in the body of the component.

const { name, email } = user;

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+  };
+
+  const handleSelect = (
+    event: React.ChangeEvent<HTMLSelectElement>,
+  ) => {
+    setSelect(+event.target.value);
+    setHasSelectError(false);
+  };
+
+  const reset = () => {
```
mykhailoyurkovskyi, line 32: resetForm😉

mykhailoyurkovskyi (approved):
Good work!✔👌

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+import { getMaxId, getUser, preparedTodoList } from './helpers';
+import { Todo } from './types';
+
+export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>(preparedTodoList);
+
+  const [title, setTitle] = useState('');
+  const [hasTitleError, setHasTitleError] = useState(false);
+
+  const [select, setSelect] = useState(0);
```
BudnikOleksii, line 15: What select is it?

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
       <form action="/api/todos" method="POST">
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <input
+            type="text"
+            data-cy="titleInput"
+            placeholder="Enter title"
+            value={title}
+            onChange={(event) => handleTitle(event)}
```
BudnikOleksii, line 78:
```suggestion
            onChange={handleTitle}
```
You can simplify

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,31 @@
...
+  todo: Todo,
+};
+
+export const TodoInfo: React.FC<Props> = ({
+  todo: {
+    id,
+    title,
+    completed = false,
+    user,
+  },
```
BudnikOleksii, line 16: As mentioned [mykhailoyurkovskyi](https://github.com/mykhailoyurkovskyi)
It's better to do destructuring in the body of the component.

`const { id, title, completed, user } = todo;`

Also, completed is a required field in the `Todo` interface so you don't need to set default value

src/components/UserInfo/UserInfo.tsx
BudnikOleksii: > I'm not sure that it's a good idea to do it here. I think you should do destructuring in the body of the component.
> 
> const { name, email } = user;

Good point

BudnikOleksii (changes requested):
Good job 👍 
Left a few suggestions and comments, check them

src/App.tsx
roman-mirzoian: @OstinM not fixed 😢

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
-          <select data-cy="userSelect">
+          <select
+            data-cy="userSelect"
+            value={userId}
+            defaultValue={0}
+            onChange={handleSelect}
+            required
+          >
             <option value="0" disabled>Choose a user</option>
+            {usersFromServer.map((user) => (
```
roman-mirzoian, line 94: Use destructuring on the `user` object.

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,34 @@
...
+
+import { Todo } from '../../types';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todo,
+};
+
+export const TodoInfo: React.FC<Props> = (props) => {
+  const { todo } = props;
```
roman-mirzoian, line 11: Yo can simplify this:
```suggestion
export const TodoInfo: React.FC<Props> = ({ todo }) => {
```

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,19 @@
-export const UserInfo = () => {};
+import { User } from '../../types';
+
+type Props = {
+  user: User,
+};
+
+export const UserInfo: React.FC<Props> = (props) => {
+  const { user } = props;
```
roman-mirzoian, line 8:
```suggestion
export const UserInfo: React.FC<Props> = ({ user }) => {
```

src/helpers.tsx
```diff
@@ -0,0 +1,28 @@
...
+  return maxId + 1;
+};
+
+export const getUser = (userId: number) => {
+  const requiredUser = usersFromServer.find(user => (
+    user.id === userId
+  ));
+
+  return requiredUser;
+};
```
roman-mirzoian, line 23:
```suggestion
export const getUser = (userId: number) => {
  return usersFromServer.find(({ id }) => id === userId);
};
```

roman-mirzoian (changes requested):
Almost done!

roman-mirzoian (approved):
Well done 👍

PR #2212, opened 2023-12-04
add task solution

ksyvvkaa:
[DEMO LINK](https://ksyvvkaa.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,55 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { useState } from 'react';
+
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { AddTodoForm } from './components/AddTodoForm';
+import { TodoList } from './components/TodoList';
+import { Todo } from './type/Todo';
+import { User } from './type/User';
+
+function getUserById(userId: number): User | null {
+  return usersFromServer.find(user => user.id === userId) || null;
+}
+
+function getNewTodoId(todos: Todo[]) {
+  const maxId = Math.max(
+    ...todos.map(todo => todo.id),
+  );
+
+  return maxId + 1;
+}
+
+const preparedTodos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
 
 export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>(preparedTodos);
+
+  const addTodo = (title: string, userId: number) => {
+    const newTodo = {
+      id: getNewTodoId(todos),
+      title,
+      completed: false,
+      userId,
+      user: getUserById(userId),
+    };
+
+    setTodos(currentTodos => [...currentTodos, newTodo]);
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/todos" method="POST">
-        <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
-        </div>
+      <AddTodoForm users={usersFromServer} onAdd={addTodo} />
 
-        <div className="field">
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,61 +1,55 @@
...
-          <a className="UserInfo" href="mailto:Julianne.OConner@kory.org">
-            Patricia Lebsack
-          </a>
-        </article>
-      </section>
+      <AddTodoForm users={usersFromServer} onAdd={addTodo} />
+
+      <TodoList
+        todos={todos}
+      />
```
polosanya, line 52:
```suggestion
      <TodoList todos={todos} />
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,19 @@
...
+interface Props {
+  todos: Todo[];
+}
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="TodoList">
+      {todos.map(todo => (
+        <Fragment key={todo.id}>
+          <TodoInfo todo={todo} />
```
polosanya, line 14: What's the purpose of Fragment here? 
You can just return TodoInfo with key

polosanya (changes requested):
Almost perfect!

Anton-Kuchmasov (approved):
Splendidly!

src/App.tsx
```diff
@@ -1,61 +1,53 @@
...
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { useState } from 'react';
+
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { AddTodoForm } from './components/AddTodoForm';
+import { TodoList } from './components/TodoList';
+import { Todo } from './type/Todo';
+import { User } from './type/User';
```
artemforsoff, line 10: this is not wrong, just an improvement
so that the developer can clearly see that the import is being imported and can be written like this
`import type { User } from './type/User';`

src/App.tsx
```diff
@@ -1,61 +1,53 @@
...
+  const addTodo = (title: string, userId: number) => {
+    const newTodo = {
+      id: getNewTodoId(todos),
+      title,
+      completed: false,
+      userId,
+      user: getUserById(userId),
+    };
+
+    setTodos(currentTodos => [...currentTodos, newTodo]);
```
artemforsoff, line 41: usually the newest entities are added to the beginning

src/components/AddTodoForm/AddTodoForm.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const filteredTitle = event.target.value
+      .replace(/[^a-zA-Z0-9\s\u0400-\u04FF]/g, '');
+
+    setTitle(filteredTitle);
+    setHasTitleError(false);
+  };
+
+  const handleUserIdChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setUserId(+event.target.value);
```
artemforsoff, line 25: to make the cat read more human-like text you can write like this:
`setUserId(Number(event.target.value));`

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,13 @@
-export const UserInfo = () => {};
+import { User } from '../../type/User';
+
+interface Props {
+  user: User | null;
```
artemforsoff, line 4: if `null` is the value then the link will be empty
therefore, it is better to require that the user always be an object
so that this component can be used like this
`{user && <UserInfo user={user} />}`
or if you still leave the value as `null` then better write this:
```
if (!user) {
  return null;
}

const { email, name } = user;

return (
  <a className="UserInfo" href={`mailto:${email}`}>
    {name}
  </a>
);
```

artemforsoff (approved):
Keep up the wonderful work! 🚀

PR #2217, opened 2023-12-05
done task add-todo-form

Anton-Kuchmasov:
[DEMO LINK]
(https://Anton-Kuchmasov.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -6,11 +6,7 @@ import usersFromServer from './api/users';
 
 import { Todo } from './types/todo';
 import { PostForm } from './components/TodoForm/TodoForm';
-
-function getUserById(userId: number) {
-  return usersFromServer.find(user => user.id === userId)
-      || null;
-}
+import { getUserById } from './utils/getUserById';
 
 const todosToRender:Todo[] = todosFromServer.map(todo => ({
   ...todo,
--- a/src/components/TodoForm/TodoForm.tsx
+++ b/src/components/TodoForm/TodoForm.tsx
@@ -1,7 +1,8 @@
 import { useState } from 'react';
 import { Todo } from '../../types/todo';
 
-import { User, getUserById } from '../../types/user';
+import { User } from '../../types/user';
+import { getUserById } from '../../utils/getUserById';
 
 type Props = {
   onSubmit: (todo: Todo) => void;
@@ -86,6 +87,7 @@ export const PostForm: React.FC<Props> = ({ onSubmit, todoId, users }) => {
 
           {users.map((user: User) => (
             <option
+              key={user.id}
               value={user.id}
             >
               {user.name}
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -4,18 +4,16 @@ import { Todo } from '../../types/todo';
 interface Props {
   todo: Todo;
 }
-export const TodoInfo = ({ todo }: Props) => (
-  <article
-    data-id={todo.id}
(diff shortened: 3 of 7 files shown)


src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,104 @@
...
+          value={selectedUserId}
+          onChange={handleUserChange}
+        >
+          <option value="0" disabled defaultValue="0">
+            Choose a user
+          </option>
+
+          {users.map((user: User) => (
+            <option
+              value={user.id}
```
polosanya, line 89: Don't forget about key attribute

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,21 @@
...
+    className={`TodoInfo ${todo.completed ? 'TodoInfo--completed' : ''}`}
+  >
+    <h2
+      className="TodoInfo__title"
+    >
+      {todo.title}
+    </h2>
+    <UserInfo
+      user={todo.user}
+    />
```
polosanya, line 19:
```suggestion
    </h2>
    
    <UserInfo user={todo.user} />
```
Also, you can destructure all the needed props of todo

src/types/user.ts
```diff
@@ -0,0 +1,13 @@
...
+  id: number;
+  name: string;
+  username: string;
+  email: string;
+};
+
+export function getUserById(userId: number) {
+  return usersFromServer.find(user => user.id === userId)
+      || null;
+}
```
polosanya, line 13: Would be great to create a separate file for functions-helpers

src/App.tsx
```diff
@@ -1,61 +1,51 @@
...
+import usersFromServer from './api/users';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { Todo } from './types/todo';
+import { PostForm } from './components/TodoForm/TodoForm';
+
+function getUserById(userId: number) {
+  return usersFromServer.find(user => user.id === userId)
+      || null;
```
polosanya, line 12: You can import it from helpers, not need to duplicate

polosanya (changes requested):
Nice! Let's improve your code a bit)

src/components/TodoInfo/TodoInfo.tsx
PonomarchukLana: fixed

src/utils/getUserById.ts
```diff
@@ -0,0 +1,6 @@
+import usersFromServer from '../api/users';
+import { User } from '../types/user';
+
+export function getUserById(userId: number) {
+  return usersFromServer.find(user => user.id === userId) as User;
```
PonomarchukLana, line 5: Don't use " as ", it's bad practice. User can be "null" and it's ok

PonomarchukLana (approved):
Cool! You add component for form! Good work! :fire:

src/utils/getUserById.ts
polosanya: Good point. Find method can possibly return undefined

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,19 @@
-export const UserInfo = () => {};
+import { User } from '../../types/user';
+
+interface Props {
+  user: User | null;
+}
```
polosanya, line 5: You shouldn't render UserInfo component if the User is null. User should be required in this component

polosanya (approved):
Good job! I'm approving it but please pay attention to the comments)

PR #2219, opened 2023-12-05
task solution

VeDono:
[DEMO LINK](https://VeDono.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,146 @@
-import './App.scss';
+import { useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import './App.scss';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { TodoList } from './components/TodoList';
+import { PreparedTodo, Todo } from './types';
+
+function getUserById(userIdRecived: number) {
+  return (
+    usersFromServer.find((currentUser) => currentUser.id === userIdRecived)
+    || null
+  );
+}
+
+function getPreparedTodos(todosRecived: Todo[]) {
+  return todosRecived.map((todo) => ({
+    ...todo,
+    user: getUserById(todo.userId),
+  }));
+}
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [userSelectedId, setUserSelectedId] = useState(0);
+
+  const [preparedTodos, setPreparedTodos] = useState<PreparedTodo[]>(
+    getPreparedTodos(todosFromServer),
+  );
+
+  const [hasTitleError, setHasTitleError] = useState(false);
+  const [hasUserError, setHasUserError] = useState(false);
+
+  const resetForm = (): void => {
+    setTitle('');
+    setUserSelectedId(0);
+  };
+
+  const getNewTodoId = (): number => {
+    const maxId = Math.max(...preparedTodos.map(preparedTodo => (
+      preparedTodo.id
+    )));
+
+    return maxId + 1;
+  };
+
+  const addNewTodo = (): void => {
+    const user = usersFromServer.find(userFromServer => (
+      userFromServer.id === userSelectedId
+    )) || null;
+
+    const newTodo = {
+      id: getNewTodoId(),
+      title,
+      completed: false,
+      userId: userSelectedId,
+      user,
+    };
+
+    setPreparedTodos(currentPreparedTodos => (
+      [...currentPreparedTodos, newTodo]
+    ));
+
+    resetForm();
+  };
+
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,154 @@
...
+
+import { TodoList } from './components/TodoList';
+import { PreparedTodo, Todo } from './types';
+
+function getUserById(userIdRecived: number) {
+  return (
+    usersFromServer.find((currentUser) => currentUser.id === userIdRecived)
+    || null
+  );
+}
```
roman-mirzoian, line 15:
```suggestion
function getUserById(userId: number) {
  return usersFromServer.find(({ id }) => id === userId);
}
```

src/App.tsx
```diff
@@ -1,61 +1,154 @@
...
+      preparedTodo.id
+    )));
+
+    return maxId + 1;
+  };
+
+  const addNewTodo = (): void => {
+    const user = usersFromServer.find(userFromServer => (
+      userFromServer.id === userSelectedId
+    )) || null;
```
roman-mirzoian, line 51: You have a `getUserById` helper - you can reuse it here

src/App.tsx
```diff
@@ -1,61 +1,154 @@
...
+            <input
+              type="text"
+              id="titleInput"
+              data-cy="titleInput"
+              placeholder="Enter a title"
+              value={title}
+              onChange={event => {
+                setTitle(event.target.value);
+                setHasTitleError(false);
+              }}
```
roman-mirzoian, line 102: For better readability, this callback should be separated as a separate handler

src/App.tsx
```diff
@@ -1,61 +1,154 @@
...
-          </select>
+          <label htmlFor="userSelector">
+            <span>User: </span>
+            <select
+              data-cy="userSelect"
+              value={userSelectedId}
+              onChange={event => {
+                setUserSelectedId(+event.target.value);
+                setHasUserError(false);
+              }}
```
roman-mirzoian, line 122: The same issue regarding separated callback

src/App.tsx
```diff
@@ -1,61 +1,154 @@
...
+            >
+              <option value="0" disabled>
+                Choose a user
+              </option>
+
+              {usersFromServer.map((userFromServer) => (
```
roman-mirzoian, line 128: Use destructuring on the `userFromServer` object.

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,24 @@
...
+import cn from 'classnames';
+
+import { PreparedTodo } from '../../types';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: PreparedTodo
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
```
roman-mirzoian, line 11: Use destructuring on the `todo` object.

src/types.ts
```diff
@@ -0,0 +1,21 @@
...
+  userId: number,
+}
+
+export interface PreparedTodo {
+  id: number,
+  title: string,
+  completed: true | false,
+  userId: number,
+  user: User | null;
+}
```
roman-mirzoian, line 21: These interfaces are very similar to each other, so you should use `extends` to avoid copying code

roman-mirzoian (changes requested):

src/App.tsx
VeDono: I need whole `usersFromServer.find((user) => user.id === userId) || null;` cause user can not be found (i think, but i reimplement that little-bit)

olegnovikov10 (approved):
good job :)

src/App.tsx
roman-mirzoian: @VeDono By default, the `find` method returns `undefined` if it does not find anything, so it is not necessary to specify `null`, you can simply specify an additional type of `undefined` for the `user` interface
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

src/App.tsx
```diff
@@ -1,61 +1,153 @@
...
+
 import './App.scss';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { TodoList } from './components/TodoList';
+import { PreparedTodo, Todo } from './types';
+
+function getUserById(userId: number) {
+  return usersFromServer.find((user) => user.id === userId) || null;
```
roman-mirzoian, line 11: @VeDono By default, the `find` method returns `undefined` if it does not find anything, so it is not necessary to specify `null`, you can simply specify an additional type of `undefined` for the `user` interface
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

PR #2222, opened 2023-12-06
Solution

Barosz30:
https://barosz30.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -61,11 +61,11 @@ export const App = () => {
       setTitleError(true);
     }
 
-    if (selectedUser === 0) {
+    if (!selectedUser) {
       setUserError(true);
     }
 
-    if (title && selectedUser !== 0) {
+    if (title && selectedUser) {
       pushUser({
         id: Math.max(...todos.map(({ id }) => id)) + 1,
         title,


src/App.tsx
```diff
@@ -1,61 +1,135 @@
...
+        id: Math.max(...todos.map(({ id }) => id)) + 1,
+        title,
+        completed: false,
+        userId: 0,
+        user: getUser(option),
+      });
+
+      reset();
+    }
+  };
```
mvjl000, line 82: Can you modify this function a bit, so this part of the code is not inside of `if` statement?

What I mean is, you are already checking title and option at the beginning of the , I would like see how would you approach this ;)
```
      pushUser({
        id: Math.max(...todos.map(({ id }) => id)) + 1,
        title,
        completed: false,
        userId: 0,
        user: getUser(option),
      });

      reset();
```

src/App.tsx
```diff
@@ -1,61 +1,135 @@
...
+          <label>
+            User:&nbsp;
+            <select
+              data-cy="userSelect"
+              value={option}
+              onChange={handleUserChange}
+            >
+              <option value="0" disabled>Choose a user</option>
+              {usersFromServer.map((user) => (
+                <option value={user.id}>
```
mvjl000, line 117: Since we have access to a unique value (`user.id`) it would be good to set it as key

src/App.tsx
```diff
@@ -1,61 +1,135 @@
...
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [currentTodos, setCurrentTodos] = useState<Todo[]>(todos);
+
+  const [title, setTitle] = useState('');
+  const [titleError, setTitleError] = useState(false);
+
+  const [option, setOption] = useState(0);
```
mvjl000, line 29: Two things here:

- `option` doesn't say much, something like `selectedUser` would be way better.

- For the default value (no user selected) I would use null
```suggestion
  const [selectedUser, setSelectedUser] = useState<number | null>(null);
```

src/App.tsx
```diff
@@ -1,61 +1,135 @@
...
+  const [userError, setUserError] = useState(false);
+
+  const pushUser = (todo: Todo) => {
+    const newTodo = {
+      ...todo,
+    };
```
mvjl000, line 35: You don't need to create a new object. `todo` has just been created in handleSubmit

mvjl000 (changes requested):

src/App.tsx
Barosz30: It kinda prevents code from starting on 0 field, that's 'chose a user', but it got implemented ;)

Barosz30: Cant do that. Tried, but it makes it impossible to use proper starting state. Also Value cannot use null type

src/App.tsx
```diff
@@ -1,61 +1,132 @@
...
+    setUserError(false);
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+    if (!title) {
+      setTitleError(true);
+    }
+
+    if (selectedUser === 0) {
```
kate-prakofyeva, line 64: selectedUser === 0 the same as !selectedUser 
```
Falsy values are values that resolve to the Boolean false.
In JavaScript, we have 6 falsy values:
false
0 (zero)
‘’ or “” (empty string)
null
undefined
NaN
```

kate-prakofyeva (approved):

mvjl000 (approved):
All good!

PR #2223, opened 2023-12-06
add task solution

yuliiababych:
[DEMO LINK](https://yuliiababych.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,1 +1,23 @@
-export const TodoInfo = () => {};
+import React from 'react';
+import { UserInfo } from '../UserInfo';
+import { Todo } from '../types/Todo';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const todoClass = todo.completed ? 'TodoInfo--completed' : '';
+
+  return (
+    <article
+      data-id={todo.id}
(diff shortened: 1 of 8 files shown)


nazarmatsevych (changes requested):
The code looks good but I can add todo with an empty title, use trim() method to fix it
image

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,23 @@
...
+
+  return (
+    <article
+      data-id={todo.id}
+      className={`TodoInfo ${todoClass}`}
+    >
+      <h2 className="TodoInfo__title">
+        {todo.title}
+      </h2>
+      <UserInfo user={todo.user} />
```
etojeDenys, line 20: user can be null we need to handle this case here

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,22 @@
...
+};
+
+type Props = {
+  user: User | null;
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  if (!user) {
+    return null;
+  }
```
etojeDenys, line 15: let's remove this logic from component, UserInfo targets to render the user information, let's not complicate it with the existing logic

etojeDenys (changes requested):
probably you forgot to re-deploy
image

etojeDenys (approved):
good job

PR #2231, opened 2023-12-08
Develop

jacekkosiba:
[DEMO LINK](https://jacekkosiba.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/UserInfo/UserInfo.tsx
+++ b/src/components/UserInfo/UserInfo.tsx
@@ -1,1 +1,9 @@
-export const UserInfo = () => {};
+import { UserInfoProps } from '../../types';
+
+export const UserInfo = ({ user }: UserInfoProps) => {
+  return (
+    <a className="UserInfo" href="mailto:Sincere@april.biz">
+      {user?.name}
+    </a>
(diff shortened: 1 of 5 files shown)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,9 @@
-export const UserInfo = () => {};
+import { UserInfoProps } from '../../types';
+
+export const UserInfo = ({ user }: UserInfoProps) => {
+  return (
+    <a className="UserInfo" href={`mailto:${user?.email}`}>
+      {user?.name}
```
zaura333, line 6: Checking for user could be moved to TodoInfo before rendering the component (as suggested in checklist 5)

zaura333 (changes requested):
Overall, very good job!

I've noticed one thing: after adding a correct todo, the form immediately shows error messages on empty inputs, you might want to check that out

On the side note, I think the form could be moved to a new component, but that's probably a matter of preference.

zaura333 (approved):

mbruhler (approved):

PR #2232, opened 2023-12-09
Added_todo-form_OkM

Oksana-Maruniak:
[DEMO LINK](https://Oksana-Maruniak.github.io/react_add-todo-form/)


SanyaBratashchuk (changes requested):
1. Improve validation (I can add todo with title that contains only spaces)
image

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,123 @@
...
+import { getUserById } from '../../services/user_service';
+import { getNextAvailableId } from '../../services/todo_service';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  onSubmit: (todo: Todo) => void;
+};
+
+export const TodoForm: React.FC<Props> = ({ onSubmit }) => {
+  // #region state
```
VitaliyBodnar, line 12: please remove all comments

src/services/user_service.ts
```diff
@@ -0,0 +1,6 @@
+import usersFromServer from '../api/users';
+
+export function getUserById(userId: number) {
+  return usersFromServer.find(user => user.id === userId)
+  || null;
```
VitaliyBodnar, line 5: `|| null` is redundant cause method find always return an array 
```suggestion
  return usersFromServer.find(user => user.id === userId);
```

VitaliyBodnar (changes requested):

PR #2233, opened 2023-12-09
addreact_add-todo-form

Dizelkos:
[DEMO LINK](https://Dizelkos.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,111 @@
 import './App.scss';
+import { useState } from 'react';
+import { Todo } from './types/Todo';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { TodoList } from './components/TodoList';
 
 export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>(todosFromServer);
+
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,121 @@
...
+    setUserId(0);
+  };
+
+  const handleOnSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    setTitleError(!title);
+    setUserError(!userId);
+
+    if (!title || !userId) {
```
etojeDenys, line 40:
```suggestion
    if (!title.trim() || !userId.trim()) {
```
to avoid todo only with spaces
image

src/App.tsx
```diff
@@ -1,61 +1,121 @@
...
+            <option value={0} disabled>Choose a user</option>
+
+            {usersFromServer.map(user => {
+              const { id, name } = user;
+
+              return (
+                <option
+                  value={id}
+                  key={id}
+                >
```
etojeDenys, line 99:
```suggestion
            {usersFromServer.map(({id, name})=> (
                <option
                  value={id}
                  key={id}
                >
```

etojeDenys (changes requested):
nits:

anastasiiavorobiova (changes requested):
Great work! Consider showing an error if the user tries to submit spaces only 
Screenshot 2023-12-10 at 18 24 48

volodymyr-soltys97 (approved):
Great work 🔥

PR #2236, opened 2023-12-10
Solution

Patrykmclaren14:
[solution](https://Patrykmclaren14.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,107 @@
+import { useState, ChangeEvent, FormEvent } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { TodoList } from './components/TodoList';
 
 export const App = () => {
+  const [todoTitle, setTodoTitle] = useState<string>('');
+  const [todoUser, setTodoUser] = useState<string>('0');
+  const [todos, setTodos] = useState(todosFromServer);
+  const [userError, setUserError] = useState<boolean>(false);
+  const [titleError, setTitleError] = useState<boolean>(false);
+
+  const title = (event: ChangeEvent<HTMLInputElement>) => {
+    setTodoTitle(event.target.value);
+    setTitleError(false);
+  };
+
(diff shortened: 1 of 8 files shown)


MarcinLigmann (approved):
looks goos

src/App.tsx
```diff
@@ -1,61 +1,107 @@
...
+
+  const title = (event: ChangeEvent<HTMLInputElement>) => {
+    setTodoTitle(event.target.value);
+    setTitleError(false);
+  };
+
+  const person = (event: ChangeEvent<HTMLSelectElement>) => {
+    setTodoUser(event.target.value);
+    setUserError(false);
+  };
```
mvjl000, line 24: Please change the names of those handler functions to something like: `handleTitleChange`

src/App.tsx
```diff
@@ -1,61 +1,107 @@
...
+
+  const createTodo = (e: FormEvent) => {
```
mvjl000, line 26: Please keep using one convention. You were using `event` in two previous handler functions so stick to that

src/App.tsx
```diff
@@ -1,61 +1,107 @@
...
+        (acc, curr) => (curr.id > acc.id ? curr : acc), todos[0],
+      ).id + 1;
+
+    if (todoUser === '0') {
+      setUserError(true);
+    }
+
+    if (todoTitle.length === 0) {
+      setTitleError(true);
+    }
```
mvjl000, line 39: First validate, then calculate things.

Move these if statements to the top of the function. You should also stop the function if one of them is falsy to prevent looking for user when form is invalid.

mvjl000 (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,107 @@
...
+
+    event.preventDefault();
+    const newId
+      = todos.reduce(
+        (acc, curr) => (curr.id > acc.id ? curr : acc), todos[0],
+      ).id + 1;
```
JoyCoffeeAddict, line 39: i would sort and then find the biggest id, but i guess that also makes sense (but your approach may be less readable for other people at the project)

JoyCoffeeAddict (approved):
great job, but think about my suggestion

PR #2244, opened 2023-12-11
Solution 

adrian200002:
[DEMO LINK](https://adrian200002.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,98 @@
 import './App.scss';
-
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { FormEventHandler, useState } from 'react';
+import usersFromServer from './api/users';
+import { TodoList } from './components/TodoList';
+import todosFromServer from './api/todos';
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [selectedUser, setSelectedUser] = useState('');
+  const [titleError, setTitleError] = useState('');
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,98 @@
...
+    const foundUser = usersFromServer.find(user => user.name === selectedUser);
+    const newId = Math.max(...todosFromServer.map(todo => todo.id));
+
+    if (!title) {
+      setTitleError('Please enter a title');
+    }
+
+    if (!selectedUser) {
+      setUserError('Please choose a user');
+    }
```
mvjl000, line 25: It would make sense to first validate and then look for data necessary to create new item. If form is invalid, prevent it from happening

src/App.tsx
```diff
@@ -1,61 +1,98 @@
...
+          <label htmlFor="title">Title: </label>
+          <input
+            type="text"
+            id="title"
+            placeholder="Enter a title"
+            data-cy="titleInput"
+            value={title}
+            onChange={handleTitle}
+          />
+          <span className="error">{titleError}</span>
```
mvjl000, line 67: Render this span conditionally so it doesn't get rendered when there is no error

mvjl000: Same for user error

mvjl000 (changes requested):

Sibunas (approved):

mbruhler (approved):

PR #2251, opened 2023-12-14
Solution

Sibunas:
[DEMO LINK](https://sibunas.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,92 @@
+import { SetStateAction, useState } from 'react';
 import './App.scss';
-
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
 export const App = () => {
+  const [todos, setTodos] = useState(todosFromServer);
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState('');
+  const [titleError, setTitleError] = useState('');
+  const [userError, setUserError] = useState('');
+
+  const handleTitleChange = (event:
+  { target: { value: SetStateAction<string>; }; }) => {
+    setTitle(event.target.value);
+    if (titleError) {
+      setTitleError('');
+    }
+  };
+
+  const handleUserChange = (event:
+  { target: { value: SetStateAction<string>; }; }) => {
+    setUserId(event.target.value);
+    if (userError) {
+      setUserError('');
+    }
+  };
+
+  const addTodo = () => {
+    const newTodo = {
+      id: todos.length + 1,
+      title,
+      userId: Number(userId),
+      completed: false,
+    };
+
+    setTodos([...todos, newTodo]);
+    setTitle('');
+    setUserId('');
+  };
+
+  const handleSubmit = (event: { preventDefault: () => void; }) => {
+    event.preventDefault();
+
+    if (!title.trim()) {
+      setTitleError('Please enter a title');
+
+      return;
+    }
+
+    if (!userId) {
+      setUserError('Please choose a user');
+
+      return;
+    }
+
+    addTodo();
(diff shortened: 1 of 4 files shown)


WiktoriaBielec (changes requested):
Test didn't pass and demo link doesn't work.

WiktoriaBielec (approved):
Works great:)

src/App.tsx
```diff
@@ -1,61 +1,147 @@
...
+  const [userId, setUserId] = useState('');
+  const [titleError, setTitleError] = useState('');
+  const [userError, setUserError] = useState('');
+
+  useEffect(() => {
+    const todosWithUsers = todosFromServer.map(todo => {
+      const user = usersFromServer.find(u => u.id === todo.userId);
+
+      if (!user) {
+        throw new Error('User not found');
```
JoyCoffeeAddict, line 33: don't throw without any reason, just add `console.error` and return

src/App.tsx
```diff
@@ -1,61 +1,147 @@
...
+      }
+
+      return {
+        ...todo,
+        user,
+      };
+    });
+
+    setTodos(todosWithUsers);
+  }, []);
```
JoyCoffeeAddict, line 43: You don't need `useEffect` for setting thtat. Just use a variable. That is also a sign, that you should read what the `useEffect` is in the documentation. In a nutshell, it's for doing side effects (like fetch, displaying errors, synchronizing components based on state)

src/App.tsx
```diff
@@ -1,61 +1,147 @@
...
+    }
+
+    const newTodo = {
+      id: maxId + 1,
+      title,
+      user,
+      completed: false,
+    };
+
+    setTodos([...todos, newTodo]);
```
JoyCoffeeAddict, line 81: use prevState syntax

JoyCoffeeAddict (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,129 @@
...
+      };
+
+      return [...prevTodos, newTodo];
+    });
+
+    setTitle('');
+    setUserId('');
+  };
+
+  const handleSubmit = (event: { preventDefault: () => void; }) => {
```
mpiskadlo, line 68: Use formHanlder type, your IDE should suggest you how should it look like

mpiskadlo: Hint: you will need to import it from React

src/App.tsx
```diff
@@ -1,61 +1,129 @@
...
+  const handleTitleChange = (event:
+  { target: { value: SetStateAction<string>; }; }) => {
+    setTitle(event.target.value);
+    if (titleError) {
+      setTitleError('');
+    }
+  };
+
+  const handleUserChange = (event:
+  { target: { value: SetStateAction<string>; }; }) => {
```
mpiskadlo, line 33: Those types for events look really messy here, try do do it in a similar way as I suggested in `handleSubmit`

src/App.tsx
```diff
@@ -1,61 +1,129 @@
...
 
 export const App = () => {
+  const todosWithUsers = todosFromServer.map(todo => {
+    const user = usersFromServer.find(u => u.id === todo.userId) || null;
+
+    return {
+      ...todo,
+      user,
+    };
+  });
```
mpiskadlo, line 16: This could be declared outside of the component

src/App.tsx
```diff
@@ -1,61 +1,129 @@
...
+    }
+
+    if (!userId) {
+      setUserError('Please choose a user');
+      error = true;
+    }
+
+    if (!error) {
+      addTodo();
+    }
```
mpiskadlo, line 84: Does the same thing but I think that it looks better ;) 
```suggestion
    if (error) return;
    
    addTodo();
```

mpiskadlo (changes requested):

PR #2253, opened 2023-12-16
Develop

stunnerboyone:
[DEMO LINK](https://stunnerboyone.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/UserInfo/UserInfo.tsx
+++ b/src/components/UserInfo/UserInfo.tsx
@@ -1,1 +1,30 @@
-export const UserInfo = () => {};
+import React from 'react';
+
+import usersFromServer from '../../api/users';
+
+import { User } from '../../react-app-env';
+
+interface Props {
+  userId: number;
+}
+
+export const UserInfo: React.FC<Props> = ({
+  userId,
+}) => {
+  let user: User | undefined = usersFromServer.find(u => u.id === +userId);
+
+  if (!user) {
+    user = {
+      id: Date.now(),
+      name: 'No name',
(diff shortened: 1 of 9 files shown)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,30 @@
...
+  let user: User | undefined = usersFromServer.find(u => u.id === +userId);
+
+  if (!user) {
+    user = {
+      id: Date.now(),
+      name: 'No name',
+      username: 'No name',
+      email: '#',
+    };
+  }
```
VitaliyBondarenko1982, line 23:
```suggestion
  if (!user) {
   return null
```
it can fix tests

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,30 @@
...
+import { User } from '../../react-app-env';
+
+interface Props {
+  userId: number;
+}
+
+export const UserInfo: React.FC<Props> = ({
+  userId,
+}) => {
+  let user: User | undefined = usersFromServer.find(u => u.id === +userId);
```
VitaliyBondarenko1982, line 14:
```suggestion
  const user: User | undefined = usersFromServer.find(u => u.id === userId);
```
userId already number. looks like no need to make number from number.

VitaliyBondarenko1982 (changes requested):
Good job.
Chack comments and try to fix tests.

SanyaBratashchuk (approved):

PR #2256, opened 2023-12-19
react_add-todo-form__solution

olha-rypich:
[DEMO LINK](https://olha-rypich.github.io/react_add-todo-form/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
Use trim() before add title so that it is not possible to add an empty title
image

lerastarynets (approved):
lgtm 🚀

PR #2260, opened 2023-12-21
add task solution

Nataliia95254:
[DEMO LINK](https://Nataliia95254.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,1 +1,32 @@
-export const TodoInfo = () => {};
+import classnames from 'classnames';
+import { User } from '../../type/user';
+import { Todo } from '../../type/todo';
+
+type Props = {
+  todo: Todo;
+  findUser: (id: number) => User | undefined;
+};
+
+export const TodoInfo: React.FC<Props> = (
+  {
+    todo: {
+      id, completed, title, userId,
+    },
+    findUser,
+  },
+) => {
+  return (
+    <article
+      data-id={id}
+      className={classnames('TodoInfo', {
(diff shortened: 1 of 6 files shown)


nazarmatsevych (changes requested):
Good work but I can add todo with an empty title, use trim() method to fix it
image

If you have any questions - feel free to ask them in fe_chat

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,32 @@
-export const TodoInfo = () => {};
+import classnames from 'classnames';
+import { User } from '../../type/user';
+import { Todo } from '../../type/todo';
+
+type Props = {
+  todo: Todo;
+  findUser: (id: number) => User | undefined;
```
etojeDenys, line 7: why do you need to pass that function here as the prop if you just can add it in the component as the simple function

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,32 @@
...
+    <article
+      data-id={id}
+      className={classnames('TodoInfo', {
+        'TodoInfo--completed': completed,
+      })}
+      key={id}
+    >
+      <h2 className="TodoInfo__title">{title}</h2>
+      <a className="UserInfo" href={`mailto:${findUser(userId)?.email}`}>
+        {findUser(userId)?.name}
```
etojeDenys, line 28: better to save foundUser in some var and reuse that var here

etojeDenys (changes requested):

maxim2310 (approved):
Good work!!!

PR #2267, opened 2023-12-28
Develop

IShamkii:
[DEMO LINK](https://IShamkii.github.io/react_add-todo-form/)

### DIFF
--- a/src/types/todo.ts
+++ b/src/types/todo.ts
@@ -3,7 +3,7 @@ export interface Todo {
   title: string;
   completed: boolean;
   userId: number;
-  user: {
+  user?: {
     id: number;
     name: string;
     username: string;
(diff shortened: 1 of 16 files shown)


src/types/todo.ts
```diff
@@ -0,0 +1,12 @@
...
+  title: string;
+  completed: boolean;
+  userId: number;
+  user?: {
+    id: number;
+    name: string;
+    username: string;
+    email: string;
+  } | null;
+}
```
etojeDenys, line 12:
```suggestion
  user?: User | null;
}
```

etojeDenys (changes requested):
image

src/types/todo.ts
etojeDenys: let's fix that
image

etojeDenys (changes requested):

TarasHoliuk (approved):

PR #2269, opened 2023-12-29
happy new year

kalkvova:
[DEMO LINK](https://kalkvova.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,106 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
 
 export const App = () => {
+  const [todos, SetTodos] = useState(todosFromServer);
+
+  const [newTitle, setNewTitle] = useState('');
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,61 +1,106 @@
...
-            <option value="0" disabled>Choose a user</option>
+          <label htmlFor="user">User: </label>
+          <select
+            data-cy="userSelect"
+            id="user"
+            value={selectedUserId}
+            onChange={handleSelectUserId}
+          >
+            <option value="">Choose a user</option>
+            {usersFromServer.map((user) => (
```
volodymyr-soltys97, line 88: Use a destruction for `user`

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,12 @@
-export const UserInfo = () => {};
+import { User } from '../../types/User';
+
+interface Props {
+  user: User,
+}
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  return (
+    <a className="UserInfo" href={`mailto:${user.email}`}>
+      {user.name}
```
volodymyr-soltys97, line 9: The same as above

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
Need to add the message error text when I try add the empty title, or only the spaces
image

PR #2274, opened 2024-01-06
complete task

boytrend0108:
[DEMO LINK](https://boytrend0108.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,1 +1,34 @@
-export const TodoInfo = () => {};
+import cn from 'classnames';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    title,
+    id,
+    user,
+    completed,
+  } = todo;
+
+  return (
+    <article
+      data-id={id}
+      className={cn('TodoInfo', {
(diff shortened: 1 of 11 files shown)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,13 @@
...
+type Props = {
+  user: User | undefined,
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  return (
+    <a className="UserInfo" href={`mailto:${user?.email}`}>
+      {user?.name}
+    </a>
+  );
```
etojeDenys, line 12:
```suggestion
type Props = {
  user: User,
};

export const UserInfo: React.FC<Props> = ({ user }) => {
  return (
    <a className="UserInfo" href={`mailto:${user.email}`}>
      {user.name}
    </a>
  );
```
let's not render this component if a user doesn't exist, we don't want to complicate the logic of component, his goal is just to render the user information

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,34 @@
...
+      })}
+    >
+      <h2
+        className="TodoInfo__title"
+        style={{ color: completed ? 'green' : 'red' }}
+      >
+        {title}
+      </h2>
+
+      <UserInfo user={user} />
```
etojeDenys, line 31: just add the check here
```suggestion
      {user && <UserInfo user={user} />}
```

etojeDenys (changes requested):
image

etojeDenys (approved):
lgtm

PR #2275, opened 2024-01-07
react_add-todo-form

serjgrm:
[DEMO LINK](https://serjgrm.github.io/react_add-todo-form/)

### DIFF
--- a/src/types/TodosFromServer.ts
+++ b/src/types/TodosFromServer.ts
@@ -1,0 +1,6 @@
+export interface TodosFromServer {
+  id: number,
+  title: string,
+  completed: boolean,
+  userId: number,
+}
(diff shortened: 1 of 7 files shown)


SanyaBratashchuk (approved):

src/types/TodosFromServer.ts
```diff
@@ -0,0 +1,6 @@
+export interface TodosFromServer {
```
SanyaBratashchuk, line 1:
```suggestion
export interface Todo {
```

src/types/UserFromServer.ts
```diff
@@ -0,0 +1,6 @@
+export interface UserFromServer {
```
SanyaBratashchuk, line 1:
```suggestion
export interface User {
```

SanyaBratashchuk (approved):

PR #2283, opened 2024-01-11
Develop

ivan-baranovskyi:
[DEMO LINK](https://ivan-baranovskyi.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,7 +4,7 @@ import React, { ChangeEvent, useState } from 'react';
 import usersFromServer from './api/users';
 import todosFromServer from './api/todos';
 import { TodoList } from './components/TodoList';
-import { Todos, getMaxId } from './components/TodoInfo';
+import { Todos } from './components/TodoInfo';
 
 function getUserById(userId: number) {
   return usersFromServer.find((user) => user.id === userId);
@@ -14,6 +14,12 @@ export const todos = todosFromServer.map((todo) => ({
   ...todo,
   user: getUserById(todo.userId),
 }));
+
+export function getMaxId(todo: Todos[]) {
+  const maxId = Math.max(0, ...todo.map((item) => item.id));
+
+  return maxId + 1;
+}
 
 const initialUserState = {
   id: 0,
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,134 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,61 +1,140 @@
...
+            <select
+              id="chooseUser"
+              data-cy="userSelect"
+              value={todo.userId}
+              onChange={handleSelected}
+            >
+              <option value="0" disabled>
+                Choose a user
+              </option>
+              {usersFromServer.map((user) => (
```
volodymyr-soltys97, line 119: Use a destruction for `user`

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
Need to use a trim for the title before added
There will be no possibility to add an empty title as spaces
image

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import React, { ChangeEvent, useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Todos } from './components/TodoInfo';
+
+function getUserById(userId: number) {
+  return usersFromServer.find((user) => user.id === userId);
```
denys-danyliuk, line 10:
```suggestion
  return usersFromServer.find((user) => user.id === userId) || null;
```
there can be no user

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
+  completed: false,
+};
 
 export const App = () => {
+  const [todo, setTodo] = useState(initialUserState);
+
+  const [hasTitleError, setHasTitleError] = useState(false);
+  const [hasUserError, setHasUserError] = useState(false);
+
+  const [newTodo, setNewTodo] = useState<Todos[]>(todos);
```
denys-danyliuk, line 37: your variable called 'newTodo', but at the beginning it has initial value as todos array, you should pick a better name or rethink this logic

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
+
+  const newUser = getUserById(todo.userId);
```
denys-danyliuk, line 39: it's not actually new :)

denys-danyliuk (approved):
Keep it up! Pay attention to the comments and overall naming of variables

PR #2295, opened 2024-01-17
add solution

Xilston:
[DEMO LINK](https://xilston.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/AddTodoForm/AddTodoForm.tsx
+++ b/src/components/AddTodoForm/AddTodoForm.tsx
@@ -9,13 +9,13 @@ interface Props {
 export const AddTodoForm = (props: Props) => {
   const { users, addTodo } = props;
 
-  const [selectedUser, setSelectedUser] = useState(0);
+  const [selectedUser, setSelectedUser] = useState<number | null>(null);
   const [newTitle, setNewTitle] = useState('');
 
   const [showErrors, setShowErrors] = useState(false);
 
   function resetForm() {
-    setSelectedUser(0);
+    setSelectedUser(null);
     setNewTitle('');
   }
 
@@ -64,7 +64,7 @@ export const AddTodoForm = (props: Props) => {
             <select
               name="userSelect"
               data-cy="userSelect"
-              value={selectedUser}
+              value={selectedUser !== null ? selectedUser : 0}
               onChange={(event) => setSelectedUser(+event.target.value)}
             >
               <option value="0" key={0} disabled>Choose a user</option>
--- a/src/components/UserInfo/UserInfo.tsx
+++ b/src/components/UserInfo/UserInfo.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { User } from '../../types/User';
 
 interface Props {
-  user: User | undefined;
+  user: User | null;
 }
 
 export const UserInfo: React.FC<Props> = ({ user }) => {
--- a/src/types/Todo.ts
+++ b/src/types/Todo.ts
@@ -4,9 +4,3 @@ export interface Todo {
   completed: boolean;
   userId: number;
 }
-
-export function getLatestTodoId(arr: Todo[]): number {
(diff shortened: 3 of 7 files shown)


src/components/AddTodoForm/AddTodoForm.tsx
```diff
@@ -0,0 +1,93 @@
...
+  function resetForm() {
+    setSelectedUser(0);
+    setNewTitle('');
+  }
+
+  function handleSubmit(event: React.FormEvent) {
+    event.preventDefault();
+    if (!showErrors) {
+      setShowErrors(true);
+    }
```
YegorKorenkov, line 26: why do you show errors before checking them? maybe it would be better to check first and then show?

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,14 @@
-export const UserInfo = () => {};
+import React from 'react';
+import { User } from '../../types/User';
+
+interface Props {
+  user: User | undefined;
```
YegorKorenkov, line 5: better to use `null` instead of undefined when your work with objects. `Null` describes that this is an empty object

src/types/Todo.ts
```diff
@@ -0,0 +1,12 @@
...
+  title: string;
+  completed: boolean;
+  userId: number;
+}
+
+export function getLatestTodoId(arr: Todo[]): number {
+  const ids: number[] = arr.map(todo => todo.id);
+
+  return Math.max(...ids);
+}
```
YegorKorenkov, line 12: How does this relate to the types that are located in this file. Please move this function to another file, utils for example

src/types/User.ts
```diff
@@ -0,0 +1,10 @@
+export interface User {
+  id: number;
+  name: string;
+  username: string;
+  email: string;
+}
+
+export function getUserById(array: User[], id: number): User | undefined {
+  return array.find(user => user.id === id);
```
YegorKorenkov, line 9: better to return null if object not founded, check my comment above

src/components/AddTodoForm/AddTodoForm.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+interface Props {
+  users: User[];
+  addTodo: (title: string, userId: number) => void;
+}
+
+export const AddTodoForm = (props: Props) => {
+  const { users, addTodo } = props;
+
+  const [selectedUser, setSelectedUser] = useState(0);
```
YegorKorenkov, line 12: Use null instead of 0 for initial state

src/components/AddTodoForm/AddTodoForm.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const [newTitle, setNewTitle] = useState('');
+
+  const [showErrors, setShowErrors] = useState(false);
+
+  function resetForm() {
+    setSelectedUser(0);
```
YegorKorenkov, line 18: Same, set null

YegorKorenkov (changes requested):
Goooood job! Please check my comments and fix minor mistakes

src/components/AddTodoForm/AddTodoForm.tsx
```diff
@@ -0,0 +1,93 @@
...
+          )}
+        </div>
+
+        <div className="field">
+          <label htmlFor="userSelect">
+            {'User: '}
+            <select
+              name="userSelect"
+              data-cy="userSelect"
+              value={selectedUser !== null ? selectedUser : 0}
```
DanilWeda, line 67:
```suggestion
              value={selectedUser !== null ? selectedUser : defaultUserValue}
```
What does mean zero? Need constant which describes what means this value ( magic numbers )

DanilWeda (approved):
Good job! 
Approved. 
Many thanks.

PR #2324, opened 2024-01-23
add task solutipn

VazilX:
[DEMO LINK](https://VazilX.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,150 @@
 import './App.scss';
+import { useState } from 'react';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { User } from './types/User';
+import { TodoFromServer } from './types/TodoFromServer';
+
+const getTodoId = () => {
+  const ids = todosFromServer.map(todo => todo.id);
+
+  return Math.max(...ids) + 1;
+};
+
+const getUser = (users: User[], userId: number) => {
+  return users.find(user => user.id === userId) as User;
+};
+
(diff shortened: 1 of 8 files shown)


src/types/Todo.ts
```diff
@@ -0,0 +1,12 @@
...
+  userId: number;
+  id: number;
+  title: string;
+  completed: boolean;
+  user: {
+    id: number;
+    name: string;
+    username: string;
+    email: string;
+  };
```
etojeDenys, line 11:
```suggestion
  user: User | null;
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,26 @@
...
+        'TodoInfo--completed': todo.completed,
+      })}
+    >
+      <h2 className="TodoInfo__title">
+        {todo.title}
+      </h2>
+
+      <UserInfo
+        user={todo.user}
+      />
```
etojeDenys, line 23: user can be null, let's handle it here
```suggestion
      {todo.user && <UserInfo
        user={todo.user}
      />}
```

src/App.tsx
```diff
@@ -1,61 +1,160 @@
...
+import { TodoFromServer } from './types/TodoFromServer';
+
+const getTodoId = () => {
+  const ids = todosFromServer.map(todo => todo.id);
+
+  return Math.max(...ids) + 1;
+};
+
+const getUser = (users: User[], userId: number) => {
+  return users.find(user => user.id === userId) as User;
```
etojeDenys, line 17:
```suggestion
  return users.find(user => user.id === userId) || null;
```

etojeDenys (changes requested):

anastasiiavorobiova (approved):
Great work!

PR #2327, opened 2024-01-25
add task react_add-todo-form

P-Nazar:
[DEMO LINK](https://P-Nazar.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoForm/TodoForm.tsx
+++ b/src/components/TodoForm/TodoForm.tsx
@@ -1,0 +1,115 @@
+import React, { useState } from 'react';
+
+import { Users } from '../../types/User';
+import { Todo, Todos } from '../../types/Todo';
+import { getUserById } from '../../services/getUserById';
+import { generateNewId } from '../../services/generateNewId';
+
+interface Props {
+  users: Users
+  onSubmit: (todo: Todo) => void
+  todosArr: Todos
+}
+
+export const TodoForm: React.FC<Props> = ({
+  users,
+  onSubmit,
+  todosArr,
+}) => {
+  const [title, setTitle] = useState('');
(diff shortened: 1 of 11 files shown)


etojeDenys (changes requested):
image

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,115 @@
...
+      <div className="field">
+        <label htmlFor="userSelect">User: </label>
+        <select
+          data-cy="userSelect"
+          id="userSelect"
+          value={userId}
+          onChange={handleUserChange}
+        >
+          <option value="0" disabled>Choose a user</option>
+          {users.map(user => (
```
loralevitska, line 94:
```suggestion
          {users.map(({ id, name })=> (
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,16 @@
...
+import { Todos } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo';
+
+interface Props {
+  todos: Todos
+}
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="TodoList">
```
loralevitska, line 10: Use semantic tags for list `ul` `li`. Don't forget to remove default `li` dots and `ul` paddings

loralevitska (changes requested):
- implement this for empty spaces too ([string.trim()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim) can be helpful)
image

etojeDenys (approved):
awesome

PR #2331, opened 2024-01-27
add_todo_form

JokerMTM:
[DEMO LINK](https://JokerMTM.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,105 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { getNewId } from './servises/NewId';
 
 export const App = () => {
+  const [todos, setTodos] = useState(todosFromServer);
+
+  const [title, setTitle] = useState('');
+  const [hasTitleError, setHasTitleError] = useState(false);
+
+  const [userId, setUserId] = useState(0);
+  const [hasUserIdError, setHasUserIdError] = useState(false);
+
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,116 @@
...
+          </label>
+          <br />
+          <select
+            id="userSelect"
+            data-cy="userSelect"
+            value={userId}
+            onChange={handleUserIdChange}
+          >
             <option value="0" disabled>Choose a user</option>
+            {usersFromServer.map(user => (
```
volodymyr-soltys97, line 90: Use the destruction for `user`

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
Need to fix position inputs
Working version
image

Demo
image

The title cannot be an empty, fix it
image

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,29 @@
...
+import React from 'react';
+
+type UserProps = {
+  user: {
+    id: number;
+    name: string;
+    username: string;
+    email: string;
+  } | null;
+};
```
etojeDenys, line 10:
```suggestion
type UserProps = {
  user: {
    id: number;
    name: string;
    username: string;
    email: string;
  };
};
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,40 @@
...
+        { 'TodoInfo--completed': completed },
+      )}
+    >
+      <h2 className="TodoInfo__title">
+        {title}
+      </h2>
+
+      <UserInfo
+        user={user}
+      />
```
etojeDenys, line 36: better to render this component only when user exists

etojeDenys (changes requested):
image

TarasHoliuk (approved):

PR #2332, opened 2024-01-27
add task solutions

MrWhiteres:
[DEMO LINK](https://MrWhiteres.github.io/react_add-todo-form/)


src/api/index.ts
```diff
@@ -0,0 +1,14 @@
+import { PreparedTodo, User } from '../types';
+import todosFromServer from './todos';
+import Users from './users';
+
+export function findUser(userId:number):User {
+  return [...Users].find(user => user.id === userId) as User;
```
etojeDenys, line 6:
```suggestion
  return [...Users].find(user => user.id === userId) || null;
```

src/types/PreparedTodo.ts
```diff
@@ -0,0 +1,9 @@
+import { User } from './User';
+
+export interface PreparedTodo {
+  id: number;
+  title: string;
+  completed: boolean;
+  userId: number;
+  user: User;
```
etojeDenys, line 8:
```suggestion
  user: User | null;
```

etojeDenys (changes requested):
image

PR #2339, opened 2024-02-03
solution

Dmytrosp:
[DEMO LINK](https://Dmytrosp.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,109 @@
+import { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/Todo';
 
 export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>(todosFromServer);
+
+  const [userIdError, setUserIdError] = useState(false);
+  const [inputTitleError, setInputTitleError] = useState(false);
+
+  const [userId, setUserId] = useState(0);
+  const [inputTitle, setInputTitle] = useState('');
+
+  const resetFunc = () => {
+    setUserId(0);
+    setInputTitle('');
+  };
+
+  const handleInputChange = (newInput: string) => {
+    setInputTitleError(false);
+    setInputTitle(newInput);
+  };
+
+  const handleUserChange = (newUser: number) => {
+    setUserIdError(false);
+    setUserId(newUser);
+  };
+
+  const handleAddClick = (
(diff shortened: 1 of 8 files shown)


VitaliyBondarenko1982 (changes requested):
Looks good.
Demo link leads to 404 page.
Looks like wasn't any deploy yet
https://github.com/Dmytrosp/react_add-todo-form/actions

Please deploy your work and check demo link before next re request.

Dmytrosp:
[DEMO LINK](https://Dmytrosp.github.io/react_add-todo-form/)

src/App.tsx
```diff
@@ -1,61 +1,109 @@
...
+  const [userId, setUserId] = useState(0);
+  const [inputTitle, setInputTitle] = useState('');
+
+  const resetFunc = () => {
+    setUserId(0);
+    setInputTitle('');
+  };
+
+  const handleInputChange = (newInput: string) => {
+    setInputTitleError(false);
```
Andriy-Shymkiv, line 24: need to set errors state to `false` only after first touch, not after every character

src/App.tsx
```diff
@@ -1,61 +1,109 @@
...
+    setInputTitle(newInput);
+  };
+
+  const handleUserChange = (newUser: number) => {
+    setUserIdError(false);
```
Andriy-Shymkiv, line 29: same here

src/App.tsx
```diff
@@ -1,61 +1,109 @@
...
+      setUserIdError(true);
+    }
+
+    if (!inputTitle.length) {
+      setInputTitleError(true);
+    }
+
+    if (!inputTitle.length || !userId) {
+      return;
+    }
```
Andriy-Shymkiv, line 48:
```suggestion
    if (!userId) {
      setUserIdError(true);
      return;
    }

    if (!inputTitle.length) {
      setInputTitleError(true);
      return;
    }
```

src/App.tsx
```diff
@@ -1,61 +1,109 @@
...
+
+    const newTodo = {
+      id: [...todos]
+        .sort((todo1, todo2) => todo2.id - todo1.id)[0].id + 1,
```
Andriy-Shymkiv, line 52: you can just take the id of the last todo in array and increase it +1 (no need to sort)

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,16 @@
-export const UserInfo = () => {};
+import React from 'react';
+import { User } from '../../types/User';
+
+type Props = {
+  user: User
```
Andriy-Shymkiv, line 5: user can be undefined

Andriy-Shymkiv (changes requested):

etojeDenys (changes requested):
image

PR #2342, opened 2024-02-05
add task solution

YevhenProtasov:
[DEMO LINK](https://YevhenProtasov.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,114 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { useState } from 'react';
+import usersFromServer from './api/users';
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+          >
+            <option
+              value="0"
+              disabled
+            >
+              Choose a user
+            </option>
+
+            {usersFromServer.map((user, index) => (
+              <option value={index + 1} key={user.id}>
```
volodymyr-soltys97, line 105: Also use the destruction for `user`
```suggestion
            {usersFromServer.map((user) => (
              <option value={user.name} key={user.id}>
```

volodymyr-soltys97 (approved):
Good job 👍

PR #2346, opened 2024-02-07
Todo Form. Solution

momos1703:
[DEMO LINK](https://momos1703.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,174 @@
+import { useState } from 'react';
 import './App.scss';
+import users from './api/users';
+import todos from './api/todos';
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+const initialUser: User = {
+  id: 0,
+  name: '',
+  username: '',
+  email: '',
+};
+const initialTodo: Todo = {
+  id: 0,
+  title: '',
+  completed: false,
+  userId: 0,
+  user: initialUser,
+};
+const findUser = (id: number): User => (
+  users.filter(user => user.id === id)[0]
+);
+const setId = (todoList: Todo[]) => {
+  const id = Math.max(
+    ...todoList.map(todo => todo.id),
+  );
+
+  return id + 1;
+};
+
+const todoListWithUsers = todos.map(todo => (
+  {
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,61 +1,184 @@
...
+};
+const initialTodo: Todo = {
+  id: 0,
+  title: '',
+  completed: false,
+  userId: 0,
+  user: initialUser,
+};
+const findUser = (id: number): User => (
+  users.filter(user => user.id === id)[0]
```
etojeDenys, line 23: better to use the find method, 'cause you want to find the user

etojeDenys (changes requested):
try to fix this test
image

or don't hesitate to ask for some help with that in the chat

src/App.tsx
momos1703: Done!

momos1703:
> try to fix this test

There is no problem to fix it. The problem is in ability to adding titles with using spaces only.

src/App.tsx
```diff
@@ -1,61 +1,192 @@
...
+};
+const initialTodo: Todo = {
+  id: 0,
+  title: '',
+  completed: false,
+  userId: 0,
+  user: initialUser,
+};
+const findUser = (id: number): User => (
+  users.find(user => user.id === id) || initialUser
```
etojeDenys, line 24:
```suggestion
  users.find(user => user.id === id) || null
```

src/types/Todo.ts
```diff
@@ -0,0 +1,9 @@
+import { User } from './User';
+
+export interface Todo {
+  id: number;
+  title: string;
+  completed: boolean;
+  userId: number;
+  user: User;
```
etojeDenys, line 8:
```suggestion
  user: User | null;
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,27 @@
...
+        'TodoInfo--completed': todoInfo.completed,
+      })}
+    >
+      <h2 className="TodoInfo__title">
+        {todoInfo.title}
+      </h2>
+
+      <UserInfo
+        user={todoInfo.user}
+      />
```
etojeDenys, line 24: and add the check here, if the user exists

etojeDenys (approved):
good job

PR #2347, opened 2024-02-08
react add todo form

TVChulkova:
[DEMO LINK](https://TVChulkova.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,1 +1,24 @@
-export const TodoInfo = () => {};
+import cn from 'classnames';
+import { UserInfo } from '../UserInfo';
+import { Todo } from '../../types/Todo';
+
+type Props = {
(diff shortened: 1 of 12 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,24 @@
...
+import cn from 'classnames';
+import { UserInfo } from '../UserInfo';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todo: Todo,
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  return (
```
etojeDenys, line 10: better to move logic with the user here
image

and if the user doesn't exist then do not render the UserInfo

etojeDenys (changes requested):
i can create the todo only with spaces, better to fix it
image

anastasiiavorobiova (approved):
Awesome work!

PR #2364, opened 2024-02-19
Solution

Krzychu92:
https://krzychu92.github.io/react_add-todo-form


src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+}
+
+export const tasks = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
 
 export const App = () => {
+  const [taskUser, setTaskUser] = useState('0');
+  const [task, setTask] = useState('');
```
yevhenii-pyl, line 22: `task` is an object, right? 
You can't set a string as a default value to something that should contain an object. Default value should be empty object or, if you want to emphasise that it is empty, you may choose undefined.

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+  const [errorTitle, setTitleError] = useState(false);
```
yevhenii-pyl, line 23: same here, 
you probably want to place a string here when there is an error, thus it should be either '' or undefined.

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+  const [errorUser, setUserError] = useState(false);
+  const [todos, setTasks] = useState(tasks);
```
yevhenii-pyl, line 25: tasks here a based on some API response that was additionally parsed and enhanced with user property. You can't use them as a default value. In theory, it's something async that will be added later. 
Default value should be just an empty array.

src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+    setTitleError(false); // Reset error when user starts typing again
+  };
+
+  const handleUserChange = (event: ChangeEvent<HTMLSelectElement>) => {
+    setTaskUser(event.target.value);
+    setUserError(false); // Reset error when user starts typing again
+  };
+
+  const handleSubmit = (event: { preventDefault: () => void; }) => {
+    event.preventDefault();
```
yevhenii-pyl, line 39: Looks like you have already extracted preventDefault function from event.

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,30 @@
...
+import { Todo } from '../../Types/Todo';
+import './TodoInfo.scss';
+import usersFromServer from '../../api/users';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
```
yevhenii-pyl, line 12: Good job, you can and should extract properties here or at the highest level.

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,14 @@
-export const UserInfo = () => {};
+import { User } from '../../Types/User';
+import './UserInfo.scss';
+
+type Props = {
+  user: User;
+};
+
+export const UserInfo = ({ user }: Props) => {
```
yevhenii-pyl, line 8: Let's extract name and mail here.

yevhenii-pyl (changes requested):
Good job, a few adjustments and it's good to go 👍 

As a general advice here: please, consider your variable naming. It may take some time but always worth it.

Krzychu92:
hey, I have problem with some conflicts but I cannot fix it. May you help me ? I was trying to change some things what you commented but now I cant fix that with conflicts

Zibi95 (changes requested):
Hey Krzysztof,
I noticed this is a really old PR. I assume the conflicts shouldn't be too hard to resolve. It seems that most of the changes during this time were related to the config files, including an upgrade to Node 20 and a change in the build tool. There's a QA session today at 6 PM, or you could simply create a new fork.

PR #2370, opened 2024-02-22
solution

AlinaOvod:
[DEMO LINK](https://AlinaOvod.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/PostForm/PostForm.tsx
+++ b/src/components/PostForm/PostForm.tsx
@@ -1,0 +1,116 @@
+import React, { useState } from 'react';
+import { Todo } from '../../Types/Todo';
+import usersFromServer from '../../api/users';
+
+type Props = {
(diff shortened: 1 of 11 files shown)


etojeDenys (changes requested):
image

src/components/PostForm/PostForm.tsx
```diff
@@ -0,0 +1,117 @@
...
+          onChange={handleUserChange}
+        >
+          <option value="0" disabled>Choose a user</option>
+
+          {usersFromServer.map(user => (
+            <option
+              value={user.id}
+              key={user.id}
+            >
+              {user.name}
```
volodymyr-soltys97, line 104: You can use destruction for `user`

volodymyr-soltys97 (approved):
Good job 👍

PR #2381, opened 2024-03-02
solution

HolyRaveOS:
[DEMO LINK](https://HolyRaveOS.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,34 @@
-import './App.scss';
+import { useState } from 'react';
+import { TodoList } from './components/TodoList';
+import { TodoForm } from './components/TodoForm';
+import { Todo } from './types/todo';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import './App.scss';
+import 'bulma/css/bulma.css';
+
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+const getUserById = (userId: number) =>
+  usersFromServer.find(user => user.id === userId) || null;
+
+const InitialTodos = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
 
 export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>(InitialTodos);
+
+  const addTodo = (newTodo: Todo) => {
+    setTodos([...todos, newTodo]);
+  };
+
   return (
     <div className="App">
-      <h1>Add todo form</h1>
-
-      <form action="/api/todos" method="POST">
-        <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
-        </div>
-
-        <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>
-              Choose a user
-            </option>
-          </select>
-
-          <span className="error">Please choose a user</span>
-        </div>
-
-        <button type="submit" data-cy="submitButton">
-          Add
-        </button>
-      </form>
-
-      <section className="TodoList">
-        <article data-id="1" className="TodoInfo TodoInfo--completed">
-          <h2 className="TodoInfo__title">delectus aut autem</h2>
-
-          <a className="UserInfo" href="mailto:Sincere@april.biz">
-            Leanne Graham
-          </a>
-        </article>
-
-        <article data-id="15" className="TodoInfo TodoInfo--completed">
-          <h2 className="TodoInfo__title">delectus aut autem</h2>
-
-          <a className="UserInfo" href="mailto:Sincere@april.biz">
-            Leanne Graham
-          </a>
-        </article>
-
-        <article data-id="2" className="TodoInfo">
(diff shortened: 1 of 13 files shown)


src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,114 @@
...
+          <span className="error">Please enter a title</span>
+        )}
+      </div>
+
+      <div className="field">
+        <label className="select is-block is-fullwidth">
+          <select
+            data-cy="userSelect"
+            value={user}
+            onChange={e => setUser(e.target.value)}
```
nk8bnj, line 86: The same.

nk8bnj (approved):
Good job!

src/App.tsx
```diff
@@ -1,61 +1,34 @@
...
-        </article>
+export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>(InitialTodos);
 
-        <article data-id="2" className="TodoInfo">
-          <h2 className="TodoInfo__title">
-            quis ut nam facilis et officia qui
-          </h2>
+  const addTodo = (newTodo: Todo) => {
+    setTodos([...todos, newTodo]);
```
BudnikOleksii, line 24: If you need to update the state depending on the previous one use the functional `setState`

src/components/TodoForm/TodoForm.scss
```diff
@@ -0,0 +1,4 @@
+.todo-form {
+  max-width: 300px;
+
+}
```
BudnikOleksii, line 4: Remove unnecessary blank line and add one at the end of the file

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+export const TodoForm: React.FC<Props> = ({ todos, users, onSubmit }) => {
+  const [title, setTitle] = useState('');
+  const [user, setUser] = useState('0');
+  const [submitting, setSubmitting] = useState(false);
+
+  const usersNames = users.map(({ name }) => name);
+
+  const isTitleError = title === '';
+  const isUserError = user === '0';
```
BudnikOleksii, line 22: You don't need to compare with an empty line or with 0, just check if a value exists.
https://developer.mozilla.org/en-US/docs/Glossary/Falsy

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,114 @@
...
+interface Props {
+  todos: Todo[];
+  users: User[];
+  onSubmit: (newTodo: Todo) => void;
+}
+
+export const TodoForm: React.FC<Props> = ({ todos, users, onSubmit }) => {
+  const [title, setTitle] = useState('');
+  const [user, setUser] = useState('0');
+  const [submitting, setSubmitting] = useState(false);
```
BudnikOleksii, line 17:
```suggestion
  const [isSubmitting, setIsSubmitting] = useState(false);
```
Boolean variable should have prefix `is` or `has`

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+interface Props {
+  todos: Todo[];
+  users: User[];
+  onSubmit: (newTodo: Todo) => void;
+}
+
+export const TodoForm: React.FC<Props> = ({ todos, users, onSubmit }) => {
+  const [title, setTitle] = useState('');
+  const [user, setUser] = useState('0');
```
BudnikOleksii, line 16: It looks userId, not user itself

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,114 @@
...
+  const [submitting, setSubmitting] = useState(false);
+
+  const usersNames = users.map(({ name }) => name);
+
+  const isTitleError = title === '';
+  const isUserError = user === '0';
+
+  const handleTitleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
```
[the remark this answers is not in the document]
BudnikOleksii: > Screenshot 2024-03-02 at 19 56 59

Good point

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,114 @@
...
+    const inputValue = e.target.value;
+    const pattern = /^[A-Za-zА-ЩЇІЄҐа-щїієґ0-9\s]+$/;
+
+    const cleanedInput = inputValue
+      .split('')
+      .filter(char => pattern.test(char) || char === ' ')
+      .join('');
```
BudnikOleksii, line 31: - With regex, you can test all value, you don't need to split, filter, etc
- Also, if you need to get rid of invalid characters you can use replace, it works with regex 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace

BudnikOleksii (changes requested):
Good work 👍 
But could be even better, check the suggestions

BudnikOleksii (approved):
👍

PR #2388, opened 2024-03-06
Develop

dmitruz:
[DEMO LINK](https://dmitruz.github.io/react_add-todo-form/)


src/type/Type.ts
```diff
@@ -0,0 +1,14 @@
...
+  userId: number;
+  user: User | null;
+}
+
+export interface User {
+  id: number;
+  name: string;
+  username: string;
+  email: string;
+}
```
SanyaBratashchuk, line 14: split these types into different files

SanyaBratashchuk (changes requested):
1. Improve validation I can add todo with just spaces 
image

src/App.tsx
```diff
@@ -1,61 +1,102 @@
...
+          <label htmlFor="author">Author: </label>
+          <select
+            data-cy="userSelect"
+            value={userId}
+            onChange={handleUserIdChange}
+          >
+            <option value="0">Choose a user</option>
+            {usersFromServer.map(user => (
+              <option key={user.id} value={user.id}>
+                {user.name}
```
volodymyr-soltys97, line 86: You can use destruction for `user`

volodymyr-soltys97 (changes requested):
GJ!
Let's improve your code
This part of the task is incomplete
You need to show the error message for the author
image
image

src/App.tsx
```diff
@@ -1,61 +1,104 @@
...
+      setTitleError(false);
+    }
+  };
+
+  const handleUserIdChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setUserId(+event.target.value);
+
+    if (userError) {
+      setUserError(false);
+    }
```
vadiimvooo, line 27:
```suggestion
     setUserError(false);
```

src/App.tsx
```diff
@@ -1,61 +1,104 @@
...
+
+  const handleChange = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    const trimedTitle = title.trim();
+
+    setTitleError(!trimedTitle);
+    setUserError(!trimedTitle);
+
+    if (!trimedTitle || userId === 0) {
```
vadiimvooo, line 43:
```suggestion
    if (!trimedTitle || !userId) {
```

src/App.tsx
```diff
@@ -1,61 +1,104 @@
...
+    setTitle('');
+  };
+
+  const handleChange = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    const trimedTitle = title.trim();
+
+    setTitleError(!trimedTitle);
+    setUserError(!trimedTitle);
```
vadiimvooo, line 41: Here you should check if user selected or not. You don't need to use title value.

vadiimvooo (changes requested):
Almost done. A few changes required.

SanyaBratashchuk (approved):

PR #2391, opened 2024-03-07
added solution

shikoryak:
https://shikoryak.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,98 @@
+import { useCallback, useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import { TodoList } from './components/TodoList';
+import todosFromServer from './api/todos';
+import { Todo, User } from './types';
 
-export const App = () => {
+const defaultFormValue = {
+  title: '',
+  userId: 0,
+};
+
+export const App: React.FC = () => {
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,61 +1,106 @@
...
+          <input
+            type="text"
+            id="titleInput"
+            value={formValue.title}
+            data-cy="titleInput"
+            placeholder="Enter a title"
+            onChange={event => {
+              updateFormValue('title', event.target.value);
+              setTitleTouched(false);
+            }}
```
lerastarynets, line 68: Pls move the function inside `onChange` handler to a separate const and use as a callback here
**PLS DO THE SAME FOR ALL SUCH CASES IN THE PROJECT**

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,24 @@
...
+
+  return (
+    <article
+      data-id={todo.id}
+      className={cn('TodoInfo', {
+        'TodoInfo--completed': todo.completed,
+      })}
+    >
+      <h2 className="TodoInfo__title">{todo.title}</h2>
+      {user ? <UserInfo user={user} /> : null}
```
lerastarynets, line 21:
```suggestion
      {user && <UserInfo user={user} /> }
```

lerastarynets (changes requested):
addTodoForm

lerastarynets (approved):
You're doing amazing, keep it up!

PR #2396, opened 2024-03-11
Add solution

Vasya564:
[DEMO LINK](https://Vasya564.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/UserInfo/UserInfo.tsx
+++ b/src/components/UserInfo/UserInfo.tsx
@@ -1,1 +1,13 @@
-export const UserInfo = () => {};
+import { User } from '../../types/User';
+
+type Props = {
+  user: User | null;
+};
+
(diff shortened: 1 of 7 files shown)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,13 @@
-export const UserInfo = () => {};
+import { User } from '../../types/User';
+
+type Props = {
+  user: User | null;
```
BudnikOleksii, line 4: user should have type only `User`, it doesn't make sense to render this component without the user

BudnikOleksii (changes requested):
Looks good 🚀 
Just one moment that should be fixed

BudnikOleksii (approved):
👍

PR #2399, opened 2024-03-12
Develop

yurovych:
https://yurovych.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,116 +5,13 @@ import usersFromServer from './api/users';
 import todosFromServer from './api/todos';
 import { TodoList } from './components/TodoList';
 
-type UsersItem = {
-  id: number;
-  name: string;
-  username: string;
-  email: string;
-};
-
-interface AppProps {
-  users: UsersItem[];
-}
-
-export const App: React.FC<AppProps> = () => {
-  const selectInitialValue = 'Сhoose a user';
-  const [inputValue, setInputValue] = useState('');
-  const [selectValue, setSelectValue] = useState(selectInitialValue);
-  const [isInputError, setIsInputError] = useState(false);
-  const [isSelectError, setIsSelectError] = useState(false);
+export const App = () => {
   const [users] = useState(usersFromServer);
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,61 +1,40 @@
...
+  completed: boolean;
+  userId: number;
+  user?: User;
+};
+
+export const App: React.FC = () => {
+  const todos: TodoItem[] = todosFromServer;
+  const users: User[] = usersFromServer;
+  const todosWithUsers: TodoItem[] = todos.map(todo => {
+    const currentUser = users.find(u => u.id === todo.userId);
```
volodymyr-soltys97, line 25:
```suggestion
    const currentUser = users.find(user => user.id === todo.userId);
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,36 @@
...
+export const TodoInfo: React.FC<TodoInfoProps> = ({ todo }) => {
+  return (
+    <article
+      key={todo.id}
+      data-id={todo.id}
+      className={classNames('TodoInfo', {
+        'TodoInfo--completed': todo.completed,
+      })}
+    >
+      <h2 className="TodoInfo__title">{todo.title}</h2>
```
volodymyr-soltys97, line 32: Use destruction for `todo`

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,155 @@
...
+          </span>
+        </div>
+
+        <button type="submit" data-cy="submitButton">
+          Add
+        </button>
+      </form>
+      <section className="TodoList">
+        {todosData.map(item => (
+          <TodoInfo todo={item} />
```
volodymyr-soltys97, line 150: Don't forget about the key attribute

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better
I can't add only spaces as title, so you need to fix it, use `trim()` method
image

SanyaBratashchuk (approved):

PR #2402, opened 2024-03-13
solution

ChrisHryts:
https://chrishryts.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,38 @@
+import React, { useState } from 'react';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { TodoForm } from './components/TodoForm/TodoForm';
+import { Todo } from './components/types/Todo';
+import { getUserById } from './services/getUserById';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+const initialTodos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
 
-export const App = () => {
+function getNewTodoId(todos: Todo[]) {
+  return Math.max(...todos.map(todo => todo.id)) + 1;
+}
+
+export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>(initialTodos);
+
+  const addTodo = (todo: Todo) => {
+    const newTodo = {
+      ...todo,
+      id: getNewTodoId(todos),
+    };
+
+    setTodos(currentTodos => [...currentTodos, newTodo]);
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/todos" method="POST">
-        <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
-        </div>
-
-        <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>
-              Choose a user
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,61 +1,38 @@
...
-            <option value="0" disabled>
-              Choose a user
-            </option>
-          </select>
-
-          <span className="error">Please choose a user</span>
-        </div>
+function getNewTodoId(todos: Todo[]) {
+  return Math.max(...todos.map(todo => todo.id)) + 1;
+}
```
HolyRaveOS, line 16: You can move it to a separate file as you did with getUserById

src/App.tsx
```diff
@@ -1,61 +1,38 @@
...
+      ...todo,
+      id: getNewTodoId(todos),
+    };
 
-          <a className="UserInfo" href="mailto:Sincere@april.biz">
-            Leanne Graham
-          </a>
-        </article>
+    setTodos(currentTodos => [...currentTodos, newTodo]);
+  };
```
HolyRaveOS, line 28: You can do it like this 
```
  const addTodo = (newTodo: Todo) => {
        setTodos(prevTodos => [...prevTodos, newTodo]);
  };
  ```

File: src/components/TodoForm/TodoForm.tsx [hunk 3]
```diff
@@ -0,0 +1,115 @@
...
+type Props = {
+  onSubmit: (todo: Todo) => void;
+};
+
+export const TodoForm: React.FC<Props> = ({ onSubmit }) => {
+  const [title, setTitle] = useState('');
+  const [hasTitleError, setHasTitleError] = useState(false);
+
+  const [userId, setUserId] = useState(0);
+  const [hasUserIdError, setHasUserIdError] = useState(false);
```
HolyRaveOS, line 16: I'm not sure that you need state for title and id error.
You can check it like that.

```
  const isTitleError = !title;
  const isUserError = !userId;
```
And than you can handle id change like that
```
   <select
            data-cy="userSelect"
            value={userId}
            onChange={event => setUserId(event.target.value)}
          >
          ```
          instead of handleUserIdChange

File: src/components/types/Todo.ts [hunk 4]
```diff
@@ -0,0 +1,9 @@
+import { User } from './User';
+
+export type Todo = {
+  id: number;
+  title: string;
+  completed?: boolean;
+  userId: number;
+  user?: User;
```
HolyRaveOS, line 8: user can't be optional. It's a required value.

HolyRaveOS (changes requested):
Good job!
 There is small fix that need to be done.
 First 3 comments is optional)
 Pay attention to the last one

src/components/types/Todo.ts
ChrisHryts: its not required as we are checking in TodoInfo if user exists

image

src/App.tsx
ChrisHryts: this method skips adding the ID

src/App.tsx
BudnikOleksii: > You can move it to a separate file as you did with getUserById

Good point

src/components/TodoForm/TodoForm.tsx
BudnikOleksii: > const isTitleError = !title;

If you do ` const isTitleError = !title;` a user will see an error message even they did nothing, so it's ok

BudnikOleksii (approved):
Looks great 🔥

PR #2404, opened 2024-03-13
add solution

RusLan2186:
[DEMO LINK](https://RusLan2186.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -80,7 +80,9 @@ export const App = () => {
             onChange={handleUserIdChange}
             data-cy="userSelect"
           >
-            <option value="0">Choose a user</option>
+            <option value="0" disabled={userId !== 0}>
+              Choose a user
+            </option>
             {usersFromServer.map(user => (
               <option key={user.id} value={user.id}>
                 {user.name}
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -8,14 +8,12 @@ type Props = {
 
 export const TodoInfo: React.FC<Props> = ({ todo }) => {
   return (
-    <>
-      <article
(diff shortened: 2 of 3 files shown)


src/App.tsx
```diff
@@ -1,61 +1,103 @@
...
-              Choose a user
-            </option>
+          <label htmlFor="userIdSelect">User: </label>
+          <select
+            value={userId}
+            id="userIdSelect"
+            onChange={handleUserIdChange}
+            data-cy="userSelect"
+          >
+            <option value="0">Choose a user</option>
```
ViktoriiaRepo, line 83: This option in demo is desabled after choosing some option. So you can change 
```js
 <option value="" disabled={userId !== 0}>

**Review by ViktoriiaRepo — APPROVED** [review]
Good work!) Approved, but check a comment to improve your solution

File: src/App.tsx (hunk 1 above)
Comment (BudnikOleksii) [reply src/App.tsx:83]: > This option in demo is desabled after choosing some option. So you can change
> 
> ```js
>  <option value="" disabled={userId !== 0}> 
> ```

Good point

File: src/components/TodoInfo/TodoInfo.tsx [hunk 2]
```diff
@@ -1 +1,21 @@
...
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  return (
+    <>
```
BudnikOleksii, line 11: React fragment is redundant in this case

BudnikOleksii (changes requested):
Looks good 👍 
Let's make a few improvements in accordance with suggestions

BudnikOleksii (approved):
👍

PR #2409, opened 2024-03-15
solution

iamgranovskaya:
[DEMO](https://iamgranovskaya.github.io/react_add-todo-form/)


anastasiiavorobiova (changes requested):
Great work! Avoid adding an empty todo with spaces only 
Screenshot 2024-03-15 at 14 36 42

PR #2416, opened 2024-03-17
task solution

Devery11:
[DEMO LINK](https://devery11.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,1 +1,21 @@
-export const TodoInfo = () => {};
+import React from 'react';
+import { Todo } from '../../types/todo';
+import classNames from 'classnames';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => (
+  <article
+    data-id={todo.id}
+    className={classNames('TodoInfo', {
+      'TodoInfo--completed': todo.completed,
+    })}
+  >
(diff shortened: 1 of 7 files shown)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,12 @@
...
+
+type Props = {
+  user: User | undefined;
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => (
+  <a className="UserInfo" href={`mailto:${user?.email}`}>
+    {user?.name}
+  </a>
+);
```
etojeDenys, line 12:
```suggestion
type Props = {
  user: User;
};

export const UserInfo: React.FC<Props> = ({ user }) => (
  <a className="UserInfo" href={`mailto:${user.email}`}>
    {user.name}
  </a>
);
```
let's render this component only when a user exists

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,21 @@
...
+export const TodoInfo: React.FC<Props> = ({ todo }) => (
+  <article
+    data-id={todo.id}
+    className={classNames('TodoInfo', {
+      'TodoInfo--completed': todo.completed,
+    })}
+  >
+    <h2 className="TodoInfo__title">{todo.title}</h2>
+
+    <UserInfo user={todo.user} />
```
etojeDenys, line 19: and add the check here
```suggestion
    {todo.user && <UserInfo user={todo.user} />}
```

etojeDenys (changes requested):
image

PR #2422, opened 2024-03-19
add solution

yaGodya:
# React Add TODO Form

Implement the ability to add TODOs to the `TodoList` implemented in the **Static List of TODOs**
> Here is [the working example](https://mate-academy.github.io/react_add-todo-form/)

1. Create an `App` component storing the `todos` array and displaying it using the `TodoList`.
1. Create a form to add new TODOs:
    - there should be a text input for the `title` with `data-cy="titleInput"` attribute;
    - add a `<select>` with `data-cy="userSelect"` attribute showing all the given users;
    - add labels and placeholders where they are needed;
    - add a new todo to the list after clicking the `Add` button;
    - each TODO should have `id`, `title`, `userId`, and `completed` (`false` by default);
    - `id` is the largest `id` in the array + 1 (add `data-id={todo.id}` attribute to each `.TodoInfo`).
1. Add validation to the form:
    - add a default empty option `Choose a user` to the select;
    - before creating a todo, check if a `user` was selected; if not, show an error message next to the `select` (`Please choose a user`);
    - if the `title` is empty, show an error message next to the `title` field (`Please enter a title`);
    - errors should appear only after clicking the `Add` button;
    - hide the message immediately after any change of the field with an error;
1. If the form is valid, add a todo to the list and clear the form.
1. (* **Optional**) Allow entering only letters (`ua` and `en`), digits, and `spaces` in the `title` field. Just remove any other characters from the `title`.

## Instructions
 [DEMO LINK](https://yaGodya.github.io/react_add-todo-form/)


tetlisna (approved):
Looks good! add labels to input and select tags

src/App.tsx
```diff
@@ -1,61 +1,138 @@
...
+import { useState } from 'react';
+
+export interface User {
+  id: number;
+  name: string;
+  username: string;
+  email: string;
+}
+
+export interface Todo {
```
BudnikOleksii, line 15: Move interfaces to the types folder/file

src/App.tsx
```diff
@@ -1,61 +1,138 @@
...
+const inTodo: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [isEmptyTitle, setIsEmptyTitle] = useState(false);
+  const [userId, setUserId] = useState(0);
+  const [isSelectUser, setIsSelectUser] = useState(false);
```
BudnikOleksii, line 42: The name `isSelectUser` but you set it to true when actually user isn't selected 🤷

BudnikOleksii (approved):
Good work 👍

PR #2423, opened 2024-03-19
add task solution

TetianaBukhenko:
[DEMO LINK](https://TetianaBukhenko.github.io/react_add-todo-form/)

### DIFF
--- a/src/functions.tsx
+++ b/src/functions.tsx
@@ -1,0 +1,78 @@
+import users from './api/users';
+import { TodoWithUser } from './components/types/TodoWithUser';
+import { User } from './components/types/User';
+import { Todo } from './components/types/Todo';
+import { TodoError } from './components/types/TodoError';
+
+export function getUserById(userId: number): User {
+  return [...users].filter(currentUser => currentUser.id === userId)[0];
+}
+
+export function getTodosWithUsers(todos: Todo[]): TodoWithUser[] {
+  return todos.map(todo => {
+    const { userId } = todo;
+
+    return {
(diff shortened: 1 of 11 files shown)


src/functions.tsx
```diff
@@ -0,0 +1,78 @@
+import users from './api/users';
+import { TodoWithUser } from './components/types/TodoWithUser';
+import { User } from './components/types/User';
+import { Todo } from './components/types/Todo';
+import { TodoError } from './components/types/TodoError';
+
+export function getUserById(userId: number): User {
+  return [...users].filter(currentUser => currentUser.id === userId)[0];
```
etojeDenys, line 8: better to use the find method, 'cause you want to find the user, not filter the array

etojeDenys (changes requested):
image
i can create the empty todo(only with spaces), let's fix that

src/functions.tsx
```diff
@@ -0,0 +1,115 @@
+import users from './api/users';
+import { TodoWithUser } from './components/types/TodoWithUser';
+import { User } from './components/types/User';
+import { Todo } from './components/types/Todo';
+import { TodoError } from './components/types/TodoError';
+
+export function getUserById(userId: number): User | null {
+  return [...users].find(currentUser => currentUser.id === userId) || null;
```
SanyaBratashchuk, line 8:
```suggestion
  return users.find(currentUser => currentUser.id === userId) || null;
```

SanyaBratashchuk (approved):

PR #2429, opened 2024-03-19
solution

IRONH4MMER:
[DEMO LINK](https://IRONH4MMER.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,135 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/Todo';
+import { User } from './types/User';
+import { useState } from 'react';
+
+interface serverTodo {
+  completed: boolean;
+  id: number;
+  title: string;
+  userId: number;
+}
+
+function getUserFromId(todo: serverTodo): User {
+  return usersFromServer.find(user => user.id === todo.userId)!;
+}
+
+const preparedTodos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserFromId(todo),
+}));
+
+const regex = /^[a-zA-Za-яА-ЯiІіЇїґҐєЄ0-9\s]+$/;
 
 export const App = () => {
+  const [todosArray, setTodosArray] = useState<Todo[]>(preparedTodos);
+
+  const [selectedUser, setSelectedUser] = useState(0);
+  const [hasSelectedUserError, setHasSelectedUserError] = useState(false);
+  const [isDefaultDisabled, setIsDefaultDisabled] = useState(false);
+
+  const [title, setTitle] = useState('');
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,144 @@
...
+import { useState } from 'react';
+
+interface ServerTodo {
+  completed: boolean;
+  id: number;
+  title: string;
+  userId: number;
+}
+
+function getUserFromId(todo: ServerTodo): User {
```
BudnikOleksii, line 17:
```suggestion
function getUserById(todo: ServerTodo): User {
```

src/App.tsx
```diff
@@ -1,61 +1,144 @@
...
+  const [hasTitleError, setHasTitleError] = useState(false);
+
+  const handleSelectedUserError = () => {
+    if (selectedUser === 0) {
+      setHasSelectedUserError(true);
+    }
+  };
+
+  const handleHasTitleError = () => {
+    if (title === '') {
```
BudnikOleksii, line 45:
```suggestion
    if (!title) {
```
Don't compare with an empty string, check if a value exists

src/App.tsx
```diff
@@ -1,61 +1,144 @@
...
+    event: React.ChangeEvent<HTMLSelectElement>,
+  ) => {
+    setSelectedUser(+event.target.value);
+    setHasSelectedUserError(false);
+    setIsDefaultDisabled(true);
+  };
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    if (regex.test(event.target.value[event.target.value.length - 1])) {
+      setTitle(event.target.value);
```
BudnikOleksii, line 63: - You get `event.target.value` 3 times, use destructing
- For the future, if you have some validations, you should notify user in case of invalid input

src/App.tsx
```diff
@@ -1,61 +1,144 @@
...
+    setHasTitleError(false);
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    handleSelectedUserError();
+    handleHasTitleError();
+
+    if (selectedUser === 0 || title === '') {
```
BudnikOleksii, line 84:
```suggestion
    if (!selectedUser || !title) {
```

src/App.tsx
```diff
@@ -1,61 +1,144 @@
...
+      return;
+    }
+
+    addTodo({
+      completed: false,
+      id: Math.max(...todosArray.map(todo => todo.id)) + 1,
+      title: title,
+      user: usersFromServer.find(user => user.id === selectedUser) as User,
```
BudnikOleksii, line 92: Don't use `as`

BudnikOleksii (changes requested):
Good work 👍 
Just a few minor fixes are required

src/App.tsx
IRONH4MMER: What should I use here instead of `as`?

BudnikOleksii: > What should I use here instead of `as`?

Actual user or null. `as` is a bad practice, you trick the ts but it can cause future errors

BudnikOleksii (approved):
👍

PR #2437, opened 2024-03-20
first save

VadimDrobyazko:
[DEMO LINK](https://vadimdrobyazko.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,1 +1,25 @@
-export const TodoInfo = () => {};
+import cn from 'classnames';
+
+import { Todos } from '../../types/todos';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todos;
+};
+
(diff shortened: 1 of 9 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,26 @@
...
+      data-id={todo.id}
+      className={cn('TodoInfo', {
+        'TodoInfo--completed': todo.completed,
+      })}
+    >
+      <h2 className="TodoInfo__title" data-cy="titleInput">
+        {todo.title}
+      </h2>
+
+      <UserInfo user={todo.user as Users} />
```
IvanFesenko, line 23: user is possibly null, add check if user exists instead if `as`
```suggestion
      <UserInfo user={todo.user} />
```

IvanFesenko (changes requested):
- Let show user that only spaces is not valid input 
2024-03-22_17h01_59

vadiimvooo (approved):
LGTM

PR #2445, opened 2024-03-21
add task solution

SerjDSA:
[DEMO LINK](https://serjdsa.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,1 +1,139 @@
-export const TodoList = () => {};
+import { useState } from 'react';
+import { TodoInfo } from '../TodoInfo';
+import usersFromServer from '../../api/users';
+import { TodoItem } from '../../types/TodoItem';
+
+interface TodoListProps {
+  todos: TodoItem[];
+}
+
+export const TodoList: React.FC<TodoListProps> = ({ todos }) => {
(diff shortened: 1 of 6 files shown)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,17 @@
...
+import { User } from '../../types/User';
+
+interface Props {
+  user: User;
+}
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  if (!user) {
+    return null;
+  }
```
etojeDenys, line 10:
```suggestion

```
looks redundant here, 'cause you set in the Props the user cannot be falsy

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,139 @@
...
+    title: '',
+    completed: false,
+    userId: 0,
+    user: {
+      id: 0,
+      name: '',
+      username: '',
+      email: '',
+    },
+  });
```
etojeDenys, line 31: you don't need this state, just create the newCard var in the handleSubmit

etojeDenys (changes requested):

SerjDSA (commented):
[DEMO LINK](https://serjdsa.github.io/react_add-todo-form/)

maxim2310 (approved):
Good job!!!

PR #2453, opened 2024-03-21
Develop

pawelnowicki87:
[DEMO LINK](https://pawelnowicki87.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,134 @@
+import React, { ChangeEvent, FormEvent, useState } from 'react';
+import fetchedUsers from './api/users';
+import fetchedTodos from './api/todos';
+import { User } from './Types';
+import { TodoList } from './components/TodoList';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+const todosWithAssociatedUsers = fetchedTodos.map(todo => ({
+  ...todo,
+  user: fetchedUsers.find(user => user.id === todo.userId),
+}));
 
 export const App = () => {
+  const [todoItems, setTodoItems] = useState(todosWithAssociatedUsers);
+  const [formValues, setFormValues] = useState({ title: '', user: '' });
+  const [isValidTitle, setIsValidTitle] = useState(false);
+  const [isValidUser, setIsValidUser] = useState(false);
+  const [displayTitleError, setDisplayTitleError] = useState(false);
+  const [displayUserError, setDisplayUserError] = useState(false);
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,61 +1,142 @@
+import React, { ChangeEvent, FormEvent, useState } from 'react';
+import fetchedUsers from './api/users';
+import fetchedTodos from './api/todos';
+import { User } from './Types';
+import { TodoList } from './components/TodoList';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+const todosWithAssociatedUsers = fetchedTodos.map(todo => ({
```
yevhenii-pyl, line 8: Nice naming here. Some may say that it's too long, but I can call it readable and easy to grasp :)

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,10 @@
-export const UserInfo = () => {};
+import React from 'react';
+import { UserInfoProps } from '../../Types';
+
+export const UserInfo: React.FC<UserInfoProps> = ({ user }) => {
```
yevhenii-pyl, line 4: Don't forget to extract properties here, it's just a nice little approach to get used to:
`{ user: { email, name } }`

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,15 @@
-export const TodoInfo = () => {};
+import React from 'react';
+import { TodoInfoProps } from '../../Types';
+import { UserInfo } from '../UserInfo';
+
+export const TodoInfo: React.FC<TodoInfoProps> = ({ todo }) => {
```
yevhenii-pyl, line 5: Same with extraction here

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
+      }
+
+      if (!isValidUser) {
+        setDisplayUserError(true);
+      }
+
+      return;
+    }
+
+    const selectedUser = fetchedUsers.find(
```
yevhenii-pyl, line 61: What if we have a few users with exact same name? 
I believe, it's better to use userId here instead of name. This will require you to do some refactoring with selector options and form values state - but this will make your code more reliable :)

yevhenii-pyl (changes requested):
Good and clean job! 
However, let's change your approach with setting user name to setting user Id.

yevhenii-pyl (approved):
LGTM 🚀

PR #2458, opened 2024-03-23
add task solution

marichkamt:
[DEMO LINK](https://marichkamt.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,112 @@
+import { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { Todo } from './components/types';
+import { TodoList } from './components/TodoList';
+
+function getUserById(id: number) {
+  return usersFromServer.find(user => user.id === id) || null;
+}
+
+function getHighestTodoId(list: Todo[]) {
+  const listOfId = list.map(item => item.id);
+
+  return Math.max(...listOfId);
+}
+
+const defaultList = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
 
 export const App = () => {
+  const [todos, setTodos] = useState(defaultList);
+
+  const [title, setTitle] = useState('');
+  const [titleError, setTitleError] = useState(false);
+
+  const [userId, setUserId] = useState(0);
+  const [userError, setUserError] = useState(false);
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,61 +1,120 @@
...
+const defaultList = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
 
 export const App = () => {
+  const [todos, setTodos] = useState(defaultList);
+
+  const [title, setTitle] = useState('');
+  const [titleError, setTitleError] = useState(false);
```
GUSILLUS, line 28: Better use `is` to boolean state name

isLoading, isError, isSelected, etc

src/App.tsx
```diff
@@ -1,61 +1,120 @@
...
+
+  const [userId, setUserId] = useState(0);
+  const [userError, setUserError] = useState(false);
```
GUSILLUS, line 31: same

src/App.tsx
```diff
@@ -1,61 +1,120 @@
...
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
GUSILLUS, line 33: Don't forget that you can import event types from `'react'` like useState

src/App.tsx
```diff
@@ -1,61 +1,120 @@
...
+
+    if (!title.trim() || !userId) {
+      // eslint-disable-next-line
+      return;
+    }
+
+    setTodos(currentList => [
+      ...currentList,
+      {
+        id: getHighestTodoId(currentList) + 1,
```
GUSILLUS, line 57: Read about [crypto.randomUUID()](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) to use it in future when you need a random id

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,22 @@
...
+  const { id, title, user, completed } = todo;
+
+  return (
+    <article
+      data-id={id}
+      className={classNames('TodoInfo', { 'TodoInfo--completed': completed })}
+    >
+      <h2 className="TodoInfo__title">{title}</h2>
+
+      <UserInfo user={user} />
```
GUSILLUS, line 19: Can we do conditional rendering only if we have a user to secure?

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,14 @@
-export const UserInfo = () => {};
+import { User } from '../types';
+import './UserInfo.scss';
+
+type Props = {
+  user: User | null;
```
GUSILLUS, line 5: When we do conditional rendering we can write here only `user: User`

GUSILLUS (changes requested):
GJ! But:

PR #2471, opened 2024-03-31
React add todo form

Denys209818:
[DEMO LINK](https://denys209818.github.io/react_add-todo-form)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,126 @@
+import { useState } from 'react';
 import './App.scss';
+import { TodoList, TodoType, UserType } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+interface Errors {
+  title: string;
+  user: string;
+}
 
 export const App = () => {
+  const [title, setTitle] = useState<string>('');
+  const [author, setAuthor] = useState<string>('0');
+  const [errors, setErrors] = useState<Errors>({ title: '', user: '' });
+
+  const getTodos = (): TodoType[] => {
+    const todos = todosFromServer.map(item => {
+      return {
+        ...item,
+        user: usersFromServer.find(us => us.id === item.userId) as UserType,
+      };
+    });
+
+    return todos;
+  };
+
+  const setTitleCustomized = (value: string) => {
+    if (value.match(/^[a-zA-Z0-9\s]+$/g) || value === '') {
+      setTitle(value);
+    }
+  };
+
+  const [todos, setTodos] = useState<TodoType[]>(getTodos());
+
+  const onSubmit = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+
+    if (!title) {
+      setErrors(prev => {
+        return {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+import todosFromServer from './api/todos';
+
+interface Errors {
+  title: string;
+  user: string;
+}
 
 export const App = () => {
+  const [title, setTitle] = useState<string>('');
+  const [author, setAuthor] = useState<string>('0');
```
etojeDenys, line 15:
```suggestion
  const [title, setTitle] = useState('');
  const [author, setAuthor] = useState('0');
```
it's redundant to write types for primitives

src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+  const [errors, setErrors] = useState<Errors>({ title: '', user: '' });
+
+  const getTodos = (): TodoType[] => {
+    const todos = todosFromServer.map(item => {
+      return {
+        ...item,
+        user: usersFromServer.find(us => us.id === item.userId) as UserType,
```
etojeDenys, line 22:
```suggestion
        user: usersFromServer.find(us => us.id === item.userId) || null,
```
try to avoid assertions as much as you can

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,23 @@
...
+  return (
+    <article
+      data-id={todo.id}
+      className={classNames('TodoInfo', {
+        'TodoInfo--completed': todo.completed,
+      })}
+    >
+      <h2 className="TodoInfo__title">{todo.title}</h2>
+
+      <UserInfo user={todo.user} />
```
etojeDenys, line 20: user can be null, so let's add the check here
```suggestion
      {todo.user && <UserInfo user={todo.user} />}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,37 @@
...
+  return (
+    <section className="TodoList">
+      {todos.map(todo => {
+        return (
+          <React.Fragment
+            key={todo.id + Math.ceil(Math.random() * 1000) / 1000}
+          >
+            <TodoInfo todo={todo} />
+          </React.Fragment>
+        );
```
etojeDenys, line 33: fragment is redundant here

https://react.dev/reference/react/Fragment#usage
```suggestion
        return (
            <TodoInfo key={todo.id} todo={todo} />
        );
```

etojeDenys (changes requested):
image
i can create the todo only with spaces, let's fix that here too

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,36 @@
...
+}
+
+export const TodoList: React.FC<TodoListType> = ({ todos }) => {
+  return (
+    <section className="TodoList">
+      {todos.map(todo => {
+        return (
+          <TodoInfo
+            todo={todo}
+            key={todo.id + Math.ceil(Math.random() * 1000) / 1000}
```
etojeDenys, line 30:
```suggestion
            key={todo.id}
```
looks redundant

etojeDenys (approved):
lgtm

PR #2475, opened 2024-04-01
with optional task (validate title for en/ua letters, numbers and spaces)

IhorRome:
[DEMO LINK](https://IhorRome.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,152 @@
+import { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import { TodoList } from './components/TodoList';
+import { SaveTodoDTO } from './interfaces/saveTodoDto';
+import { SaveTodoError } from './interfaces/saveTodoError';
+import { Todo } from './interfaces/todo';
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,61 +1,160 @@
...
+              {
+                if (saveTodoError.user) {
+                  setSaveTodoErrorFn('user', '');
+                }
+
+                setSaveTodo(prev => {
+                  return { ...prev, userId: +event.target.value };
+                });
+              }
+            }}
```
VitaliyBondarenko1982, line 135: it is not good practice to make anonymous arrow functions in jsx.
better to make function in component and pass reference on it here.

src/App.tsx
```diff
@@ -1,61 +1,160 @@
...
+    field: keyof SaveTodoError,
+    value: string,
+  ): void => {
+    setSaveTodoError(prev => {
+      const newErr = { ...prev };
+
+      newErr[field] = value;
+
+      return newErr;
+    });
```
VitaliyBondarenko1982, line 88:
```suggestion
    setSaveTodoError(prev => ({
      ...prev,
     [field]: value,
    }));
```
same but shorter.

VitaliyBondarenko1982 (changes requested):
Good job.
Suggest a couple improvements

VitaliyBondarenko1982 (approved):
Well done

PR #2477, opened 2024-04-03
Add test

olehmarushchak:
[DEMO LINK] https://olehmarushchak.github.io/react_add-todo-form/

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -10,6 +10,7 @@ interface Props {
 
 export const TodoInfo: React.FC<Props> = ({ todo }) => {
   const { id, title, completed } = todo;
+  const user = findIdUser(todo);
 
   return (
     <article
@@ -18,7 +19,7 @@ export const TodoInfo: React.FC<Props> = ({ todo }) => {
     >
       <h2 className="TodoInfo__title">{title}</h2>
 
-      <UserInfo user={findIdUser(todo)} />
+      {user && <UserInfo user={user} />}
     </article>
   );
 };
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,61 +1,119 @@
...
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import React, { useState } from 'react';
+import { TodoList } from './components/TodoList';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { Todos, Users } from './components/types';
 
 export const App = () => {
+  const [count, setCount] = useState<number>(2);
```
DarkMistyRoom, line 9: No need to store this in State. You can fined the max value of all todos' ids and increment it

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,25 @@
...
+  return (
+    <article
+      data-id={id}
+      className={cn('TodoInfo', { 'TodoInfo--completed': completed })}
+    >
+      <h2 className="TodoInfo__title">{title}</h2>
+
+      <a className="UserInfo" href={`mailto:${findIdUser(todo)?.email}`}>
+        {findIdUser(todo)?.name}
+      </a>
```
DarkMistyRoom, line 22: It should be in the separate component

DarkMistyRoom (changes requested):
Good job! But some changes are required

src/components/UserInfo/UserInfo.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { Users } from '../types';
+
+interface Props {
+  user?: Users;
```
anastasiia-tilikina, line 5: make this prop required. Your userInfo always should get user. And you will just render this component conditionally if user exists

anastasiia-tilikina (changes requested):
Just one more small comment

vadiimvooo (approved):
Well done!)

PR #2481, opened 2024-04-05
solution

Maksym-Atanasov:
[DEMO LINK](https://Maksym-Atanasov.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -44,8 +44,10 @@ export const App = () => {
     const creator =
       usersFromServer.find(user => user.id === selectedUser) || defaultCreator;
 
+    const idsOfTodos = todosFromServer.map(todo => todo.id);
+
     todos.push({
-      id: todos.length,
+      id: Math.max(...idsOfTodos) + 1,
       title,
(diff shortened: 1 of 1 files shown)


src/utils/defaultCreator.ts
```diff
@@ -0,0 +1,8 @@
+import { Creator } from '../types/types';
+
+export const defaultCreator: Creator = {
+  id: 0,
+  name: 'NonName',
+  username: 'noUsername',
+  email: 'noEmail@gmail.com',
+};
```
volodymyr-soltys97, line 8: Redundant

volodymyr-soltys97 (changes requested):
GJ!
Let's improve your code
You need to fix the failed test, if you need a help feel free in the fe_chat
You need to fix design as on the working example
image

Demo
image

src/utils/defaultCreator.ts
Maksym-Atanasov: I made it only for ts, because he think that .find() can return undefined

PR #2483, opened 2024-04-05
cope

true-binar-White-Rabbit:
[DEMO LINK](https://true-binar-White-Rabbit.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -2,24 +2,18 @@ import { UserInfo } from '../UserInfo';
 import { Todo, User } from '../../types/types';
 
 type Props = {
-  todos: Todo[];
+  todo: Todo;
   users: User[];
 };
 
-export const TodoInfo = ({ todos, users }: Props) => {
+export const TodoInfo = ({ todo, users }: Props) => {
+  const person = users.find(user => user.id === todo.userId);
+
   return (
-    <>
-      {todos.map(todo => {
-        const USER = users.find(user => user.id === todo.userId);
+    <article key={todo.id} className="TodoInfo TodoInfo--completed">
+      <h2 className="TodoInfo__title">{todo.title}</h2>
 
-        return (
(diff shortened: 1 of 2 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You need to fix the failed tests
image

The title cannot be empty
image

true-binar-White-Rabbit:
in your working expample i can add a todo with empty title

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,25 @@
...
+type Props = {
+  todos: Todo[];
+  users: User[];
+};
+
+export const TodoInfo = ({ todos, users }: Props) => {
+  return (
+    <>
+      {todos.map(todo => {
+        const USER = users.find(user => user.id === todo.userId);
```
etojeDenys, line 13:
```suggestion
        const user = users.find(user => user.id === todo.userId);
```
it's not a constant, so better to name it in camelCase

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,15 @@
...
+
+type Props = {
+  todos: Todo[];
+  users: User[];
+};
+
+export const TodoList = ({ todos, users }: Props) => {
+  return (
+    <section className="TodoList">
+      <TodoInfo todos={todos} users={users} />
```
etojeDenys, line 12: looks strange, you shoud map the todos array here and save in TodoInfo only todo information
```
{todos.map(todo => (
   <TodoInfo todo={todo} users={users}/>
))}
```

etojeDenys (changes requested):

src/components/TodoList/TodoList.tsx
true-binar-White-Rabbit: ok, thx)

vadiimvooo (approved):
LGTM

PR #2484, opened 2024-04-05
Implement adding todo's functionality

olafchuszno:
[DEMO LINK](https://olafchuszno.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,133 @@
+import { FormEventHandler, useState } from 'react';
 import './App.scss';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+export type Todo = {
+  id: number;
+  title: string;
+  completed: boolean;
+  userId: number;
+  user: User;
+};
+
+export type User = {
+  id: number;
+  name: string;
+  username: string;
+  email: string;
+};
 
 export const App = () => {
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,133 @@
...
+  name: string;
+  username: string;
+  email: string;
+};
 
 export const App = () => {
+  const todosWithUsers: Todo[] = todosFromServer.map(todo => {
+    return {
+      ...todo,
+      user: usersFromServer[todo.userId - 1],
```
BogdanMaliuta, line 27: bad idea to find user by index, what if simeine will delete 1 user from api?

src/App.tsx
```diff
@@ -1,27 +1,133 @@
...
+  const handleSubmit: FormEventHandler = event => {
+    event.preventDefault();
+
+    let highestTodoId = 0;
+
+    todos.forEach((todo: Todo) => {
+      if (todo.id > highestTodoId) {
+        highestTodoId = todo.id;
+      }
+    });
```
BogdanMaliuta, line 46: make separate function to get max id, also for each not better way

src/App.tsx
```diff
@@ -1,27 +1,133 @@
...
+
+    // Get the new todo
```
BogdanMaliuta, line 48: delete comments

BogdanMaliuta (changes requested):
Nice, but ome things needs to be fixes

BogdanMaliuta (approved):
Looks better, well done

PR #2492, opened 2024-04-08
Solution

LabPetro:
[DEMO LINK](https://LabPetro.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,106 @@
 import './App.scss';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { UserType } from './Types/UserType';
+import { ToDoType } from './Types/ToDoType';
+import { useState } from 'react';
+
+const getUser = (userId: number): UserType | null => {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,114 @@
...
-          <span className="error">Please enter a title</span>
+          <input
+            type="text"
+            data-cy="titleInput"
+            value={title}
+            placeholder="Enter your title"
+            onChange={e => {
+              setTitle(e.target.value);
+              setNoTitle('');
+            }}
```
lerastarynets, line 77: Pls move the function inside `onChange` handler to a separate const and use as a callback here

src/App.tsx
```diff
@@ -1,61 +1,114 @@
...
 
         <div className="field">
-          <select data-cy="userSelect">
+          <select
+            data-cy="userSelect"
+            value={userId}
+            onChange={e => {
+              setUserId(e.target.value);
+              setNoUser('');
+            }}
```
lerastarynets, line 89: Pls move the function inside `onChange` handler to a separate const and use as a callback here

lerastarynets (changes requested):
addTodoForm

lerastarynets (approved):
lgtm 🚀

PR #2496, opened 2024-04-09
Task resolved

bartKarb:
[DEMO LINK](https://bartKarb.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,92 @@
 import './App.scss';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import React, { useState } from 'react';
+import { TodoList } from './components/TodoList';
+import { Todo } from './components/types';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+const updatedTodos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: usersFromServer.find(user => user.id === todo.userId),
+}));
 
 export const App = () => {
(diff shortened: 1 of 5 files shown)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,15 @@
...
+
+type Props = {
+  user: User | undefined;
+};
+export const UserInfo: FC<Props> = ({ user }) => {
+  return (
+    user && (
+      <a className="UserInfo" href={`mailto:${user.email}`}>
+        {user.name}
+      </a>
```
choeqq, line 12: this check should be in UserList, when user isn't defined we shouldn't even render UserInfo

choeqq (changes requested):

src/App.tsx
```diff
@@ -1,61 +1,92 @@
...
+    const maxId = Math.max(...todos.map(todo => todo.id));
+    const newTodo = {
+      id: maxId + 1,
+      title: title,
+      completed: false,
+      userId: selectedUser,
+      user: usersFromServer.find(u => u.id === selectedUser),
+    };
+
+    setTodos([...todos, newTodo]);
```
nataliaklonowska, line 44: always when using prev state:
```suggestion
    setTodos(prevTodos => [...prevTodos, newTodo]);
```

nataliaklonowska (changes requested):

nataliaklonowska (approved):

PR #2519, opened 2024-04-24
solution

ArtemZakhar:
Hi)

Have a nice day)

[DEMO LINK](https://ArtemZakhar.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,11 +1,13 @@
-import './App.scss';
-
-import todosFromServer from './api/todos';
 import { useState } from 'react';
 import { TodoList } from './components/TodoList';
+import { TodoForm } from './components/TodoForm/TodoForm';
 import { TodoWithUser } from './types/todoWithUser';
 import { findUser } from './services/findUser';
-import TodoForm, { InputValues } from './components/TodoForm/TodoForm';
+
+import todosFromServer from './api/todos';
+
+import { InputValues } from './types/inputValues';
+import './App.scss';
 
 const todosWithUsers = todosFromServer.map(todo => ({
   ...todo,
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,42 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import todosFromServer from './api/todos';
+import { useState } from 'react';
+import { TodoList } from './components/TodoList';
+import { TodoWithUser } from './types/todoWithUser';
+import { findUser } from './services/findUser';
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,61 +1,44 @@
...
-// import todosFromServer from './api/todos';
+const todosWithUsers = todosFromServer.map(todo => ({
+  ...todo,
+  user: findUser(todo.userId),
+}));
 
 export const App = () => {
+  const [todos, setTodos] = useState(todosWithUsers);
+
+  const handleNewUser = (inputValues: InputValues) => {
```
polosanya, line 20: Actually, you are adding a new todo, not a user.
The name is a bit confusing

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { InputValues } from '../../types/inputValues';
+
+type Props = {
+  onAddTodo: ({ title, user }: InputValues) => void;
+};
+
+export const TodoForm: React.FC<Props> = ({ onAddTodo }) => {
+  const [user, setUser] = useState(0);
+  const [title, setTitle] = useState('');
+  const [titileError, setTitleError] = useState(false);
```
polosanya, line 12:
```suggestion
  const [titleError, setTitleError] = useState(false);
```

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,101 @@
...
+    if (!user || !title) {
+      return;
+    }
+
+    onAddTodo({ title, user });
+
+    setUserError(false);
+    setTitleError(false);
+    setTitle('');
+    setUser(0);
```
polosanya, line 45: Let's create a separate function for resetting

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,101 @@
...
+  };
+
+  return (
+    <form
+      action="/api/todos"
+      onSubmit={event => handleSubmit(event)}
```
polosanya, line 51:
```suggestion
      onSubmit={handleSubmit}
```
Can be simplified

src/components/TodoForm/TodoForm.tsx
ArtemZakhar: Sorry, the first time only part of the comments were downloaded( 
I am changing this now. 
But I was sure that it is ok to use it like this if we use it only once)

src/components/TodoForm/TodoForm.tsx
ArtemZakhar: Forgot about this opportunity( 
Many thanks!)

PR #2528, opened 2024-04-28
add_solution

metrakonon:
Have a nice day!
[DEMO](https://metrakonon.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,104 @@
 import './App.scss';
+import React, { useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { Todo } from './Types/TodoType';
+
+import { TodoList } from './components/TodoList';
+
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
 export const App = () => {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,104 @@
...
+    const isUserValid = userId !== 0;
+
+    setTitleError(!isTitleValid);
+    setUserError(!isUserValid);
+
+    if (!isTitleValid || !isUserValid) {
+      return;
+    }
+
+    const maxId = Math.max(...todos.map(todo => todo.id));
```
DanilWeda, line 31: You can create func with the name generateId or something like that(as helper), then use it in places were you need it.

src/App.tsx
```diff
@@ -1,61 +1,104 @@
...
+
+    const isTitleValid = title.trim() !== '';
+    const isUserValid = userId !== 0;
+
+    setTitleError(!isTitleValid);
+    setUserError(!isUserValid);
+
+    if (!isTitleValid || !isUserValid) {
+      return;
+    }
```
DanilWeda, line 29: You can move all validation in one func, then use it in this place. ( not critical)

DanilWeda (approved):
All good. I approved with small comments. 
Thanks)

PR #2535, opened 2024-04-30
Develop

yanamomot:
[DEMO LINK](https://yanamomot.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -13,10 +13,21 @@ const todos: Todo[] = todosFromServer.map(todo => {
   return { ...todo, user: getUserByTodo(todo.userId) };
 });
 
+function getNewTodoId(newTodos: Todo[]) {
+  const maxId = Math.max(...newTodos.map(todo => todo.id));
+
+  return maxId + 1;
+}
+
 export const App = () => {
   const [newTodos, setNewTodos] = useState<Todo[]>(todos);
 
-  const addNewPost = (createdTodo: Todo) => {
+  const addNewPost = (todo: Todo) => {
+    const createdTodo = {
+      ...todo,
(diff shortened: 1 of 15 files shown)


src/App.tsx
```diff
@@ -1,61 +1,41 @@
...
 export const App = () => {
+  const [newTodos, setNewTodos] = useState<Todo[]>(todos);
+
+  const addNewPost = (todo: Todo) => {
+    const createdTodo = {
+      ...todo,
+      id: getNewTodoId(newTodos),
+    };
+
+    setNewTodos(currTodo => [...currTodo, createdTodo]);
```
alexander-ignatow, line 31: 👍 
It is a good idea to use callback here

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,99 @@
...
+type Props = {
+  onSubmit: (post: Todo) => void;
+};
+
+export const Form: React.FC<Props> = ({ onSubmit }) => {
+  const [title, setTitle] = useState<string>('');
+  const [hasTitleError, setHasTitleError] = useState<boolean>(false);
+
+  const [userId, setUserId] = useState<number>(0);
+  const [hasSelectError, setHasSelectError] = useState<boolean>(false);
```
alexander-ignatow, line 16: In most case s typescript can guess the type, so it is ok to use following:

```suggestion
  const [title, setTitle] = useState('');
  const [hasTitleError, setHasTitleError] = useState(false);

  const [userId, setUserId] = useState(0);
  const [hasSelectError, setHasSelectError] = useState(false);
```

alexander-ignatow (approved):
Good job!

PR #2537, opened 2024-05-01
done

Stepan-R:
[DEMO LINK](https://Stepan-R.github.io/react_add-todo-form/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You need to add these text as on the working example
image

image

volodymyr-soltys97 (changes requested):
Still not fixed from the previous review, the title cannot be only spaces
image

volodymyr-soltys97 (approved):
Great work 🔥

PR #2546, opened 2024-05-03
todos form solution

sevenlays:
[DEMO LINK](https://sevenlays.github.io/react_add-todo-form/)


undeletable (changes requested):
The solution looks fine, except a few places where it can be improved. Please check that.

PR #2547, opened 2024-05-03
Tests fail

luyx0:
It seems that everything works, but the test does not pass, advise what can be fixed)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,115 @@
-import './App.scss';
+import React, { useState } from "react";
+import "./App.scss";
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from "./api/users";
+import todosFromServer from "./api/todos";
+
+import Users from './types/Users';
+import TodosWithUsers from "./types/TodosWithUsers";
+
+import { TodoList } from "./components/TodoList";
 
 export const App = () => {
+  const [users] = useState<Users[]>(usersFromServer);
+  const [todos, setTodos] = useState<TodosWithUsers[]>(todosFromServer.map(todo => ({
+    ...todo,
+    user: users.find(user => user.id === todo.userId) || null
+  })))
+  const [title, setTitle] = useState<string>("");
+  const [userId, setUserId] = useState<number | undefined>(undefined);
+  const [titleError, setTitleError] = useState<boolean>(false);
+  const [userError, setUserError] = useState<boolean>(false);
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (!title || !userId) {
+      return;
+    }
+
+    const selectedUser = users.find((user) => user.id === userId);
+    if (!selectedUser) {
+      return;
+    }
+
+    const newTodo: TodosWithUsers = {
+      id: todos.length + 1,
+      title: title,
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,115 @@
...
+      return;
+    }
+
+    const selectedUser = users.find((user) => user.id === userId);
+    if (!selectedUser) {
+      return;
+    }
+
+    const newTodo: TodosWithUsers = {
+      id: todos.length + 1,
```
ArtemZakhar, line 36: 1) should add a title with the max existing ID + 1

This error means that you need to write a function that will loop through your to-do list and choose from there max id number. And you have to put id as maxId + 1;

src/App.tsx
```diff
@@ -1,61 +1,115 @@
...
+    user: users.find(user => user.id === todo.userId) || null
+  })))
+  const [title, setTitle] = useState<string>("");
+  const [userId, setUserId] = useState<number | undefined>(undefined);
+  const [titleError, setTitleError] = useState<boolean>(false);
+  const [userError, setUserError] = useState<boolean>(false);
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
```
ArtemZakhar, line 25: 2) should show a title error if the title is empty

If I remember correctly, Mihailo showed an approach when you use input value as a boolean property.
So before you stop submission you have to write smth like this setTitleError(!title.trim().)

src/App.tsx
```diff
@@ -1,61 +1,115 @@
...
+    if (!title || !userId) {
```
ArtemZakhar, line 26: 3) should show a user error if a user is not selected

the same as the previous comment but setUserError(!userId)

src/App.tsx
```diff
@@ -1,61 +1,115 @@
...
+      return;
```
ArtemZakhar, line 27: 4) should show all errors if form is empty

this one you will pass if implement an error handler.

src/App.tsx
```diff
@@ -1,61 +1,115 @@
...
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <label htmlFor="title" className="label">Title: </label>
+          <input
+            id="post-title"
+            type="text"
+            value={title}
+            data-cy="titleInput"
+            onChange={(event) => setTitle(event.target.value)}
```
ArtemZakhar, line 74: 5) should keep a user error on title change

you need a separate function where you take off an error from the title if the title state changes.

src/App.tsx
```diff
@@ -1,61 +1,115 @@
...
         <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>
+          <label htmlFor="title">User: </label>
+          <select
+            id="post-section"
+            value={userId || ""}
+            data-cy="userSelect"
+            onChange={(event) => setUserId(Number(event.target.value))}
+          >
```
ArtemZakhar, line 87: 6) should keep a title error on user change

you need a separate function where you take off an error from the user if another one is chosen and userId state changes.

ArtemZakhar (changes requested):
Great job) 
You have almost done it) 
Small changes and all tests will be passed!)

ArtemZakhar (approved):
Great job!)

PR #2548, opened 2024-05-03
add task solution

Esther-Guimaraes:
(https://Esther-Guimaraes.github.io/react_add-todo-form/)


pedro-ruas (approved):
Excellent code! I really liked especially the way you handled _TodoComplete.setId_, great use of objects and understanding of closures.

PR #2550, opened 2024-05-04
add task solution

lmuias:
[DEMO LINK](https://lmuias.github.io/react_add-todo-form/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better
image

volodymyr-soltys97 (approved):
Good job!

PR #2551, opened 2024-05-04
solution

denyssheremeta:
[DEMO LINK](https://denyssheremeta.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,119 @@
+import { useState } from 'react';
+
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-export const App = () => {
+import { TodoList } from './components/TodoList';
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -30,31 +123,7 @@ export const App = () => {
       </form>
 
       <section className="TodoList">
```
iAlfreX, line 125: This section tag should be placed in the TodoList file, so you will get rid of empty tags in that file

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,21 @@
...
+import { TodoInfo } from '../TodoInfo/TodoInfo';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <>
+      {todos.length > 0 && (
```
iAlfreX, line 12: Unnecessary check

iAlfreX (approved):
Good job!

PR #2554, opened 2024-05-05
Solution

SlavaRevit:
- [DEMO LINK](https://slavarevit.github.io/react_add-todo-form/)


etojeDenys (changes requested):
I can create the todo only with spaces, let's fix that
image

lerastarynets (approved):
You're making excellent progress!

PR #2559, opened 2024-05-07
add solution

stepanyukvova200:
[DEMO LINK](https://stepanyukvova200.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,87 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import React, { useState } from 'react';
+import { Todo } from './types/todo';
+import { TodoList } from './components/TodoList';
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [hasTitleError, setHasTitleError] = useState(false);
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,61 +1,95 @@
...
+    setTitle('');
+    setUser(0);
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/todos" method="POST">
+      <form action="/api/todos" method="POST" onSubmit={handleSubmit}>
```
ordila, line 44: According to React standartds it's better to divide all into components. So here will be better to create component for Form and use it. Because all form logic in App.tsx - its not good at all;

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,37 @@
...
+interface Props {
+  todo: Todo;
+}
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const users: User[] = usersFromServer;
+  const user: User = users.find(
+    currentUser => currentUser.id === todo.userId,
+  ) ?? {
+    id: 0,
```
ordila, line 19: Maybe it will be better to create some const for this and use here;

ordila (approved):
Nice work . 
Look at comments, there are small tips.
Thank you

PR #2560, opened 2024-05-07
add solving

ordila:


src/components/AddToDoForm/AddToDoForm.tsx
```diff
@@ -0,0 +1,87 @@
...
+            placeholder="Enter todo"
+            style={{ marginBottom: '10px' }}
+          />
+        </label>
+
+        {errors.title && <span className="error">Please enter a title</span>}
+      </div>
+
+      <div className="field">
+        <label>
```
dyakym, line 53: Use label instead of span and add : after User

src/components/AddToDoForm/AddToDoForm.tsx
```diff
@@ -0,0 +1,87 @@
...
+      id: findMaxUserId(usersFromServer) + 1,
+    });
+
+    reset();
+  };
+
+  return (
+    <form action="/api/todos" method="POST" onSubmit={handleSubmit(onSubmit)}>
+      <div className="field">
+        <label>
```
dyakym, line 35: Use label instead of span and rename Todo to Title:

dyakym (approved):
In general good job, please fix small issues

PR #2561, opened 2024-05-07
Solution

dyakym:
[DEMO LINK](https://dyakym.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,112 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+ import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { Todo, TodoWithUser } from './types/Todo';
+import { User } from './types/User';
+import { useState } from 'react';
+import { TodoList } from './components/TodoList';
+
+function getUserById(userId: number): User | null {
+  return usersFromServer.find(user => user.id === userId)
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,120 @@
...
+
+function getUserById(userId: number): User | null {
+  return usersFromServer.find(user => user.id === userId)
+    || null;
+};
+
+const todosWithUsers: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
```
ilyasolo881, line 18: Can cocncat this in 1 function and do Function Expression
const todosWithUsers: TodoWithUser[] = todosFromServer.map(todo => {
  const user =
    usersFromServer.find(person => person.id === todo.userId) || null;

  return {
    ...todo,
    user,
  };
});

src/App.tsx
```diff
@@ -1,61 +1,120 @@
...
+    setHasTitleError(!title);
+
+    if (!title || !userId) {
+      return;
+    }
+
+    const newTodo: TodoWithUser = {
+      id: findMaxId(todos) + 1,
+      title,
+      userId: +userId,
```
ilyasolo881, line 64: No need to use +userId, becouse u have number in state:

ilyasolo881 (approved):
Nice work! Small improvments in comments

PR #2569, opened 2024-05-07
add solution

andriy-kostiuk:
[DEMO LINK](https://andriy-kostiuk.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,1 +1,22 @@
-export const TodoInfo = () => {};
+import { TodoWithUser } from '../../interfaces/Todo';
+import cn from 'classnames';
+import { UserInfo } from '../UserInfo';
+
+interface Props {
+  todo: TodoWithUser;
+}
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const { id, title, completed, user } = todo;
+
+  return (
+    <article
+      data-id={id}
+      className={cn('TodoInfo', { 'TodoInfo--completed': completed })}
+    >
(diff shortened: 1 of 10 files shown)


src/services/user.ts
```diff
@@ -0,0 +1,5 @@
+import { User } from '../interfaces/User';
+
+export const getUserById = (users: User[], id: number): User => {
+  return users.find(user => user.id === id) as User;
```
etojeDenys, line 4:
```suggestion
  return users.find(user => user.id === id) || null;
```
try to avoid assertions as much as you can

src/interfaces/Todo.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export interface Todo {
+  id: number;
+  title: string;
+  completed: boolean;
+  userId: number;
+}
+
+export interface TodoWithUser extends Todo {
+  user: User;
```
etojeDenys, line 11:
```suggestion
  user: User | null;
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,22 @@
...
+  const { id, title, completed, user } = todo;
+
+  return (
+    <article
+      data-id={id}
+      className={cn('TodoInfo', { 'TodoInfo--completed': completed })}
+    >
+      <h2 className="TodoInfo__title">{title}</h2>
+
+      <UserInfo user={user} />
```
etojeDenys, line 19: don't forget to add a check here, if the user is null

etojeDenys (changes requested):
try to fix the demo link or feel free to ask for some help in the chat

PR #2570, opened 2024-05-08
solution

tarassamchenko:
[DEMO LINK](https://tarassamchenko.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,125 @@
 import './App.scss';
+import { useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoWithUser } from './types/Todo';
+
+const todosWithUsers: TodoWithUser[] = todosFromServer.map(todo => {
+  const user =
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+  const ids = todos.map(todo => todo.id);
+
+  return Math.max(...ids);
+};
 
 export const App = () => {
+  const [todos, setTodos] = useState<TodoWithUser[]>(todosWithUsers);
+
+  const [title, setTitle] = useState('');
+  const [titleError, setTitleError] = useState(false);
```
stepanyukvova200, line 30: better to name hasTitleError or something like that

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+
+    handleResetForm();
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/todos" method="POST">
+      <form action="/api/todos" method="POST" onSubmit={handleAddTodo}>
```
stepanyukvova200, line 79: It's better to move all parameters onto separate lines, because there are more than two of them, and it's hard to read them.

stepanyukvova200 (approved):
Good job!
Everything works in the right way,  just a few remarks from me )

PR #2577, opened 2024-05-12
adding my form

vinasxz:
[DEMO LINK](https://vinasxz.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,102 @@
+import { FormEvent, useState } from 'react';
 import './App.scss';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+const todosWithUser: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: usersFromServer.find(user => user.id === todo.userId),
+}));
 
 export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>(todosWithUser);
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+const todosWithUser: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: usersFromServer.find(user => user.id === todo.userId),
+}));
 
 export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>(todosWithUser);
```
EdPirro, line 15: No need to explicitly define types which can be implied from value

src/App.tsx
```diff
@@ -1,61 +1,110 @@
...
+  const [name, setName] = useState('');
+  const [user, setUser] = useState<number>(0);
```
EdPirro, line 17: No need to explicitly define types which can be implied from value

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,23 @@
-export const TodoInfo = () => {};
+import { FC } from 'react';
+import { Todo } from '../../types';
+import { UserInfo } from '../UserInfo';
+import classNames from 'classnames';
+
+type Props = {
+  todo: Todo;
+};
```
EdPirro, line 8: As best a practice include the component name in the Prop type name `TodoInfoProps`; this helps avoid confusion and redefinition of types in larger projects

EdPirro (approved):
Great job, @vinasxz! Keep it up! 🚀

PR #2580, opened 2024-05-13
Solution

przwojwwp:
[DEMO LINK](https://przwojwwp.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -11,6 +11,10 @@ button {
   cursor: pointer;
 }
 
+.userSelect {
+  width: 165.6px;
+}
+
 .error {
   color: red;
 }
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -28,27 +28,25 @@ export const App = () => {
   const handleSubmit = (event: FormEvent<HTMLFormElement>) => {
     event.preventDefault();
 
-    if (!titleTodo && selectedUser === '0') {
-      setErrorTitleInput('Please enter a title');
-      setErrorUserSelect('Please choose a user');
+    let titleError = '';
+    let userError = '';
 
-      return;
-    } else if (!titleTodo) {
-      setErrorTitleInput('Please enter a title');
-      setErrorUserSelect('');
-
-      return;
-    } else if (selectedUser === '0') {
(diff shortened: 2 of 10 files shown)


BogdanMaliuta (changes requested):
Hello, some tests wasn`t passed, fix it pls.
image

src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+import { FormEvent, useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  // if there is no user with a given userId
```
BogdanMaliuta, line 12: remove comments

src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+      return;
+    } else if (selectedUser === '0') {
+      setErrorTitleInput('');
+      setErrorUserSelect('Please choose a user');
+
+      return;
+    } else {
+      setErrorTitleInput('');
+      setErrorUserSelect('');
+    }
```
BogdanMaliuta, line 49: if you are using returns in if/else, you can just use if without else

BogdanMaliuta: also you dont need to describe this all errors, you just need to check if title and user are not empty

BogdanMaliuta (changes requested):
Nice work, but fix handle submit pls

src/App.scss
```diff
@@ -11,6 +11,10 @@ button {
   cursor: pointer;
 }
 
+.userSelect {
+  width: 165.6px;
+}
```
BogdanMaliuta, line 16: not good practice to set pixels using dot

src/App.tsx
BogdanMaliuta: why didnt remove?

BogdanMaliuta (approved):
Looks better, 
pls fix what I described in comments

PR #2581, opened 2024-05-14
todos solution

ZupaFly:
[DEMO LINK](https://zupafly.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,110 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { ChangeEventHandler, FormEventHandler, useState } from 'react';
+import { TodoList } from './components/TodoList';
+import { User } from './types/User';
+import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+
+const todosWithUsers = todosFromServer.map(todo => {
+  return {
+    ...todo,
+    user: usersFromServer.find(user => user.id === todo.userId) as User,
+  };
+});
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserDropdown] = useState<number>(0);
+  const [failedTitle, setTitleFailed] = useState(false);
+  const [failedUser, setUserFailed] = useState(false);
+  const [todos, setTodos] = useState(todosWithUsers);
+
+  const resetForm = () => {
+    setUserDropdown(0);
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,118 @@
...
+import { ChangeEventHandler, FormEventHandler, useState } from 'react';
+import { TodoList } from './components/TodoList';
+import { User } from './types/User';
+import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+
+const todosWithUsers = todosFromServer.map(todo => {
+  return {
+    ...todo,
+    user: usersFromServer.find(user => user.id === todo.userId) as User,
```
anastasiia-tilikina, line 12: this is not a good idea using as in this case
it is better to handle undefined with || operator

now ts won't warn you about potential undefined value

src/App.tsx
```diff
@@ -1,61 +1,118 @@
...
+  };
+});
 
 export const App = () => {
+  const [title, setTitle] = useState('');
+  const [userId, setUserDropdown] = useState<number>(0);
```
anastasiia-tilikina, line 18: it is better to name state consistently
like `[title, setTitle]` or `[userId, setUserId]`

src/App.tsx
```diff
@@ -1,61 +1,118 @@
...
+      id:
+        Math.max(
+          ...todos.map(t => {
+            return t.id;
+          }),
+        ) + 1,
+      user: usersFromServer.find(user => user.id === Number(userId)) as User,
+    };
+
+    setTodos([...todos, newTodo]);
```
anastasiia-tilikina, line 65: use callback for setting state based on the previous one

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,20 @@
...
+type TodoInfoProps = {
+  todo: Todo & { user: User };
+};
+
+export const TodoInfo: React.FC<TodoInfoProps> = ({
+  todo: { title, completed, user, id },
+}) => (
+  <article
+    data-id={id}
+    className={`TodoInfo ${completed ? 'TodoInfo--completed' : ''}`}
```
anastasiia-tilikina, line 15: pls use `classnames` library for dynamic classes

anastasiia-tilikina (changes requested):
GJ

pls check my comments

PR #2582, opened 2024-05-14
Solution

ropanefarias:
[DEMO LINK](https://ropanefarias.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,111 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import todos from './api/todos';
+import users from './api/users';
+import { TodoList } from './components/TodoList';
+
+interface Todo {
+  id: number;
+  title: string;
+  completed: boolean;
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,61 +1,119 @@
...
+  username: string;
+  email: string;
+}
 
 export const App = () => {
+  const [addedTodos, setAddedTodos] = useState<Todo[]>(todos);
+  const [newTodoTitle, setNewTodoTitle] = useState<string>('');
+  const [selectedUserId, setSelectedUserId] = useState<number>(0);
+  const [titleError, setTitleError] = useState<string>('');
+  const [userError, setUserError] = useState<string>('');
```
EdPirro, line 27: No need to explicitly define types where it can be implied.
Instead of `const [newTodoTitle, setNewTodoTitle] = useState<string>('');` simply use `const [newTodoTitle, setNewTodoTitle] = useState('');`; type `string` in implicit

EdPirro (approved):
Great job, @ropanefarias! Keep it up! 🚀

PR #2589, opened 2024-05-21
add task solution

wasik07:
[DEMO LINK](https://wasik07.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,26 @@
 import './App.scss';
+import { useCallback, useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { TodoForm } from './components/TodoForm';
+import { filteredTodos, getMaxId } from './services/todos';
+import { Todo } from './types/Todo';
 
-export const App = () => {
+export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>(filteredTodos);
+  const maxId: number = getMaxId(todos) + 1;
+
+  const getPreparedTodos = useCallback((todo: Todo) => {
+    setTodos(currentTodos => currentTodos.concat(todo));
+  }, []);
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/todos" method="POST">
-        <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
-        </div>
+      <TodoForm onAdd={getPreparedTodos} newTodoId={maxId} />
 
-        <div className="field">
-          <select data-cy="userSelect">
-            <option value="0" disabled>
-              Choose a user
-            </option>
-          </select>
-
-          <span className="error">Please choose a user</span>
-        </div>
-
-        <button type="submit" data-cy="submitButton">
-          Add
-        </button>
-      </form>
-
-      <section className="TodoList">
-        <article data-id="1" className="TodoInfo TodoInfo--completed">
-          <h2 className="TodoInfo__title">delectus aut autem</h2>
-
-          <a className="UserInfo" href="mailto:Sincere@april.biz">
-            Leanne Graham
-          </a>
-        </article>
-
-        <article data-id="15" className="TodoInfo TodoInfo--completed">
-          <h2 className="TodoInfo__title">delectus aut autem</h2>
-
-          <a className="UserInfo" href="mailto:Sincere@april.biz">
-            Leanne Graham
-          </a>
-        </article>
-
-        <article data-id="2" className="TodoInfo">
-          <h2 className="TodoInfo__title">
-            quis ut nam facilis et officia qui
-          </h2>
-
-          <a className="UserInfo" href="mailto:Julianne.OConner@kory.org">
-            Patricia Lebsack
-          </a>
-        </article>
-      </section>
+      <TodoList OnTodos={todos} />
     </div>
   );
 };
--- a/src/components/TodoForm/TodoForm.tsx
+++ b/src/components/TodoForm/TodoForm.tsx
@@ -1,0 +1,107 @@
+import { useState } from 'react';
+
+import { getAllUsers, getUserById } from '../../services/user';
+import { isValidLetter } from '../../services/Pattern';
+import { Todo } from '../../types/Todo';
+
+interface Props {
+  onAdd: (todo: Todo) => void;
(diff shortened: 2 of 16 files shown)


src/App.tsx
```diff
@@ -1,61 +1,26 @@
...
+import { TodoForm } from './components/TodoForm';
+import { filteredTodos, getMaxId } from './services/todos';
+import { Todo } from './types/Todo';
+
+export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>(filteredTodos);
+  const maxId: number = getMaxId(todos) + 1;
+
+  const getPreparedTodos = useCallback((todo: Todo) => {
+    setTodos(currentTodos => currentTodos.concat(todo));
```
anastasiia-tilikina, line 14:
```suggestion
    setTodos(currentTodos => [...currentTodos, todo]);
```
this looks a bit cleaner

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,107 @@
+import { useState } from 'react';
+
+import { getAllUsers, getUserById } from '../../services/user';
+import { isValidLetter } from '../../services/Pattern';
+import { Todo } from '../../types/Todo';
+
+interface Props {
+  onAdd: (todo: Todo) => void;
+  newTodoId: number;
```
anastasiia-tilikina, line 9: you don't need this as a prop. you can handle this logic in App component

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,107 @@
...
+  };
+
+  const reset = () => {
+    setTitle('');
+    setUserId(0);
+
+    setHasTitleError(false);
+    setHasUserIdError(false);
+
+    setCount(count + 1);
```
anastasiia-tilikina, line 45: why do you need this?

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const [hasUserIdError, setHasUserIdError] = useState(false);
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const TitleTetters = event.target.value;
+
+    if (isValidLetter(TitleTetters.slice(-1))) {
+      setTitle(TitleTetters);
+    } else {
+      setTitle(TitleTetters.slice(0, -1));
+    }
```
anastasiia-tilikina, line 28: this logic in not really clear. why do you need slice here?

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+
+  const [hasTitleError, setHasTitleError] = useState(false);
+  const [hasUserIdError, setHasUserIdError] = useState(false);
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const TitleTetters = event.target.value;
```
anastasiia-tilikina, line 22:
```suggestion
    const titleTetters = event.target.value;
```

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,107 @@
...
+    setHasTitleError(false);
+    setHasUserIdError(false);
+
+    setCount(count + 1);
+  };
+
+  const handleAdd = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    setHasTitleError(!title);
```
anastasiia-tilikina, line 51: you need to handle scenario when title contains spaces only. I shouldn't be able to add todo in this case

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,14 @@
-export const TodoList = () => {};
+import { TodoInfo } from '../TodoInfo';
+import { Todo } from '../../types/Todo';
+
+interface Props {
+  OnTodos: Todo[];
```
anastasiia-tilikina, line 5: using `todos` name for props is more understandable

src/services/todos.ts
```diff
@@ -0,0 +1,16 @@
+import todosFromServer from '../api/todos';
+import { Todo } from '../types/Todo';
+import { getUserById } from './user';
+
+export const filteredTodos = todosFromServer
+  .map(todo => ({
+    ...todo,
+    user: getUserById(todo.userId),
+  }))
+  .filter(todo => todo.user?.id);
```
anastasiia-tilikina, line 10: why do you need this filter?
you can just handle case when user is undefined

src/services/user.ts
```diff
@@ -0,0 +1,10 @@
+import { User } from '../types/User';
+import usersFromServer from '../api/users';
+
+export function getAllUsers(): User[] {
+  return usersFromServer;
+}
```
anastasiia-tilikina, line 6: this function does nothing. you can just import a variable where you need. Try to not overthink)

anastasiia-tilikina (changes requested):
Good job!

It is great how you are trying to structure and organize your code!

but let's make some fixes here

src/components/TodoForm/TodoForm.tsx
wasik07: I don`t need this)

src/services/todos.ts
wasik07: I deleted this filter.
In my opinion, it is needed in order to discard standard titles in which there are no users. I know that there are none, but in theory they could be)

src/components/TodoForm/TodoForm.tsx
wasik07: I did it to complete this task.
(* Optional) Allow entering only letters (ua and en), digits, and spaces in the title field. Just remove any other characters from the title.
Using isValidLetter(TitleTetters.slice(-1)), I check the last added character in the string, and if it matches the pattern is Valid,i use Titleletters. If it doesn't match, I shorten the string using setTitle(TitleTetters.slice(0, -1)), thereby removing the last character from the string.

src/components/TodoForm/TodoForm.tsx
etojeDenys: image
     
image
still not fixed

etojeDenys (changes requested):

volodymyr-soltys97 (approved):
Good job 👍

PR #2590, opened 2024-05-22
Solved

meljaszuk:
[DEMO LINK](https://meljaszuk.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,8 +3,10 @@ import './App.scss';
 
 import usersFromServer from './api/users';
 import todosFromServer from './api/todos';
+/* eslint-disable */
 import { Todo } from './types/Todo';
 import { User } from './types/User';
+/* eslint-enable */
 import { TodoList } from './components/TodoList';
 
 const getUserById = (userId: number): User | null =>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,117 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,61 +1,119 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+/* eslint-disable */
```
yevhenii-pyl, line 6: Why do you have to disable eslint here?

src/App.tsx
```diff
@@ -1,61 +1,119 @@
...
+  const [errorTitle, setErrorTitle] = useState('');
+  const [errorID, setErrorID] = useState('');
+
+  const addTodo = (newTodo: Todo) => {
+    setTodos(currentTodos => [...currentTodos, newTodo]);
+  };
+
+  const createNewId = () => Math.max(...todos.map(todo => todo.id)) + 1;
+
+  const resetFields = () => {
```
yevhenii-pyl, line 33: 👍

src/App.tsx
```diff
@@ -1,61 +1,119 @@
...
+  };
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (!todoTitle) {
+      setErrorTitle('Please enter a title');
+    }
+
+    if (!userId) {
```
yevhenii-pyl, line 45: Cleaner approach would be to create a helper `validate` function, not a necessity here, but still makes sense.

PR #2592, opened 2024-05-25
add task solution

ArsenSkandalenko:
[DEMO LINK](https://ArsenSkandalenko.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,122 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { useState } from 'react';
+import { User } from './types/User.type';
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,130 @@
...
+            data-cy="userSelect"
+            onChange={handleSelectedUserId}
+          >
+            <option className="default-option" value="0" disabled selected>
               Choose a user
             </option>
+
+            {usersFromServer.map(user => (
+              <option key={user.id} value={user.id}>
+                {user.name}
```
volodymyr-soltys97, line 112: Use destruction for `user`

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
If you add only spaces to the title, then when you click on the add button, nothing happens, you need to show an error
image

VitaliyBondarenko1982 (approved):
Looks good.

PR #2597, opened 2024-05-28
react add todo form

shelyaa:
[DEMO LINK](https://shelyaa.github.io/react_add-todo-form/)


src/App.tsx
```diff
@@ -1,61 +1,103 @@
...
+  const handleUserChange = (event: ChangeEvent<HTMLSelectElement>) => {
+    setUserId(+event.target.value);
+    setUserError('');
+  };
+
+  const handleSubmit = (event: ChangeEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    let isValid = true;
+
+    if (title.trim() === '') {
```
anastasiia-tilikina, line 30:
```suggestion
    if (!title.trim()) {
```

src/App.tsx
```diff
@@ -1,61 +1,103 @@
...
+      isValid = false;
+      setTitleError('Please enter a title');
+    }
+
+    if (userId === 0) {
```
anastasiia-tilikina, line 35:
```suggestion
    if (!userId) {
```

src/App.tsx
```diff
@@ -1,61 +1,103 @@
...
+      isValid = false;
+      setUserError('Please choose a user');
+    }
+
+    if (isValid) {
+      const user = usersFromServer.find(u => u.id === userId);
+      const newTodo: Todo = {
+        id: todos.length > 0 ? Math.max(...todos.map(todo => todo.id)) + 1 : 1,
```
anastasiia-tilikina, line 43: pls move getting id to a separate method

src/App.tsx
```diff
@@ -1,61 +1,103 @@
...
+        title,
+        userId,
+        user,
+        completed: false,
+      };
+
+      setTodos([...todos, newTodo]);
```
anastasiia-tilikina, line 50: always use callback for setting state based on the previous one

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,11 @@
...
+
+type Props = {
+  user: User | undefined;
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => (
+  <a className="UserInfo" href={user ? `mailto:${user.email}` : '#'}>
+    {user ? user.name : 'Unknown user'}
+  </a>
+);
```
anastasiia-tilikina, line 11: it is better to handle undefined case in parent component
User should always accept User type as a prop

in parent component you can use just conditional rendering

anastasiia-tilikina (changes requested):
Well done!

just a couple suggestions here

lerastarynets (approved):
You're on the right track!

PR #2601, opened 2024-05-31
add task solution

ogchy:
https://ogchy.github.io/react_add-todo-form/


anastasiiavorobiova (changes requested):
Great work! Consider using `trim()` to avoid adding a todo with spaces only 
Screenshot 2024-05-31 at 16 21 35

anastasiiavorobiova (approved):
Great work!

PR #2608, opened 2024-06-07
Develop

aMtiIV:
[DEMO LINK](https://aMtiIV.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,125 @@
+import React, { useState } from 'react';
 import './App.scss';
+import { User } from './types/User';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
 
-export const App = () => {
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+function getTodosFromServer(): Todo[] {
+  return todosFromServer.map(todo => ({
+    ...todo,
+    user: getUser(todo.userId),
+  }));
+}
+
+export const App: React.FC = () => {
+  const [todos, setTodos] = useState(getTodosFromServer());
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [titleError, setTitleError] = useState(false);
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+
+  function getNewTodoId(): number {
+    return (
+      1 +
+      todos.reduce((maxId, todo) => {
+        const { id } = todo;
+
+        return id > maxId ? id : maxId;
+      }, 0)
+    );
```
BogdanMaliuta, line 37: why you are using here reducer?
You just need to find maxId, you have better methods for this

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+    event.preventDefault();
+    let errorOccured = false;
+    const trimmedTitle = title.trim();
+
+    if (trimmedTitle === '') {
+      setTitleError(true);
+      errorOccured = true;
+    }
+
+    if (userId === 0) {
```
BogdanMaliuta, line 60:
```suggestion
    if (!userId) {
```

src/App.tsx
```diff
@@ -1,61 +1,133 @@
...
+    setUserId(+event.currentTarget.value);
+    setUserError(false);
+  }
+
+  function handleSubmit(event: React.FormEvent) {
+    event.preventDefault();
+    let errorOccured = false;
+    const trimmedTitle = title.trim();
+
+    if (trimmedTitle === '') {
```
BogdanMaliuta, line 55:
```suggestion
    if (!trimmedTitle) {
```

BogdanMaliuta (changes requested):
Works good, but some things in code need to be changed

src/App.tsx
aMtiIV: I'm using reducer because it's the fastest method I know. It has time complexity of O(n). Getting first element of sorted array has time complexity of O(nlog(n)). For Math.max with map method inside it is: O(2n). The other problem with Math.max is that the 'Maximum call stack size exceeded' exception can be thrown if the array is too big because all array elements are passed as parameters. What else could I use to not sacrifice efficiency? Or maybe in this particular case efficiency isn't that important so we can focus more on code style?

natalia-klonowska (approved):
Great work 🚀 you're right about the limitations of the Math.max method 👍 Firstly we focus on the readability of your code but since you're aware of the pros and cons of both approaches your solution can stay as it is :)

PR #2610, opened 2024-06-09
Solution

tania-kuzmenko:
https://tania-kuzmenko.github.io/react_add-todo-form/

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,1 +1,23 @@
-export const TodoInfo = () => {};
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+import classNames from 'classnames';
+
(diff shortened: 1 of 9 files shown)


anastasiiavorobiova (changes requested):
Make sure to deploy your solution

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,23 @@
...
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+import classNames from 'classnames';
+
+interface Props {
+  todo: Todo;
+}
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  return (
+    todo.user && (
```
DarkMistyRoom, line 11: would be better to use conditional rendering outside of the component

DarkMistyRoom (approved):
Good job! Only one comment

PR #2619, opened 2024-06-23
solution

hsvirina:
https://hsvirina.github.io/react_add-todo-form/


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
image

volodymyr-soltys97 (approved):
Good job 👍

PR #2630, opened 2024-06-26
Develop

GoVolodya:
[DEMO LINK](https://GoVolodya.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -65,9 +65,9 @@ export const App = () => {
       return;
     }
 
-    // eslint-disable-next-line
-    const nextId: number =
-      [...todos].sort((a: Todo, b: Todo) => a.id - b.id).pop().id + 1;
+    const ids: number[] = [...todos].map((todo: Todo) => todo.id);
+    const nextId: number = Math.max(...ids) + 1;
+
     const todo = {
       id: nextId,
       userId: selectedUser,
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,118 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import React, { useState } from 'react';
+import { TodoList } from './components/TodoList';
+
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { User } from './types/types';
+import { Todo } from './types/types';
+import { ServerTodo } from './types/types';
+
+function findUser(userId: number): User | undefined {
+  return usersFromServer.find((userFromServer: User): boolean => {
+    return userFromServer.id === userId;
+  });
+}
+
+const preparedTodos: Todo[] = todosFromServer.map(
+  (todoFromServer: ServerTodo): Todo => {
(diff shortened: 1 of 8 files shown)


src/types/types.ts
```diff
@@ -0,0 +1,21 @@
...
+  email: string;
+};
+
+export type Todo = {
+  id: number;
+  title: string;
+  completed: boolean;
+  userId: number;
+  user: User;
+};
```
etojeDenys, line 14:
```suggestion
export type Todo = ServerTodo & {
  user: User;
};
```

etojeDenys (changes requested):
image
we can create the todo only with spaces, let's fix that

src/types/types.ts
GoVolodya: Good advice didn't know that trick. It makes code better.

src/App.tsx
```diff
@@ -1,61 +1,126 @@
...
+import todosFromServer from './api/todos';
+
+import { User } from './types/types';
+import { Todo } from './types/types';
+import { ServerTodo } from './types/types';
+
+function findUser(userId: number): User {
+  return usersFromServer.find((userFromServer: User): boolean => {
+    return userFromServer.id === userId;
+  })!;
```
etojeDenys, line 16:
```suggestion
  return usersFromServer.find((userFromServer: User): boolean => {
    return userFromServer.id === userId;
  }) || null;
```

src/App.tsx
```diff
@@ -1,61 +1,126 @@
...
+import { TodoList } from './components/TodoList';
+
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+
+import { User } from './types/types';
+import { Todo } from './types/types';
+import { ServerTodo } from './types/types';
+
+function findUser(userId: number): User {
```
```suggestion
function findUser(userId: number): User | null {
```

src/types/types.ts
```diff
@@ -0,0 +1,17 @@
...
+export type ServerTodo = {
+  id: number;
+  title: string;
+  completed: boolean;
+  userId: number;
+};
+
+export type Todo = ServerTodo & {
+  user: User;
+};
```
etojeDenys, line 17:
```suggestion
export type Todo = ServerTodo & {
  user?: User;
};
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,21 @@
...
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => (
+  <article
+    data-id={todo.id}
+    className={`TodoInfo ${todo.completed ? 'TodoInfo--completed' : ''}`}
+  >
+    <h2 className="TodoInfo__title">{todo.title}</h2>
+
+    <UserInfo user={todo.user} />
```
etojeDenys, line 19: and add here some check if user exists
```suggestion
    {todo.user && <UserInfo user={todo.user} />}
```

etojeDenys (approved):

PR #2637, opened 2024-06-26
solution

ikocherovets:
[DEMO LINK](https://ikocherovets.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,97 @@
 import './App.scss';
-
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { useState } from 'react';
+import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import { TodoList } from './components/TodoList';
 
 export const App = () => {
+  const [todos, setTodos] = useState(todosFromServer);
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [titleError, setTitleError] = useState('');
+  const [userError, setUserError] = useState('');
+
+  const handleAddTodo = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    let hasError = false;
+
+    if (title.trim() === '') {
+      setTitleError('Please enter a title');
+      hasError = true;
+    } else {
+      setTitleError('');
+    }
+
+    if (userId === 0) {
+      setUserError('Please choose a user');
+      hasError = true;
+    } else {
+      setUserError('');
+    }
+
+    if (hasError) {
+      return;
+    }
+
+    const newTodo = {
+      id: todos.length ? Math.max(...todos.map(todo => todo.id)) + 1 : 1,
+      title: title.trim(),
+      userId,
+      completed: false,
+    };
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,97 @@
...
+import { useState } from 'react';
+import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import { TodoList } from './components/TodoList';
 
 export const App = () => {
+  const [todos, setTodos] = useState(todosFromServer);
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [titleError, setTitleError] = useState('');
```
lesdyman, line 11: I think it would be better if the titleError and userError states had a boolean value. After all, the point is that the error exists or not, right? In the case of your current solution, you don't even need these error states. Because you can simply check whether the title exists or not, as it is done in the handler function below.

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,22 @@
...
+import { Todo } from '../../types/types';
+import usersFromServer from '../../api/users';
+import { UserInfo } from '../UserInfo';
+
+interface Props {
+  todo: Todo;
+}
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const foundUser = usersFromServer.find(u => u.id === todo.userId);
```
lesdyman, line 11: Take another look at this variable 'u'. Perhaps it would be clearer if it had a more specific name, such as 'person', 'userUnit' or something similar.

lesdyman (approved):
Well done! This is really nice solution. I've left some thoughts in the comments though.  You may find them useful or not))

src/App.tsx
```diff
@@ -1,27 +1,97 @@
...
+    let hasError = false;
+
+    if (title.trim() === '') {
+      setTitleError('Please enter a title');
+      hasError = true;
+    } else {
+      setTitleError('');
+    }
+
+    if (userId === 0) {
```
roman-mirzoian, line 26: Here you can use a variable like this: ``const DEFAULT_USER_ID = 0;``  to avoid magic values.

```suggestion
    if (userId === DEFAULT_USER_ID) {
```

src/App.tsx
```diff
@@ -1,27 +1,97 @@
...
+  const [titleError, setTitleError] = useState('');
+  const [userError, setUserError] = useState('');
+
+  const handleAddTodo = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    let hasError = false;
+
+    if (title.trim() === '') {
+      setTitleError('Please enter a title');
```
roman-mirzoian, line 20: error messages can also be output in variables

```
const EMPTY_TITLE_ERROR = 'Please enter a title';
const EMPTY_USER_ERROR = 'Please choose a user';
```

roman-mirzoian (changes requested):
Good job also pay attention to all the comments.

roman-mirzoian (approved):
Well done!

PR #2644, opened 2024-06-29
solution

SrTrace:
[DEMO LINK](https://srtrace.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -2,18 +2,18 @@ iframe {
   display: none;
 }
 
-html {
-  font-family: Arial, "Helvetica Neue", Helvetica, sans-serif;
-}
+// html {
+//   font-family: Arial, 'Helvetica Neue', Helvetica, sans-serif;
+// }
 
-label,
-button {
-  cursor: pointer;
-}
+// label,
+// button {
+//   cursor: pointer;
+// }
 
-.error {
-  color: red;
-}
+// .error {
+//   color: red;
+// }
 
 .TodoInfo {
   width: max-content;
--- a/src/components/FormTodo/FormTodo.tsx
+++ b/src/components/FormTodo/FormTodo.tsx
@@ -1,0 +1,134 @@
+import { useState } from 'react';
+import { Todo, User } from '../../types/types';
+import usersFromServer from '../../api/users';
+import { getNewTodoId } from '../../utils/utils';
+import cn from 'classnames';
+
+type FormTodoProps = {
+  todos: Todo[];
+  addTodo: (todo: Todo) => void;
+};
+
+export const FormTodo: React.FC<FormTodoProps> = ({ todos, addTodo }) => {
+  const [users] = useState<User[]>(usersFromServer);
+
+  const [title, setTitle] = useState('');
+  const [hasTitleErr, setHasTitleErr] = useState(false);
+
+  const [userId, setUserId] = useState(0);
+  const [hasUserIdErr, setHasUserIdErr] = useState(false);
+
+  const handleTitleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(e.target.value);
+    setHasTitleErr(false);
+  };
+
+  const handleUserIdChange = (e: React.ChangeEvent<HTMLSelectElement>) => {
+    setUserId(+e.target.value);
+    setHasUserIdErr(false);
+  };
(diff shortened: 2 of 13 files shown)


src/App.scss
```diff
@@ -2,18 +2,18 @@ iframe {
...
+// button {
+//   cursor: pointer;
+// }
 
-.error {
-  color: red;
-}
+// .error {
+//   color: red;
+// }
```
VitaliyBondarenko1982, line 16: why commented these styles?

src/components/FormTodo/FormTodo.tsx
```diff
@@ -0,0 +1,134 @@
...
+    if (!title || !userId) {
+      return;
+    }
+
+    addTodo({
+      id: getNewTodoId(todos),
+      title,
+      completed: false,
+      userId,
+      user: usersFromServer.find(user => user.id === userId)!,
```
VitaliyBondarenko1982, line 54: fix linter warning.
find returns undefined, if element not found.
```suggestion
      user: usersFromServer.find(user => user.id === userId),
```

VitaliyBondarenko1982 (changes requested):
Good job.
Check a couple minor comments

src/App.scss
SrTrace: I tried to use Bulma for styling, so these styles were explicit but needed for tests, so I commented them out. However, it still works in both cases, so this time I included these styles.

src/components/FormTodo/FormTodo.tsx
SrTrace: I found only two ways to fix the ESLint warning if the find method returns undefined: using the ! sign and hardcoding an object. If there is a better option, please let me know.

src/utils/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export function getTodosWithUserInfo(): Todo[] {
+  return todosFromServer.map((todo: Todos) => ({
+    ...todo,
+    user: usersFromServer.find((user: User) => user.id === todo.userId) || {
+      id: -1,
+      name: 'Unknown',
+      username: 'unknown',
+      email: 'unknown@example.com',
+    },
```
etojeDenys, line 20:
```suggestion
    user: usersFromServer.find((user: User) => user.id === todo.userId) || null,
```
better to return null if find method doesn't find a person

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,19 @@
...
+};
+export const TodoInfo: React.FC<TodoInfoProps> = ({ todo }) => {
+  return (
+    <article
+      data-id={todo.id}
+      className={cn('TodoInfo', { 'TodoInfo--completed': todo.completed })}
+    >
+      <h2 className="TodoInfo__title">{todo.title}</h2>
+
+      <UserInfo user={todo.user} />
```
etojeDenys, line 16:
```suggestion
      {todo.user && <UserInfo user={todo.user} />}
```
and just add the check here

src/components/FormTodo/FormTodo.tsx
```diff
@@ -0,0 +1,139 @@
...
+import { getNewTodoId } from '../../utils/utils';
+import cn from 'classnames';
+
+type FormTodoProps = {
+  todos: Todo[];
+  addTodo: (todo: Todo) => void;
+};
+
+export const FormTodo: React.FC<FormTodoProps> = ({ todos, addTodo }) => {
+  const [users] = useState<User[]>(usersFromServer);
```
etojeDenys, line 13: why do you need this state? looks redundant

etojeDenys (changes requested):

src/components/FormTodo/FormTodo.tsx
SrTrace: The idea of using useState for users is coming to my mind. If we imagine that userFromServer is coming from the server, then useState can save this data after the first render, and subsequent renders will use this saved data in users. This is just my model. If we need to change users or add new ones, we already have the instrument for this. But in this case, it may be unnecessary.

SanyaBratashchuk: makes sense 👍

SanyaBratashchuk (approved):

PR #2645, opened 2024-06-30
sol1

sa4ok-1:
[DEMO LINK](https://sa4ok-1.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,122 @@
+import React, { useState } from 'react';
 import './App.scss';
+import { TodoList } from './components/TodoList';
+import { User } from './types/User';
+import { TodoItem } from './types/TodoItems';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,130 @@
...
+  return maxId + 1;
+}
+
+export const App: React.FC = () => {
+  const [todo, setTodo] = useState<TodoItem[]>(
+    todosFromServer.map(todos => ({
+      ...todos,
+      user: getUserById(todos.userId),
+    })),
+  );
```
yuliaPel, line 29: I think it would be better to get the value before and use the name 'todos' instead of 'todo'

src/App.tsx
```diff
@@ -1,61 +1,130 @@
...
+          <select
+            id="user"
+            name="user"
+            data-cy="userSelect"
+            required
+            value={userId}
+            onChange={event => {
+              setUserId(+event.target.value);
+              setHasUserIdError(false);
+            }}
```
yuliaPel, line 105: it would be better to create separate handler for this

yuliaPel (approved):
Good job but your link doesn't work.

PR #2655, opened 2024-07-04
Develop

Arthur-Tolkachov:
## Description
[DEMO LINK](https://arthur-tolkachov.github.io/react_add-todo-form/)


anastasiiavorobiova (changes requested):
Great work! A user should be able to add a todo with spaces between words in the title. Consider just trimming the title and checking if it has symbols after the trim.

anastasiiavorobiova (approved):
Awesome!

PR #2667, opened 2024-07-14
My solution

Karanelus:
[DEMO LINK](https://karanelus.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,114 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/todo';
+import { useState } from 'react';
+
+type TodosProperties = {
+  todos: Todo[];
+  titleValue: string;
+  userValue: string;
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
+        isFormSubmited: true,
+      }));
+    }
+
+    const userIdSearch = usersFromServer.find(
+      user => user.name === userValue,
+    )?.id;
+
+    const newTodo: Todo = {
+      id: todos.length + 1,
```
natalia-klonowska, line 47: instead of creating id based on length you should find max id and increase it:
image

src/App.tsx
```diff
@@ -1,61 +1,122 @@
...
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <label htmlFor="titleInput">Title: </label>
+          <input
+            id="titleInput"
+            type="text"
+            data-cy="titleInput"
+            value={titleValue}
+            onChange={handleChangeTodosState}
+          />
```
natalia-klonowska, line 85: it should have a placeholder attribute 
image

natalia-klonowska (changes requested):
fix to pass tests and linter error

natalia-klonowska (approved):
Great work 😃

PR #2671, opened 2024-07-18
develop

mulytskaya:
https://mulytskaya.github.io/react_add-todo-form/


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You don't need to change tests. If you have problems with tests feel free in the fe_chat
image
image

PR #2672, opened 2024-07-21
Solution

ZakZh:
[DEMO LINK](https://ZakZh.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/AddTodoForm/AddTodoForm.tsx
+++ b/src/components/AddTodoForm/AddTodoForm.tsx
@@ -1,0 +1,101 @@
+import React, { useState } from 'react';
+import './AddTodoForm.scss';
+import { Todo, User } from '../../utils/types';
+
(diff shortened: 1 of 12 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your cpde
image

src/components/AddTodoForm/AddTodoForm.tsx
```diff
@@ -0,0 +1,103 @@
...
+          onChange={handleUserChange}
+        >
+          <option value="0">Choose a user</option>
+          {users.map(({ id, name }) => {
+            return (
+              <option key={id} value={id}>
+                {name}
+              </option>
+            );
+          })}
```
volodymyr-soltys97, line 91:
```suggestion
          {users.map(({ id, name }) => (
            <option key={id} value={id}>
              {name}
            </option>
          ))}
```

volodymyr-soltys97 (approved):
Good job 👍

PR #2686, opened 2024-08-07
add task solution

AndriiZakharenko:
[DEMO LINK](https://AndriiZakharenko.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,60 +1,126 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import React, { useState } from 'react';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Todo } from './components/types/Todo';
 
-export const App = () => {
+export function getUserById(userId: number) {
+  return usersFromServer.find(user => user.id === userId) || null;
+}
+
+function getTodosId(todoId: Todo[]) {
+  const maxTodoId = Math.max(...todoId.map(todo => todo.id));
+
+  return maxTodoId + 1;
+}
+
+const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [hasTitleError, sethasTitleError] = useState(false);
+
+  const [userId, setUserId] = useState(0);
+  const [userIdError, setUserIdError] = useState(false);
+  const [todoList, setTodoList] = useState<Todo[]>(todos);
+  const newId = getTodosId(todoList);
+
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,60 +1,126 @@
...
+}));
+
+export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [hasTitleError, sethasTitleError] = useState(false);
+
+  const [userId, setUserId] = useState(0);
+  const [userIdError, setUserIdError] = useState(false);
+  const [todoList, setTodoList] = useState<Todo[]>(todos);
+  const newId = getTodosId(todoList);
```
BudnikOleksii, line 31: Generate `newId` only when user submits the form

src/App.tsx
```diff
@@ -1,60 +1,126 @@
...
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+    setUserIdError(!userId);
+
+    if (!title.trim()) {
+      sethasTitleError(true);
+    }
+
+    if (userId > 0 && title.trim() !== '') {
```
BudnikOleksii, line 56: - Don't compare with an empty string, just check if a value exists.
- `title.trim()` you do it twice, create a variable

src/App.tsx
```diff
@@ -1,60 +1,126 @@
...
+      sethasTitleError(false);
+      setUserIdError(false);
+    } else {
+      return;
+    }
```
BudnikOleksii, line 61: Avoid `else` statements if it possible

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,23 @@
...
+import { UserInfo } from '../UserInfo';
+import { Todo } from '../types/Todo';
+import { getUserById } from '../../App';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const getUser = getUserById(todo.userId);
```
BudnikOleksii, line 12: - `user` is a more suitable name
- And actually better just to store information about a user in todo, and just pass it through props. As this component is presentation component and shouldn't have any logic

BudnikOleksii (changes requested):
Good work 👍 
Left a few comments, let's improve your solution a bit

src/App.tsx
```diff
@@ -1,60 +1,121 @@
...
+function getNextTodoId(todoList: Todo[]) {
+  const maxTodoId = Math.max(...todoList.map(todo => todo.id));
+
+  return maxTodoId + 1;
+}
+
+const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
```
DanilWeda, line 22: You can move this func outside, as helpers/utils function)

dothearts (approved):
Good job. 
I approved with small comments. 
Please fix it. 
Many thanks)

PR #2693, opened 2024-08-08
solution

coldshein:
[DEMO LINK](https://coldshein.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,104 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import React, { useState } from 'react';
+import { Todo, User } from './types';
+import { TodoList } from './components/TodoList';
+
+function getUserById(userId: number): User | null {
+  return usersFromServer.find(user => user.id === userId) || null;
+}
+
+const initialTodos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
 
 export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>(initialTodos);
+
+  const [title, setTitle] = useState('');
+  const [hasTitleError, setHasTitleError] = useState(false);
+
+  const [userId, setUserId] = useState(0);
+  const [hasUserIdError, setHasUserIdError] = useState(false);
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+    setHasTitleError(false);
+  };
+
+  const handleUserChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setUserId(+event.target.value);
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,61 +1,112 @@
...
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/todos" method="POST">
+      <form action="/api/todos" method="POST" onSubmit={handleSubmit}>
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <input
```
k-marchuk, line 73: Seems `label`s are missing.

k-marchuk (approved):
Good job :rocket: 
Just one small comment.

src/App.tsx
```diff
@@ -1,61 +1,112 @@
...
+import { TodoList } from './components/TodoList';
+
+function getUserById(userId: number): User | null {
+  return usersFromServer.find(user => user.id === userId) || null;
+}
+
+const initialTodos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
```
VenomMB, line 16: better move to separate file for example utils.ts

src/App.tsx
```diff
@@ -1,61 +1,112 @@
...
 
 export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>(initialTodos);
+
+  const [title, setTitle] = useState('');
+  const [hasTitleError, setHasTitleError] = useState(false);
+
+  const [userId, setUserId] = useState(0);
```
VenomMB, line 24:
```suggestion
  const [userId, setUserId] = useState(null);
```
better use null

src/App.tsx
```diff
@@ -1,61 +1,112 @@
...
+  const [hasUserIdError, setHasUserIdError] = useState(false);
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+    setHasTitleError(false);
+  };
+
+  const handleUserChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setUserId(+event.target.value);
```
VenomMB, line 33:
```suggestion
    setUserId(Number(event.target.value));
```
it's more readable

src/App.tsx
```diff
@@ -1,61 +1,112 @@
...
         <div className="field">
-          <select data-cy="userSelect">
+          <select
+            data-cy="userSelect"
+            value={userId}
+            onChange={handleUserChange}
+          >
             <option value="0" disabled>
               Choose a user
             </option>
```
VenomMB, line 91: add empty line

VenomMB (changes requested):
Great job, let's work on improvements now

src/App.tsx
coldshein: > better use null

I don't think so, this state is used for the initial state of id and also as a placeholder value for the option of select element

DanilWeda (approved):
Good job. 
Approved! 
Many thanks)

PR #2697, opened 2024-08-09
Add solution

OleksiiKislukhin:
[DEMO LINK](https://OleksiiKislukhin.github.io/react_add-todo-form/)

- [x] Allow entering only letters (`ua` and `en`), digits, and `spaces` in the `title` field. Just remove any other characters from the `title`.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,115 @@
+import React, { useState } from 'react';
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { getTodosWithUsers } from './services/getTodosWithUsers';
+import { getUserById } from './services/getUserById';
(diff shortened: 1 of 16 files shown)


src/App.tsx
```diff
@@ -1,61 +1,123 @@
...
+    setIsTitleHasError(false);
+
+    setUserId(0);
+    setIsUserIdHasError(false);
+  }
+
+  function handleSubmit(event: React.FormEvent) {
+    event.preventDefault();
+
+    const isTitleValid = title !== '';
```
BudnikOleksii, line 46: You don't need to compare with an empty string, just check if value exists

src/App.tsx
```diff
@@ -1,61 +1,123 @@
...
+
+      return;
+    }
+
+    const newTodo = {
+      id: getNextFreeId(todos),
+      title: title,
+      completed: false,
+      userId: userId,
+      user: getUserById(usersFromServer, userId) as User,
```
BudnikOleksii, line 61: Avoid `as`

BudnikOleksii (approved):
Looks good 🚀

PR #2699, opened 2024-08-10
Done?

NIKaragu:
[DEMO LINK](https://NIKaragu.github.io/react_add-todo-form/)

P.S.: I've been having a BIIIIIIIIIIIIG problem with tests...
I had been coding and for own comfort (but as it turns out, I've only added a lot of problems to me) created a new data structure to develop the project pushing off from this structure... and finally I've finished and it has been working well, but problems didn't end there...
When I executed tests I noticed that aaaaall tests configurated only on working with base data structures and to pass all these tests I had to rewrite the project entirely... And it was so disgusting... So I just rewrited test for my custom data structure. I didn't touch cypress test, 'cause it was already too much for me, so I just commented them out.

Conclusion: you can check working of the project (it's good). But also I'll understand if you say I have to remake it from 0.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,213 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import React, { useState } from 'react';
+
+export type User = (typeof usersFromServer)[0];
+
+export interface Todo {
+  id: number;
+  title: string;
+  completed: boolean;
+  userId: number;
+  createdAt: Date;
+}
+
+export interface UserTodos {
+  userId: number;
+  userName: string;
+  userEmail: string;
+  userLogin: string;
+  todoPosts: Todo[];
+}
+
+const usersTodos = usersFromServer
+  .map(user => {
+    const userTodo = todosFromServer.reduce((todosAcc: Todo[], todo) => {
+      if (user.id === todo.userId) {
+        todosAcc.push({ ...todo, ...{ createdAt: new Date() } });
+
+        return todosAcc;
+      }
+
+      return todosAcc;
+    }, []);
+
+    return {
+      userId: user.id,
+      userName: user.name,
+      userEmail: user.email,
+      userLogin: user.username,
+      todoPosts: userTodo,
+    };
+  })
+  .filter(userTodo => userTodo.todoPosts.length > 0);
+
+function findTheBiggestPostId(postList: UserTodos[]): number {
+  return Math.max(
+    ...postList.map(userPost =>
+      Math.max(...userPost.todoPosts.map(post => post.id)),
+    ),
+  );
+}
 
 export const App = () => {
+  // #region baseStates
+  const [currentUsersTodos, setCurrentUsersTodos] =
+    useState<UserTodos[]>(usersTodos);
+  const [postCount, setPostCount] = useState(
+    findTheBiggestPostId(currentUsersTodos) + 1,
+  );
+  // #endregion
+  // #region title
+  const [title, setTitle] = useState('');
+  // #endregion
+  // #region userSelecting
+  const [isUserSelected] = useState(false);
+  const [userSelectingValue, setUserSelectingValue] = useState<number>(0);
+  // #endregion
+
+  // #region fieldErrors
+  const [hasTitleAnError, setHasTitleAnError] = useState<boolean>(false);
+  const [hasSelectedUserAnError, setHasSelectedUserAnError] =
+    useState<boolean>(false);
+
+  const titleFieldHasAnError = title.length === 0;
+  const selectedUserFieldHasAnError = userSelectingValue === 0;
+  const formHasSomeError = hasTitleAnError || hasSelectedUserAnError;
+  // #endregion
+
+  // #region handlers
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+    setHasTitleAnError(false);
+  };
+
+  const handleSelectedUser = (userId: number): User => {
+    return usersFromServer.find(user => user.id === userId) as User;
+  };
+
+  const onSubmit = (selectedUser: User): void => {
+    const existingUser = currentUsersTodos.find(
+      user => user.userId === selectedUser.id,
+    );
+
+    const newTodo = {
+      id: postCount,
+      title: title,
+      completed: false,
+      userId: selectedUser.id,
+      createdAt: new Date(),
+    };
+
+    if (existingUser) {
+      existingUser.todoPosts.push(newTodo);
+
+      return setCurrentUsersTodos([...currentUsersTodos]);
+    }
+
+    return setCurrentUsersTodos([
+      ...currentUsersTodos,
+      {
+        userId: selectedUser.id,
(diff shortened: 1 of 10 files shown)


RimmaBuzuluk (changes requested):
image
hi, good job

but you have "error" problem.

1) I try to send a request without data (errors appear - this is good) :))

2) I add information about the user and the error does not disappear - this is a problem

NIKaragu:
> image hi, good job
> 
> but you have "error" problem.
> 
> 1. I try to send a request without data (errors appear - this is good) :))
> 2. I add information about the user and the error does not disappear - this is a problem

done 👌

RimmaBuzuluk (approved):

RimmaBuzuluk:
good job :))

src/App.tsx
```diff
@@ -1,61 +1,218 @@
...
 
 export const App = () => {
+  // #region baseStates
+  const [currentUsersTodos, setCurrentUsersTodos] =
+    useState<UserTodos[]>(usersTodos);
+  const [postCount, setPostCount] = useState(
+    findTheBiggestPostId(currentUsersTodos) + 1,
+  );
+  // #endregion
+  // #region title
```
BudnikOleksii, line 64: Remove all comments

src/App.tsx
```diff
@@ -1,61 +1,218 @@
...
+  .map(user => {
+    const userTodo = todosFromServer.reduce((todosAcc: Todo[], todo) => {
+      if (user.id === todo.userId) {
+        todosAcc.push({ ...todo, ...{ createdAt: new Date() } });
+
+        return todosAcc;
+      }
+
+      return todosAcc;
+    }, []);
```
BudnikOleksii, line 36: You don't have to make your app more complicated

src/App.tsx
```diff
@@ -1,61 +1,218 @@
...
+      completed: false,
+      userId: selectedUser.id,
+      createdAt: new Date(),
+    };
+
+    if (existingUser) {
+      existingUser.todoPosts.push(newTodo);
+
+      return setCurrentUsersTodos([...currentUsersTodos]);
+    }
```
BudnikOleksii, line 114: What is the point of doing it? And what is the reason of `existingUser` if you don't use this variable in the future

src/App.tsx
```diff
@@ -1,61 +1,218 @@
...
+        todoPosts: [
+          {
+            id: postCount,
+            title: title,
+            completed: false,
+            userId: selectedUser.id,
+            createdAt: new Date(),
+          },
+        ],
+      },
```
BudnikOleksii, line 132: When you need to update the state depending on the previous one use functional `setState`, and you don't need to store this data in the state at all

src/App.tsx
```diff
@@ -1,61 +1,218 @@
...
         </div>
 
-        <button type="submit" data-cy="submitButton">
+        <button
+          type="submit"
+          data-cy="submitButton"
+          onClick={() => {
+            setHasTitleAnError(titleFieldHasAnError);
+            setHasSelectedUserAnError(selectedUserFieldHasAnError);
+          }}
```
BudnikOleksii, line 209: It's a submit button, do validations in the submit handler instead of adding additional handler to onclick

src/components/TodoInfo/TodoInfo.spec.tsx
```diff
@@ -3,40 +3,30 @@ import { TodoInfo } from './TodoInfo';
...
-      username: 'Bret',
-      email: 'Sincere@april.biz',
+    const user1Todo = {
+      userName: 'Leanne Graham',
+      userEmail: 'Sincere@april.biz',
+      todoPost: {
+        id: 1,
+        title: 'Learn HTML',
+        completed: true,
+      },
```
BudnikOleksii, line 13: Don't change or remove tests

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,32 @@
...
+  userId: number;
+  userName: string;
+  userEmail: string;
+  userLogin: string;
+  todoPost: Pick<Todo, 'completed' | 'id' | 'title'>;
+}
+
+type Props = {
+  todo: Pick<UserTodo, 'userName' | 'userEmail' | 'todoPost'>;
+};
```
BudnikOleksii, line 15: It doesn't make much sense to create another one interface and then just pick a few properties from it

BudnikOleksii (changes requested):
That good that you have practiced how to work with different methods and so on.
But in this case, it's [overengineering](https://en.wikipedia.org/wiki/Overengineering). When you have clear requirements don't try to do it more complicated. The solution should be as simple as possible, so you can easily maintain and extend it if needed.
Let's try to simplify your solution
https://medium.com/@trienpont/when-less-is-more-the-danger-of-over-engineering-4bb1579d0c06

BudnikOleksii:
> [DEMO LINK](https://NIKaragu.github.io/react_add-todo-form/)
> 
> P.S.: I've been having a BIIIIIIIIIIIIG problem with tests... I had been coding and for own comfort (but as it turns out, I've only added a lot of problems to me) created a new data structure to develop the project pushing off from this structure... and finally I've finished and it has been working well, but problems didn't end there... When I executed tests I noticed that aaaaall tests are configurated only on working with base data structures and to pass all these tests I had to rewrite the project entirely... And it was so disgusting... So I just rewrited test for my custom data structure. I didn't touch cypress test, 'cause it was already too much for me, so I just commented them out.
> 
> **Conclusion**: you can check working of the project (it's good). But also I'll understand if you say I have to remake it from 0. And could you please leave a feedback about this solution (relevance and development comfort for future) of the task from a professional point of view)
> 
> **UPD: !important to notice** - current data structure is very comfortable (for future) to track, for example, user's activity ('cause every post is connected to user directly) and ensure compliance with the site's policy on censorship in publications, so there we have a big plus) but only in perspective. A big minus - very hard and routine sorting and maping by "todo post's" data creation, I guess it takes a lot of resourses 'cause of nested loops

And in real world you don't need to create structure like this, they are not so useful. For example you want additional information about users todos, so you don't create another structure, you just extend the one that you have:
```typescript
interface User {
  id: number;
  name: string;
  email: string;
  password: string;
}

interface UseerWithTodos extends User {
  todos: Todo[];
}
```
And in ideal world you get this from the API, as you don't store all tods at the client

src/App.tsx
NIKaragu: it's for more comfortable code view and fast orientation in my code, it's just for convenience.

src/App.tsx
NIKaragu: Here I created a new data structure and filtered users only having any todo posts

src/App.tsx
NIKaragu: if current user has any post in his `todoPost` list, then I just add one more post in his post's history insted of creating the same user with new post

src/App.tsx
NIKaragu: Here If I don't have definite user with any post, I update `currentUsersTodos` and add new user (selected user) with new post in his post's history

src/App.tsx
NIKaragu: yea, reasonably

src/components/TodoInfo/TodoInfo.spec.tsx
NIKaragu: so, then I have to remake this project( ehh

PR #2701, opened 2024-08-11
add solution

RimmaBuzuluk:
[DEMO LINK](https://RimmaBuzuluk.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,125 @@
+import { useState } from 'react';
 import './App.scss';
+// import classNames from 'classnames';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Todo, TodoUser, User } from './interfaces';
+
+const todosUsersFiltered: TodoUser[] = todosFromServer.map((todo: Todo) => {
+  const matchUser = usersFromServer.find(
+    (user: User) => user.id === todo.userId,
+  );
+
+  return {
+    ...todo,
+    user: matchUser,
+  };
+});
 
 export const App = () => {
+  const [count, setCount] = useState(16);
+  const [title, setTitle] = useState('');
+  const [selectedUser, setSelectedUser] = useState(0);
+  const [errTitle, setErrTitle] = useState(false);
+  const [errUser, setErrUser] = useState(false);
+  const [todoUsers, setTodoUser] = useState<TodoUser[]>(todosUsersFiltered);
+  const usersArr = [
+    {
+      id: 0,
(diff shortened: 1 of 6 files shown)


victoriasokolvak (approved):
Looks good!

src/App.tsx
```diff
@@ -1,61 +1,125 @@
...
+    if (selectedUser === 0) {
+      setErrUser(true);
+      hasError = true;
+    } else {
+      setErrUser(false);
+    }
+
+    if (hasError) {
+      return;
+    }
```
ArturErmolenko, line 69: can we simplify it?
let's try not to use else & let in code
 if (selectedUser === 0 || title === '') {
      setErrUser(true);
return
}

src/App.tsx
```diff
@@ -1,61 +1,125 @@
...
+
+    const user = usersFromServer.find(({ id }) => id === selectedUser);
+
+    const obj = {
```
ArturErmolenko, line 73: what mean obj?

src/App.tsx
```diff
@@ -1,61 +1,125 @@
...
+  return {
+    ...todo,
+    user: matchUser,
+  };
+});
 
 export const App = () => {
+  const [count, setCount] = useState(16);
+  const [title, setTitle] = useState('');
+  const [selectedUser, setSelectedUser] = useState(0);
```
ArturErmolenko, line 24: user can be 0
maybe user id?
but we never have id = 0

src/App.tsx
```diff
@@ -1,61 +1,125 @@
...
+  const [errTitle, setErrTitle] = useState(false);
+  const [errUser, setErrUser] = useState(false);
+  const [todoUsers, setTodoUser] = useState<TodoUser[]>(todosUsersFiltered);
+  const usersArr = [
+    {
+      id: 0,
+      name: 'Choose a user',
```
ArturErmolenko, line 31: can it be placeholder or smth else?

src/App.tsx
```diff
@@ -1,61 +1,125 @@
...
+      title,
+      userId: selectedUser,
+      completed: false,
+      user,
+    };
+
+    setTodoUser([...todoUsers, obj]);
+    setCount(count + 1);
+    setTitle('');
+    setSelectedUser(0);
```
ArturErmolenko, line 84: create reset function

ArturErmolenko (changes requested):
GJ 👍 
Let's improve code readability ;)

src/App.tsx
RimmaBuzuluk: I can add a placeholder, but in the video for this task, Misha showed exactly this method (when the index 0 stores a default value that cannot be selected)

ArturErmolenko (approved):
GJ 👍

PR #2707, opened 2024-08-13
Solution

onikishyna:
https://onikishyna.github.io/react_add-todo-form/

### DIFF
--- a/src/types/props.ts
+++ b/src/types/props.ts
@@ -1,0 +1,5 @@
+import { Todo } from './todo';
+
+export interface Props {
+  todos: Todo[];
+}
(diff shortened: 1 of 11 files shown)


YarikKedrych (approved):
not bed)

src/types/props.ts
```diff
@@ -0,0 +1,5 @@
+import { Todo } from './todo';
+
+export interface Props {
+  todos: Todo[];
+}
```
BudnikOleksii, line 5: You don't need to create files for your props interfaces, keep the along with component

src/types/propsUser.ts
```diff
@@ -0,0 +1,5 @@
+import { User } from './user';
+
+export interface PropsUser {
+  user: User | undefined;
```
BudnikOleksii, line 4:
```suggestion
  user?: User;
```
It does the same

BudnikOleksii (approved):
Good job 👍

PR #2724, opened 2024-08-23
add solution

Mazurok001:
[DEMO LINK](https://Mazurok001.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,118 @@
 import './App.scss';
+import React, { useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
 
-export const App = () => {
+import User from './types/User';
+import Todo from './types/Todo';
+
+function getUserById(userId: number): User | null {
+  return usersFromServer.find(user => user.id === userId) || null;
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [currentTodos, setCurrentTodos] = useState<Todo[]>(todos);
+
+  const [title, setTitle] = useState('');
+  const [selectedUserId, setSelectedUserId] = useState(0);
+
+  const [errorTitle, setErrorTitle] = useState(false);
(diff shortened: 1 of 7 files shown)


src/types/User.ts
```diff
@@ -0,0 +1,6 @@
+export default interface User {
```
MaxBorovique, line 1: "default" is not necessary here. As I can understand, the interface User will always be only one in this document

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,17 @@
...
+import { UserInfo } from '../UserInfo/UserInfo';
+import Todo from '../../types/Todo';
+
+interface Props {
+  todo: Todo;
+}
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => (
+  <article
+    className={`TodoInfo ${todo.completed === true ? 'TodoInfo--completed' : ''}`}
```
MaxBorovique, line 11: To avoid mistakes and make your code more readable in the future I strongly recommend you to use classNames in the same cases.
And in this case, if you write todo.completed it means it's already true, so you don't need to compare it with true.
className={classNames('TodoInfo', {
    })}

src/App.tsx
```diff
@@ -1,27 +1,118 @@
...
-          <span className="error">Please enter a title</span>
+          <label className="label" htmlFor="titleInput">
+            Title:&nbsp;
+            <input
+              type="text"
+              data-cy="titleInput"
+              id="titleInput"
+              placeholder="Enter a title"
+              value={title}
+              onChange={event => {
```
MaxBorovique, line 79: If your onChange element uses more than one function the best practice is to create a handler.  Something like that:
const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
    setTitle(event.target.value);
    setErrorTitle(false);
  };

src/App.tsx
```diff
@@ -1,27 +1,118 @@
...
-          <select data-cy="userSelect">
+          <label className="label" htmlFor="userSelected">
+            User:{' '}
+          </label>
+
+          <select
+            data-cy="userSelect"
+            id="userSelected"
+            value={selectedUserId}
+            onChange={event => {
```
MaxBorovique, line 98: The same situation with handler

MaxBorovique (changes requested):
Well done! I suggest making a few changes to enhance your code.

ArturErmolenko (approved):
Nicely done 🔥

PR #2737, opened 2024-09-09
Done

v-zagorovskii:
[DEMO LINK](https://v-zagorovskii.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,7 +15,7 @@ function getUserById(id: number): User | undefined {
 
 function newListWithUser(): TodoWithUser[] {
   return todosFromServer.map(todo => ({
-    todo,
+    ...todo,
     user: getUserById(todo.userId),
   }));
 }
@@ -58,23 +58,20 @@ export const App = () => {
       return;
     }
 
-    const lastIndex =
-      listWithUser.length > 0
-        ? listWithUser[listWithUser.length - 1].todo.id + 1
-        : 1;
+    const maxIndex = Math.max(...listWithUser.map(el => el.id));
 
     const personFromChoise = getUserById(+choise);
 
     const todoFromForm: Todo = {
-      id: lastIndex,
+      id: maxIndex + 1,
       title: title,
       completed: false,
       userId: personFromChoise?.id,
     };
 
     const newItem: TodoWithUser = {
+      ...todoFromForm,
       user: personFromChoise,
-      todo: todoFromForm,
     };
 
     setListWithUser(currentList => [...currentList, newItem]);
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,139 @@
...
+
+import { User } from './services/User';
+import { Todo } from './services/Todo';
+import { TodoWithUser } from './services/TodoWithUser';
+
+function getUserById(id: number): User | undefined {
+  return usersFromServer.find(user => user.id === id);
+}
+
+function newListWithUser(): TodoWithUser[] {
```
TarasHoliuk, line 16:
```suggestion
function getTodosWithUser(): TodoWithUser[] {
```
A function should contain a verb, otherwise it seems like `newListWithUser` variable is already prepared list. Also, it's better to be more precise - it's array of todos, "list" word says nothing here

src/App.tsx
```diff
@@ -1,61 +1,139 @@
...
+    ...todo,
+    user: getUserById(todo.userId),
+  }));
+}
 
 export const App = () => {
+  const [listWithUser, setListWithUser] =
+    useState<TodoWithUser[]>(newListWithUser());
+  const [title, setTitle] = useState('');
+  const [choise, setChoice] = useState('0');
```
TarasHoliuk, line 27:
```suggestion
  const [userId, setUserId] = useState('0');
```
type in choice word.

But again - better to rename it - we store userId here. Choice says nothing about this variable value

src/App.tsx
```diff
@@ -1,61 +1,139 @@
...
         </div>
 
         <div className="field">
-          <select data-cy="userSelect">
+          <label htmlFor="user">User: </label>
+          <select
+            id="user"
+            data-cy="userSelect"
+            value={choise}
+            onChange={e => {
```
TarasHoliuk, line 111: move it to a separate handler and use it here

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,21 @@
...
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  return (
+    <article
+      data-id={todo.id}
+      className={cn('TodoInfo', {
+        'TodoInfo--completed': todo.completed === true,
+      })}
+    >
+      <h2 className="TodoInfo__title">{todo.title}</h2>
+      {todo.user !== undefined && <UserInfo user={todo.user} />}
```
TarasHoliuk, line 18:
```suggestion
      {todo.user && <UserInfo user={todo.user} />}
```
BTW, you can use restructuring assignment to get all properties from `todo`. Same in other components

TarasHoliuk (changes requested):

TarasHoliuk (approved):
✔️

PR #2754, opened 2024-09-24
add solution

VladTk:
[DEMO LINK](https://VladTk.github.io/react_add-todo-form/)


lerastarynets (approved):
You're on the right track!

PR #2760, opened 2024-10-01
Develop

mikezhylka:
Hello

[DEMO LINK](https://mikezhylka.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/TodoForm/TodoForm.tsx
+++ b/src/components/TodoForm/TodoForm.tsx
@@ -1,0 +1,104 @@
+import React, { useState } from 'react';
+import usersFromServer from '../../api/users';
+import { Todo } from '../../types/Todo';
+
+interface TodoFormProps {
+  onAddTodo: (todo: Todo) => void;
+  maxId: number;
+}
+
+export const TodoForm: React.FC<TodoFormProps> = ({ onAddTodo, maxId }) => {
+  const [todo, setTodo] = useState({
+    id: 0,
+    title: '',
+    completed: false,
+    userId: 0,
+  });
+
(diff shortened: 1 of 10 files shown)


lerastarynets (changes requested):
addTodoForm

Pls try to pass all tests, if your tests pass locally, but you encounter a problem in server tests, pls attach a screen of all local tests passing

mikezhylka:
I fixed problem with submit and whitespaces, but tests do not pass for UserInfo component, currentUser prop of UserInfo returns undefined, but I use conditional rendering {currentUser && <UserInfo />} in TodoInfo, so idk why currentUser potentially can be undefined, maybe problems in tests. This is a conclusion that we made with mentor inchat :)

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,104 @@
...
+    setTodo(prevTodo => ({
+      ...prevTodo,
+      title: event.target.value,
+      id: maxId + 1,
+    }));
+
+    setIsTitleEntered(true);
+  }
+
+  function handleSelectChange(event: React.ChangeEvent<HTMLSelectElement>) {
```
volodymyr-soltys97, line 31: It is important to follow the same way of creating functions everywhere
```suggestion
  const handleSelectChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
```

volodymyr-soltys97 (changes requested):
Good job!
To improve:
1. Where is the label text for inputs? 
image

etojeDenys (approved):
lgtm

PR #2768, opened 2024-10-04
solved task

PikalovAndrey:
[DEMO LINK](https://PikalovAndrey.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,124 @@
 import './App.scss';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import React, { FormEvent, useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
 
-export const App = () => {
+export const App: React.FC = () => {
+  const preparedTodos = () => {
+    return todosFromServer.map(todo => {
+      const user = usersFromServer.find(
+        userFromServer => userFromServer.id === todo.userId,
+      );
+
+      return { ...todo, user: user || null };
+    });
+  };
+
+  const [tasks, setTasks] = useState<Todo[]>(preparedTodos());
+
+  const [userId, setUserId] = useState(0);
+  const [title, setTitle] = useState('');
+
+  const [errorUserId, setErrorUserId] = useState('');
+  const [errorTitle, setErrorTitle] = useState('');
+
+  const handleEnteredTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+
+    if (errorTitle) {
+      setErrorTitle('');
+    }
+  };
+
+  const handleUserSelect = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setUserId(+event.target.value);
+
+    if (errorUserId) {
+      setErrorUserId('');
+    }
+  };
+
+  const handleReset = () => {
+    setUserId(0);
+    setTitle('');
+    setErrorTitle('');
+    setErrorUserId('');
+  };
+
+  const handleSubmit = (event: FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (!title || userId === 0) {
+      if (!title) {
+        setErrorTitle('Please enter a title');
+      }
+
+      if (userId === 0) {
+        setErrorUserId('Please choose a user');
+      }
+
+      return;
+    }
+
+    const user = usersFromServer.find(
+      userFromServer => userFromServer.id === userId,
+    );
+
+    const newTaskId =
+      tasks.length > 0 ? Math.max(...tasks.map(task => task.id)) + 1 : 1;
+
+    const newTask: Todo = {
+      id: newTaskId,
+      title: title,
+      userId: userId,
+      completed: false,
+      user: user || null,
+    };
+
+    setTasks(currentTasks => [...currentTasks, newTask]);
+
+    handleReset();
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/todos" method="POST">
+      <form action="/api/todos" method="POST" onSubmit={handleSubmit}>
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <span>Title: </span>
+          <input
+            placeholder="Enter a title"
+            type="text"
+            data-cy="titleInput"
+            value={title}
+            onChange={handleEnteredTitle}
+          />
+          {errorTitle && <span className="error">{errorTitle}</span>}
         </div>
 
         <div className="field">
-          <select data-cy="userSelect">
+          <span>User: </span>
+          <select
+            data-cy="userSelect"
+            required
+            value={userId}
+            onChange={handleUserSelect}
+          >
             <option value="0" disabled>
               Choose a user
             </option>
+            {usersFromServer.map(user => (
+              <option key={user.id} value={user.id}>
+                {user.name}
+              </option>
+            ))}
           </select>
 
-          <span className="error">Please choose a user</span>
+          {errorUserId && <span className="error">{errorUserId}</span>}
         </div>
 
         <button type="submit" data-cy="submitButton">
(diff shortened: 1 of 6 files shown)


src/types/Todo.ts
```diff
@@ -0,0 +1,9 @@
+import { User } from './User';
+
+export interface Todo {
+  id: number;
+  title: string;
+  completed: boolean;
+  userId: number;
+  user: User | null;
```
KrivorotkoYurii, line 8: You can also use '?' instead of null. If you implement it you can also chenge the conditions in other files by removing  '  | |  null ' 
```suggestion
  user?: User;
```

src/App.tsx
```diff
@@ -1,61 +1,132 @@
...
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+
+export const App: React.FC = () => {
+  const preparedTodos = () => {
+    return todosFromServer.map(todo => {
+      const user = usersFromServer.find(
```
KrivorotkoYurii, line 12: You use usersFromServer.find(...) at leat twice (check the line 66 in this file). It is a dublicating of code. You can create a new folder 'utilities', create a file like 'CreateUser' and write a function with this logic. And after that import this function at the line 12 and 66

src/App.tsx
```diff
@@ -1,61 +1,132 @@
...
+      userFromServer => userFromServer.id === userId,
+    );
+
+    const newTaskId =
+      tasks.length > 0 ? Math.max(...tasks.map(task => task.id)) + 1 : 1;
+
+    const newTask: Todo = {
+      id: newTaskId,
+      title: title,
+      userId: userId,
```
KrivorotkoYurii, line 76: For the lines 75 and 76 you dont need to write twice

src/App.tsx
```diff
@@ -1,61 +1,132 @@
...
+    const user = usersFromServer.find(
+      userFromServer => userFromServer.id === userId,
+    );
+
+    const newTaskId =
+      tasks.length > 0 ? Math.max(...tasks.map(task => task.id)) + 1 : 1;
+
+    const newTask: Todo = {
+      id: newTaskId,
+      title: title,
```
KrivorotkoYurii, line 75:
```suggestion
      title,
```

src/App.tsx
KrivorotkoYurii, line 76:
```suggestion
      userId,
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,23 @@
...
+  return (
+    <article
+      data-id={todo.id}
+      className={cn(
+        { 'TodoInfo TodoInfo--completed': todo.completed },
+        { TodoInfo: !todo.completed },
+      )}
+    >
+      <h2 className="TodoInfo__title">{todo.title}</h2>
+      {todo.user && <UserInfo user={todo.user} />}{' '}
```
KrivorotkoYurii, line 20: You dont need a space in the end of the line
```suggestion
      {todo.user && <UserInfo user={todo.user} />}
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,23 @@
...
+  todo: Todo;
+}
+
+export const TodoInfo: React.FC<TodoInfoProps> = ({ todo }) => {
+  return (
+    <article
+      data-id={todo.id}
+      className={cn(
+        { 'TodoInfo TodoInfo--completed': todo.completed },
+        { TodoInfo: !todo.completed },
```
KrivorotkoYurii, line 16: You dont need this condition here. If you remove this condition and let className TodoInfo by default it will be the same logic.
```suggestion
       className={cn('TodoInfo', {
        'TodoInfo--completed': todo.completed,
      })}
```

src/App.tsx
```diff
@@ -1,61 +1,132 @@
...
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/todos" method="POST">
+      <form action="/api/todos" method="POST" onSubmit={handleSubmit}>
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <span>Title: </span>
```
KrivorotkoYurii, line 92: According to task check-list :
*add labels and placeholders where they are needed;

You have to change your span tag by a <label> here and at the line 104
You can also check it if you enter the working version example and try to click on label Title  you will be able to start entering  somethig in the input field.
```suggestion
        <label className="label" htmlFor={'title'}>
          Title:{' '}
        </label>
```

src/App.tsx
```diff
@@ -1,61 +1,132 @@
...
+          <input
+            placeholder="Enter a title"
+            type="text"
```
KrivorotkoYurii, line 95: After you will add the label above at the line 92, you also have to add a id property to your input
```suggestion
            type="text"
            id="title"
```

src/App.tsx
```diff
@@ -1,61 +1,132 @@
...
+            data-cy="titleInput"
+            value={title}
+            onChange={handleEnteredTitle}
+          />
+          {errorTitle && <span className="error">{errorTitle}</span>}
         </div>
 
         <div className="field">
-          <select data-cy="userSelect">
+          <span>User: </span>
```
KrivorotkoYurii, line 104: Here also you should change span into label

src/App.tsx
```diff
@@ -1,61 +1,132 @@
...
+    handleReset();
+  };
 
-export const App = () => {
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <form action="/api/todos" method="POST">
+      <form action="/api/todos" method="POST" onSubmit={handleSubmit}>
```
KrivorotkoYurii, line 90: It will be better if you move ur form to a separate component

KrivorotkoYurii (changes requested):
Good job! Just check the comments below

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,69 @@
...
+          id="title"
+          data-cy="titleInput"
+          value={title}
+          onChange={onTitleChange}
+        />
+        {errorTitle && <span className="error">{errorTitle}</span>}
+      </div>
+
+      <div className="field">
+        <label className="label" htmlFor={'title'}>
```
KrivorotkoYurii, line 42: If you add a label to your select form , htmlFor prop should contain relevant path .
```suggestion
        <label className="label" htmlFor={'user'}>
```

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,69 @@
...
+          User:
+        </label>
+        <select
+          data-cy="userSelect"
+          required
```
KrivorotkoYurii, line 47: If you added label with htmlFor path  you also you also need to add an id property to select with the same value as you set in label htmlFor
```suggestion
          required
          id='user'
```

KrivorotkoYurii (approved):
Nicely done!)

src/App.tsx
BudnikOleksii: > You use usersFromServer.find(...) at leat twice (check the line 66 in this file). It is a dublicating of code. You can create a new folder 'utilities', create a file like 'CreateUser' and write a function with this logic. And after that import this function at the line 12 and 66

Good point
Also, `prepareTodos` can be just a variable outside the component as you need to do it once

src/App.tsx
```diff
@@ -1,61 +1,104 @@
...
+    setErrorUserId('');
+  };
+
+  const handleSubmit = (event: FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (!title || userId === 0) {
+      if (!title) {
+        setErrorTitle('Please enter a title');
+      }
```
BudnikOleksii, line 58: Avoid nested if/else statements

src/components/TodoForm/TodoForm.tsx
BudnikOleksii: > If you add a label to your select form , htmlFor prop should contain relevant path .

Good point

src/components/TodoForm/TodoForm.tsx
BudnikOleksii: > If you added label with htmlFor path you also you also need to add an id property to select with the same value as you set in label htmlFor

Good point

src/types/Todo.ts
BudnikOleksii: > You can also use '?' instead of null. If you implement it you can also chenge the conditions in other files by removing ' | | null '

It's not necessary

BudnikOleksii (changes requested):
Good work 👍 
Check my and the previous reviewer comments

KosolapovAnatolii (approved):
Good job!

PR #2773, opened 2024-10-06
Some tests are not passed, but looking good

Pidsvynok:
https://Pidsvynok.github.io/react_add-todo-form/

### DIFF
--- a/src/components/UserInfo/UserInfo.tsx
+++ b/src/components/UserInfo/UserInfo.tsx
@@ -1,1 +1,15 @@
-export const UserInfo = () => {};
+import React from 'react';
+
+import { User } from '../../types/User';
+
+interface Props {
+  user: User;
+}
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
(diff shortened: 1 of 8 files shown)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,15 @@
...
+
+import { User } from '../../types/User';
+
+interface Props {
+  user: User;
+}
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  return (
+    <a className="UserInfo" href="mailto:Sincere@april.biz">
```
volodymyr-soltys97, line 11: Fix href attribute value

volodymyr-soltys97 (changes requested):
Hi, you need to add placeholder text for the input and fix failed tests, if you need some help feel free in the fe_chat 
image

Also, the title cannot be only spaces 
image

anastasiiavorobiova (approved):
Great work!

PR #2778, opened 2024-10-07
solution

EustasTheMonk:
DEMO LINK (https://eustasthemonk.github.io/react_add-todo-form/)


Sukulynt (approved):
Well Done!

src/App.tsx
```diff
@@ -1,61 +1,188 @@
...
+import React, { useState } from 'react';
+import { Todo } from './Types/todo';
+import { FormFields } from './Types/formFields';
+
+const formFields: FormFields = {
+  title: '',
+  user: '0',
+};
+
+const findUserByUserId = (userId: number): User | null => {
```
VolodymyrKirichenko, line 17: move it to a separate `utils` directory

src/App.tsx
```diff
@@ -1,61 +1,188 @@
...
+  return usersFromServer.find((user: User) => user.id === userId) || null;
+};
+
+const makeNewTodo = (
```
VolodymyrKirichenko, line 21: It is important to follow good practice and use the `handle` function
```suggestion
const handleAddNewTodo = (
```

src/App.tsx
```diff
@@ -1,61 +1,188 @@
...
+
+const preparedTodos = todosFromServer.map(todo => {
+  return {
+    ...todo,
+    user: findUserByUserId(todo.userId),
+  };
+});
+
+const formFieldsErrors = () => {
+  const fieldsErrors: { [key: string]: boolean } = {};
```
VolodymyrKirichenko, line 45: use Type for it
```suggestion
  const fieldsErrors: Ur Type = {};
```

src/App.tsx
```diff
@@ -1,61 +1,188 @@
...
+
+  const handleAddTodo = (
+    event: React.FormEvent<HTMLFormElement>,
+    todo: FormFields,
+  ): void => {
+    event.preventDefault();
+    setIsTouched(true);
+    handleValidation(newTodoForm);
+
+    if (mistakesInForm) {
```
VolodymyrKirichenko, line 110: if the function or variable returns a boolean value - use prefixes (`is` or `has`)
```suggestion
    if (isMistakesInForm) {
```

src/App.tsx
```diff
@@ -1,61 +1,188 @@
...
-          <span className="error">Please enter a title</span>
+          <label>
+            <span>Title: </span>
+            <input
+              type="text"
+              data-cy="titleInput"
+              placeholder="Enter a title"
+              value={newTodoForm.title}
+              onChange={changeEvent =>
+                handleFormChange('title', changeEvent.target.value.trimStart())
```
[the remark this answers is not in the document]
VolodymyrKirichenko: Good point 🔥 
Do it!

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,21 @@
...
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  return (
+    <article
+      data-id={todo.id}
+      className={cn('TodoInfo', { 'TodoInfo--completed': todo.completed })}
+    >
+      <h2 className="TodoInfo__title">{todo.title}</h2>
+
+      {todo.user ? <UserInfo user={todo.user} /> : <span>User not found</span>}
```
VolodymyrKirichenko, line 18: will be more readable
```suggestion
      {todo.user ? (
        <UserInfo user={todo.user} />
      ) : (
        <span>User not found</span>
      )}
```

src/App.tsx
```diff
@@ -1,61 +1,188 @@
...
+  return {
+    id: id,
+    title: title,
+    completed: completed,
+    userId: userId,
+    user: user,
+  };
+};
+
+const preparedTodos = todosFromServer.map(todo => {
```
VolodymyrKirichenko, line 37: add type for or variable
```suggestion
const preparedTodos: Todo[] = todosFromServer.map(todo => {
```

src/App.tsx
```diff
@@ -1,61 +1,188 @@
...
+          return {
+            ...prevState,
+            [key]: true,
+          };
+        });
+      }
+    }
+  };
+
+  const handleErrorAppear = (isIncorrectField: boolean) => {
```
VolodymyrKirichenko, line 75: `handle` or `check`?
I think ur function doesn't do anything, so this will be more descriptive
```suggestion
  const checkErrorAppearance = (isIncorrectField: boolean) => {
```

src/components/TodoInfo/TodoInfo.tsx
EustasTheMonk: > will be more readable

I tried to do this, but the linter didn’t let me do it.

src/App.tsx
```diff
@@ -1,61 +1,186 @@
...
+const handleAddNewTodo = (
+  id: number,
+  title: string,
+  userId: number,
+  user: User | null,
+  completed: boolean = false,
+) => {
+  return {
+    id: id,
+    title: title.trim(),
```
BudnikOleksii, line 28: Only what this function does trim the title and it doesn't add todo, so the name is incorrect

src/App.tsx
```diff
@@ -1,61 +1,186 @@
...
+import { TodoList } from './components/TodoList';
+import { Todo } from './Types/todo';
+import { FormFields } from './Types/formFields';
+import { findUserByUserId } from './utils/fundUserByUserId';
+import { ErrorsFields } from './Types/errorsFields';
+
+const formFields: FormFields = {
+  title: '',
+  selectedUserId: '0',
+};
```
BudnikOleksii, line 17: - Store information about errors in the component, in the state.
- And actually, you don't show an error message if there is no user selected
image

src/App.tsx
```diff
@@ -1,61 +1,186 @@
...
+    handleClearForm();
+    setIsTouched(false);
+  };
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
-
-      <form action="/api/todos" method="POST">
+      <form onSubmit={submitEvent => handleAddTodo(submitEvent, newTodoForm)}>
```
BudnikOleksii, line 132:
```suggestion
      <form onSubmit={handleAddTodo}>
```
You can simplify as info about new todo you get from the state

src/Types/formFields.ts
```diff
@@ -0,0 +1,5 @@
+export interface FormFields {
+  [key: string]: string;
```
BudnikOleksii, line 2: I think you have only a title and selectedUserId fields, not a random strings and the same with `ErrorsFields`

BudnikOleksii (changes requested):
Good work 👍 
Just a few moments should be fixed

PR #2781, opened 2024-10-07
solution

mbulchak:
[DEMO LINK](https://mbulchak.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,14 @@
 import './App.scss';
 
+import { useState } from 'react';
 import todosFromServer from './api/todos';
-
-import { Todo } from './types/Todo';
 
 import { TodoList } from './components/TodoList';
 import { AddTodoForm } from './components/AddTodoForm';
 
+import { Todo } from './types/Todo';
+
 import { getUserById } from './utils/getUserById';
-import { useState } from 'react';
 
 const neededTodos = todosFromServer.map((todo: Todo) => {
   return {
@@ -17,29 +17,18 @@ const neededTodos = todosFromServer.map((todo: Todo) => {
   };
 });
 
-function getNewTodoId(todos: Todo[]) {
-  const maxId = Math.max(...todos.map(todo => todo.id));
-
-  return maxId + 1;
-}
-
 export const App = () => {
   const [todos, setTodos] = useState<Todo[]>(neededTodos);
 
   const addTodo = (todo: Todo) => {
-    const newTodo = {
-      ...todo,
-      id: getNewTodoId(todos),
-    };
-
-    setTodos(currentTodos => [...currentTodos, newTodo]);
+    setTodos(currentTodos => [...currentTodos, todo]);
   };
 
   return (
     <div className="App">
       <h1>Add todo form</h1>
 
-      <AddTodoForm onAdd={addTodo} />
+      <AddTodoForm onAdd={addTodo} todos={todos} />
 
       <TodoList todos={todos} />
     </div>
--- a/src/components/AddTodoForm/AddTodoForm.tsx
+++ b/src/components/AddTodoForm/AddTodoForm.tsx
@@ -2,53 +2,61 @@ import React, { useState } from 'react';
 import usersFromServer from './../../api/users';
 import { Todo } from '../../types/Todo';
 import { getUserById } from '../../utils/getUserById';
+import { getNewTodoId } from '../../utils/getNewTodoId';
 
 interface Props {
   onAdd: (todo: Todo) => void;
+  todos: Todo[];
 }
 
-export const AddTodoForm: React.FC<Props> = ({ onAdd }) => {
+enum TypeErrors {
+  Default = '',
+  TitleError = 'Please enter a title',
+  UserIdError = 'Please choose a user',
+}
+
+export const AddTodoForm: React.FC<Props> = ({ onAdd, todos }) => {
   const [title, setTitle] = useState('');
-  const [titleError, setTitleError] = useState('');
+  const [titleError, setTitleError] = useState(TypeErrors.Default);
 
   const [selectedUserId, setSelectedUserId] = useState(0);
-  const [userIdError, setUserIdError] = useState('');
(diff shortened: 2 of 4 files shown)


src/App.tsx
```diff
@@ -1,61 +1,47 @@
...
+
+const neededTodos = todosFromServer.map((todo: Todo) => {
+  return {
+    ...todo,
+    user: getUserById(todo.userId),
+  };
+});
+
+function getNewTodoId(todos: Todo[]) {
+  const maxId = Math.max(...todos.map(todo => todo.id));
```
vikshiyan, line 21: Suggestion: This function you can move to another file, for example to utils

src/components/AddTodoForm/AddTodoForm.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { Todo } from '../../types/Todo';
+import { getUserById } from '../../utils/getUserById';
+
+interface Props {
+  onAdd: (todo: Todo) => void;
+}
+
+export const AddTodoForm: React.FC<Props> = ({ onAdd }) => {
+  const [title, setTitle] = useState('');
+  const [titleError, setTitleError] = useState('');
```
vikshiyan, line 12: Suggestion: You can use enum

src/components/AddTodoForm/AddTodoForm.tsx
```diff
@@ -0,0 +1,110 @@
...
+    if (selectedUserId === 0) {
+      setUserIdError('Please choose a user');
+    }
+
+    if (!title || selectedUserId === 0) {
+      return;
+    }
+
+    const newTodo: Todo = {
+      id: 0,
```
vikshiyan, line 51: CHECKLIST: `id` is the largest `id` in the array + 1

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,23 @@
...
+import cn from 'classnames';
+
+interface Props {
+  todo: Todo;
+}
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  return (
+    <article
+      key={todo.id}
```
vikshiyan, line 12: Recomended: `key={todo.id}`  I think this key is not needed because it is only wrote in TodoList when wrote map

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,23 @@
...
+      data-id={todo.id}
+      className={cn('TodoInfo', {
+        'TodoInfo--completed': todo.completed === true,
```
vikshiyan, line 15: Suggestion: ` todo.completed === true`, you can use only `todo.completed`

src/utils/getUserById.ts
```diff
@@ -0,0 +1,6 @@
+import { User } from '../types/User';
+import usersFromServer from './../api/users';
+
+export const getUserById = (userId: number): User | undefined => {
```
vikshiyan, line 4: Suggestion: I think `User | undefined` will be better use `User | null` and in return you can add `|| null`

vikshiyan (changes requested):
Great job) but you have small mistake from checklist. And check your demo, because it doesn't work. 
image

vikshiyan (changes requested):
Please fix mistake)

src/utils/getUserById.ts
mbulchak: When I write null, I get a bunch of errors, which saying "Type 'null' is not assignable to type 'User | undefined'", because in todos there are no users. So I must write undefined. But if there are users set to null, of course I type the return to User | null

src/App.tsx
```diff
@@ -1,61 +1,36 @@
...
+import todosFromServer from './api/todos';
+
+import { TodoList } from './components/TodoList';
+import { AddTodoForm } from './components/AddTodoForm';
+
+import { Todo } from './types/Todo';
+
+import { getUserById } from './utils/getUserById';
+
+const neededTodos = todosFromServer.map((todo: Todo) => {
```
VolodymyrKirichenko, line 13: will better. And u can move this function to separate directory - `utils`
```suggestion
const preparedTodos/visibleTodos = todosFromServer.map((todo: Todo) => {
```

src/App.tsx
```diff
@@ -1,61 +1,36 @@
...
+  return {
+    ...todo,
+    user: getUserById(todo.userId),
+  };
+});
 
 export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>(neededTodos);
+
+  const addTodo = (todo: Todo) => {
```
VolodymyrKirichenko, line 23:
```suggestion
  const handleAddTodo = (todo: Todo) => {
```

src/components/AddTodoForm/AddTodoForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+        <input
+          id="title"
+          type="text"
+          placeholder="Enter a title"
+          data-cy="titleInput"
+          value={title}
+          onChange={handleChangeTitle}
+        />
+
+        {titleError && <span className="error">{TypeErrors.TitleError}</span>}
```
VolodymyrKirichenko, line 85: it not `falsy` value. Use `length` for it. In all such cases
```suggestion
        {titleError.length && <span className="error">{TypeErrors.TitleError}</span>}
```

src/components/AddTodoForm/AddTodoForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { Todo } from '../../types/Todo';
+import { getUserById } from '../../utils/getUserById';
+import { getNewTodoId } from '../../utils/getNewTodoId';
+
+interface Props {
+  onAdd: (todo: Todo) => void;
+  todos: Todo[];
+}
+
+enum TypeErrors {
```
VolodymyrKirichenko, line 12: move it to separate file - `models/enum/typeErrors.ts`

VolodymyrKirichenko (approved):
Good job 🔥 
But let's do some fixes

PR #2807, opened 2024-10-31
Solution

ab3MN:
[DEMO LINK](https://ab3MN.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,24 @@
+import { useState } from 'react';
 import './App.scss';
+import { TodoList } from './components/TodoList';
+import TodosForm from './components/TodosForm/TodosForm';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import todosFromServer from './api/todos';
+
+import { ITodo } from './types/todo';
+import todosWithUser from './utils/todo/todosWithUser';
 
 export const App = () => {
+  const [todos, setTodos] = useState<ITodo[]>([...todosFromServer]);
+
+  const addTodo = (todo: ITodo) => setTodos(state => [...state, todo]);
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
-
-      <form action="/api/todos" method="POST">
+      addTodo
+      <TodosForm todos={todos} addTodo={addTodo} />
+      {/* <form action="/api/todos" method="POST">
         <div className="field">
           <input type="text" data-cy="titleInput" />
           <span className="error">Please enter a title</span>
@@ -27,9 +37,9 @@ export const App = () => {
         <button type="submit" data-cy="submitButton">
           Add
         </button>
-      </form>
-
-      <section className="TodoList">
+      </form> */}
+      <TodoList todos={todosWithUser(todos)} />
+      {/* <section className="TodoList">
         <article data-id="1" className="TodoInfo TodoInfo--completed">
           <h2 className="TodoInfo__title">delectus aut autem</h2>
 
@@ -55,7 +65,7 @@ export const App = () => {
             Patricia Lebsack
           </a>
         </article>
-      </section>
+      </section> */}
     </div>
   );
 };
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,14 +1,24 @@
...
-// import todosFromServer from './api/todos';
+import todosFromServer from './api/todos';
+
+import { ITodo } from './types/todo';
+import todosWithUser from './utils/todo/todosWithUser';
 
 export const App = () => {
+  const [todos, setTodos] = useState<ITodo[]>([...todosFromServer]);
+
+  const addTodo = (todo: ITodo) => setTodos(state => [...state, todo]);
```
BudnikOleksii, line 14: Looks like this function is redundant

src/App.tsx
```diff
@@ -1,14 +1,24 @@
...
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
-
-      <form action="/api/todos" method="POST">
+      addTodo
+      <TodosForm todos={todos} addTodo={addTodo} />
+      {/* <form action="/api/todos" method="POST">
```
BudnikOleksii, line 21: Remove all comments

BudnikOleksii (changes requested):
Good work 👍 
Let's fix a bug, even a user did nothing they see errors, check the working version for reference:
image

src/components/TodosForm/TodosForm.tsx
```diff
@@ -0,0 +1,102 @@
...
+  };
+
+  const checkTodoFormState = () => {
+    if (!title) {
+      setTitle(null);
+    }
+    if (!userId) {
+      setUserId(null);
+    }
+  };
```
BudnikOleksii, line 34: Why? It doesn't make sense

src/components/TodosForm/TodosForm.tsx
```diff
@@ -0,0 +1,102 @@
...
+    setUserId(+e.target.value);
+
+  return (
+    <form action="/api/todos.ts" method="POST" onSubmit={handleSubmit}>
+      <div className="field">
+        <input
+          type="text"
+          data-cy="titleInput"
+          name="title"
+          value={title !== null ? title : ''}
```
BudnikOleksii, line 67: I have no idea why you do this. Also, if you can use [Nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) operator

src/components/TodosForm/TodosForm.tsx
```diff
@@ -0,0 +1,102 @@
...
+          onChange={handleChangeTitle}
+          placeholder="Enter the title"
+        />
+
+        {title === null && <span className="error">Please enter a title</span>}
```
BudnikOleksii, line 72: Okay, I think I get it. If you do it to handle errors, you can just add a state for the errors and depending on it show an error message. Because it looks strange

BudnikOleksii (changes requested):
Just create a state for errors to show some messages in case of errors.
And maybe you forget to deploy as actually nothing changed and I see errors even I do nothing

src/components/TodosForm/TodosForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+  );
+
+  return (
+    <form action="/api/todos.ts" method="POST" onSubmit={handleSubmit}>
+      <div className="field">
+        <input
+          type="text"
+          data-cy="titleInput"
+          name="title"
+          value={title !== null ? title : ''}
```
BudnikOleksii, line 85:
```suggestion
          value={title}
```
You don't need it now

src/components/TodosForm/TodosForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+        />
+
+        {titleError && <span className="error">Please enter a title</span>}
+      </div>
+
+      <div className="field">
+        <select
+          data-cy="userSelect"
+          name="userId"
+          value={userId !== null ? userId : 0}
```
BudnikOleksii, line 97:
```suggestion
          value={userId}
```

BudnikOleksii (approved):
LGTM 🚀

PR #2856, opened 2024-12-05
solve

Daniil-102:
[DEMO LINK](https://Daniil-102.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,113 @@
 import './App.scss';
+import React, { FormEvent, useState } from 'react';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { Todo } from './types/Todo.type';
+import { TodoList } from './components/TodoList';
 
-export const App = () => {
+const todosWithUser: Todo[] = todosFromServer.map(todo => {
+  const curUser = usersFromServer.find(user => todo.userId === user.id);
+
+  return {
+    ...todo,
+    user: curUser as Todo['user'],
+  };
+});
+
+export const App: React.FC = () => {
+  const [input, setInput] = useState('');
+  const [select, setSelect] = useState('0');
+  const [errors, setErrors] = useState<{ [key: string]: string }>({});
+  const [todos, setTodos] = useState<Todo[]>(todosWithUser || []);
+
+  const validateForm = () => {
+    const newErrors: { [key: string]: string } = {};
+
+    if (!input.trim()) {
+      newErrors.input = 'Please enter a title';
+    }
+
+    if (select === '0') {
+      newErrors.select = 'Please choose a user';
+    }
+
+    return newErrors;
+  };
+
+  const handleSubmit = (e: FormEvent) => {
+    e.preventDefault();
+
+    const validationErrors = validateForm();
+
+    if (Object.keys(validationErrors).length > 0) {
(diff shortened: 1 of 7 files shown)


src/types/Todo.type.ts
```diff
@@ -0,0 +1,12 @@
+export interface Todo {
+  user: {
+    id: number;
+    name: string;
+    username: string;
+    email: string;
+  };
```
natalia646, line 7: for me, it's better to use an imported User type

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,18 @@
-export const UserInfo = () => {};
+import React from 'react';
+
+interface Props {
+  user: {
+    id: number;
+    name: string;
+    username: string;
+    email: string;
+  };
```
natalia646, line 9: import User and use in Props
`user: User`

natalia646 (approved):

src/App.tsx
```diff
@@ -1,61 +1,121 @@
...
+
+  return {
+    ...todo,
+    user: curUser as Todo['user'],
+  };
+});
+
+export const App: React.FC = () => {
+  const [input, setInput] = useState('');
+  const [select, setSelect] = useState('0');
```
BudnikOleksii, line 20: What input and select are they? Name should be descriptive

src/App.tsx
```diff
@@ -1,61 +1,121 @@
...
+  const [errors, setErrors] = useState<{ [key: string]: string }>({});
```
BudnikOleksii, line 21:
```suggestion
  type Field = 'title' | 'userId'

  const [errors, setErrors] = useState<Record<Field, string>>({});
```
Define what fields error state has for example with union or enum type

src/App.tsx
```diff
@@ -1,61 +1,121 @@
...
+  const [todos, setTodos] = useState<Todo[]>(todosWithUser || []);
+
+  const validateForm = () => {
+    const newErrors: { [key: string]: string } = {};
```
BudnikOleksii, line 25: This object should have the same type as a state, so create a type for them

src/App.tsx
```diff
@@ -1,61 +1,121 @@
...
+
+    const newTodo: Todo = {
+      id: maxId + 1,
+      title: input,
+      completed: false,
+      userId: +select,
+      user: usersFromServer.find(user => user.id === +select)!,
+    };
+
+    setTodos([...todos, newTodo]);
```
BudnikOleksii, line 59: When you set state depending on the previous one use functional `setState`

src/App.tsx
```diff
@@ -1,61 +1,121 @@
...
+    setInput('');
+    setSelect('0');
+    setErrors({});
```
BudnikOleksii, line 62: Create a resetting function

src/App.tsx
```diff
@@ -1,61 +1,121 @@
...
+          <label htmlFor="title">
+            {'Title: '}
+            <input
+              title="title"
+              type="text"
+              value={input}
+              onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
+                setInput(e.target.value);
+                setErrors(prev => ({ ...prev, input: '' }));
+              }}
```
BudnikOleksii, line 80: Create function that you pass to `onChange` instead of passing anonim functions

BudnikOleksii (changes requested):
Good work 👍
Let's improve your solution slightly

PR #2857, opened 2024-12-05
Solution

natalia646:
[DEMO](https://natalia646.github.io/react_add-todo-form/)

### DIFF
--- a/src/components/Form/Form.tsx
+++ b/src/components/Form/Form.tsx
@@ -1,0 +1,74 @@
+import { useState } from 'react';
+import { Todo, User } from '../../types';
+import { getNewId } from '../../utils/getNewId';
+
+import { InputTitle } from './InputTitle';
+import { SelectUser } from './SelectUser';
+
+type Props = {
+  users: User[];
+  todos: Todo[];
+  onAddTodo: (todo: Todo) => void;
+};
+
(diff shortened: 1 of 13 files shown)


src/components/Form/Form.tsx
```diff
@@ -0,0 +1,74 @@
...
+        completed: false,
+        userId,
+        user: users.find(user => userId === user.id),
+      };
+
+      onAddTodo(todo);
+      setTitle('');
+      setUserId(0);
+      setHasTitleError(false);
+      setHasUserIdError(false);
```
BudnikOleksii, line 50: Create a resetting function

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,22 @@
...
+  const { title, id, completed, user } = todo;
+
+  return (
+    <article
+      data-id={id}
+      className={cn('TodoInfo', { 'TodoInfo--completed': completed })}
+    >
+      <h2 className="TodoInfo__title">{title}</h2>
+
+      {user ? <UserInfo user={user} /> : <></>}
```
BudnikOleksii, line 19:
```suggestion
      {user && <UserInfo user={user} />}
```
You don't need to use the ternary operator if you return nothing in the second case

BudnikOleksii (approved):
LGTM 🚀

PR #2862, opened 2024-12-05
add task solution

Andrew77788:
[DEMO LINK](https://Andrew77788.github.io/react_add-todo-form/

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,1 +1,44 @@
-export const TodoInfo = () => {};
+import { Todo } from '../TodoList';
+import { User } from '../UserInfo';
+
+export const TodoInfo = ({
+  todos = [],
+  users,
+}: {
+  todos: Todo[];
+  users: User[];
+}) => {
+  const findUser = (userId: number) => {
+    const us = users.find(user => user.id === userId);
+
+    return (
+      us || {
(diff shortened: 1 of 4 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,44 @@
...
+    );
+  };
+
+  return (
+    <>
+      {todos.map(todo => (
+        <article
+          key={todo.id}
+          data-id={todo.id}
+          className={`TodoInfo ${todo.completed ? 'TodoInfo--completed' : ''}`}
```
volodymyr-soltys97, line 30: Use the classnames library for add classes with condition

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,28 @@
...
+        Choose a user
+      </option>
+      {users.map(user => (
+        <option key={user.id} value={user.id}>
+          {user.name}
+        </option>
+      ))}
+    </>
+  );
+};
```
volodymyr-soltys97, line 28: This component should only accept one user as props, not a whole list.

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
1. You need to fix these failed tests, if you need help feel free in the fe_chat 
image

2. Add the label text for inputs 
image

PR #2864, opened 2024-12-06
Solution

save-v:
https://save-v.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,114 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoList } from './components/TodoList';
+import React, { useState } from 'react';
+import { Todo } from './types/Todo';
+
+function findUserById(userId: number) {
+  return usersFromServer.find(user => user.id === userId);
+}
+
+const todos = todosFromServer.map(todo => {
+  const user = findUserById(todo.userId);
+
+  return { ...todo, user };
+});
 
 export const App = () => {
+  const [selectedUser, setSelectedUser] = useState(0);
+  const [title, setTitle] = useState('');
+  const [updatedTodos, setUpdatedTodos] = useState<Todo[]>(todos);
+  const [titleError, setTitleError] = useState(false);
+  const [selectedUserError, setSelectedUserError] = useState(false);
+
+  // <Todo[]>
+  function handleSubmit(event: React.FormEvent) {
+    event.preventDefault();
+
+    if (!selectedUser || !title) {
+      if (!selectedUser) {
+        setSelectedUserError(true);
+      }
+
+      if (!title) {
+        setTitleError(true);
+      }
+
+      return;
+    }
+
+    const newTodo: Todo = {
+      id: Math.max(...updatedTodos.map(todo => todo.id)) + 1,
+      title: title,
+      completed: false,
+      userId: selectedUser,
+      user: findUserById(selectedUser),
+    };
+
+    setUpdatedTodos([...updatedTodos, newTodo]);
+
+    setSelectedUser(0);
+    setTitle('');
+  }
+
   return (
     <div className="App">
       <h1>Add todo form</h1>
-
-      <form action="/api/todos" method="POST">
+      <form action="/api/todos" method="POST" onSubmit={handleSubmit}>
         <div className="field">
-          <input type="text" data-cy="titleInput" />
-          <span className="error">Please enter a title</span>
+          <label htmlFor="todo-title">
+            {'Title: '}
+            <input
+              placeholder="Enter a title"
+              type="text"
+              data-cy="titleInput"
+              id="todo-title"
+              value={title}
+              onChange={event => {
+                setTitle(event.target.value);
+                setTitleError(false);
+              }}
+            />
+          </label>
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,114 @@
...
-          <span className="error">Please enter a title</span>
+          <label htmlFor="todo-title">
+            {'Title: '}
+            <input
+              placeholder="Enter a title"
+              type="text"
+              data-cy="titleInput"
+              id="todo-title"
+              value={title}
+              onChange={event => {
```
pogorielova, line 69: Would be better to ceparate this function, as it have more than one action

src/App.tsx
```diff
@@ -1,61 +1,114 @@
...
-          </select>
-
-          <span className="error">Please choose a user</span>
+          <label htmlFor="user-select">
+            {'User: '}
+            <select
+              data-cy="userSelect"
+              value={selectedUser}
+              onChange={event => {
+                setSelectedUser(+event.target.value);
```
pogorielova, line 85: Would be better to ceparate this function, as it have more than one action

src/App.tsx
```diff
@@ -1,61 +1,114 @@
...
+  const user = findUserById(todo.userId);
+
+  return { ...todo, user };
+});
 
 export const App = () => {
+  const [selectedUser, setSelectedUser] = useState(0);
+  const [title, setTitle] = useState('');
+  const [updatedTodos, setUpdatedTodos] = useState<Todo[]>(todos);
+  const [titleError, setTitleError] = useState(false);
```
pogorielova, line 23: Let me be a nerd, as Vitalii Schukin  would say)) It's better to insert 'is' or 'has' to the title of the state if it is a boolean

pogorielova (approved):
Great work!

src/App.tsx
Moroz-Dmytro: Good point, fix this

src/App.tsx
```diff
@@ -1,61 +1,114 @@
...
+  const [selectedUserError, setSelectedUserError] = useState(false);
```
Moroz-Dmytro, line 24: The same as above

src/App.tsx
```diff
@@ -1,61 +1,114 @@
...
+
+  // <Todo[]>
```
Moroz-Dmytro, line 26: Don't leave comments

src/App.tsx
```diff
@@ -1,61 +1,114 @@
...
+      if (!selectedUser) {
+        setSelectedUserError(true);
+      }
+
+      if (!title) {
+        setTitleError(true);
+      }
+
+      return;
+    }
```
Moroz-Dmytro, line 40:
```suggestion
      if (!selectedUser) {
        setSelectedUserError(true);
        return;
      }

      if (!title) {
        setTitleError(true);
        return;
      }
```

src/App.tsx
```diff
@@ -1,61 +1,114 @@
...
+
+    const newTodo: Todo = {
+      id: Math.max(...updatedTodos.map(todo => todo.id)) + 1,
```
Moroz-Dmytro, line 43: Create separate function for generating new id

src/App.tsx
```diff
@@ -1,61 +1,114 @@
...
+      title: title,
```
Moroz-Dmytro, line 44:
```suggestion
      title,
```

src/App.tsx
```diff
@@ -1,61 +1,114 @@
...
+      completed: false,
+      userId: selectedUser,
+      user: findUserById(selectedUser),
+    };
+
+    setUpdatedTodos([...updatedTodos, newTodo]);
```
Moroz-Dmytro, line 50:
```suggestion
    setUpdatedTodos(prevTodos => [...prevTodos, newTodo]);
```

src/App.tsx
Moroz-Dmytro: Good point, fix this

src/App.tsx
Moroz-Dmytro: Good point, fix this

src/App.tsx
```diff
@@ -1,61 +1,114 @@
...
+              <option value="0" disabled>
+                Choose a user
+              </option>
+              {usersFromServer.map(user => {
+                return (
+                  <option key={user.id} value={user.id}>
+                    {user.name}
+                  </option>
+                );
+              })}
```
Moroz-Dmytro, line 99:
```suggestion
              {usersFromServer.map(user => (
                <option key={user.id} value={user.id}>
                  {user.name}
                </option>
              ))}
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,15 @@
...
+
+export const TodoInfo: React.FC<{ todo: Todo }> = ({ todo }) => (
+  <article
+    data-id={todo.id}
+    className={classNames('TodoInfo', {
+      'TodoInfo--completed': todo.completed,
+    })}
+  >
+    <h2 className="TodoInfo__title">{todo.title}</h2>
+    <UserInfo user={todo.user} />
```
Moroz-Dmytro, line 13: Use destructuring

Moroz-Dmytro (changes requested):
Good progress 🚀 
But fix some comments below

StasSokolov1 (approved):
Looks good to me, I'm approving it!

PR #2883, opened 2024-12-10
Solution

andrushchenkoo:
[DEMO LINK](https://andrushchenkoo.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,141 @@
 import './App.scss';
+import { TodoList } from './components/TodoList';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { useState } from 'react';
+
+export type User = {
+  id: number;
+  name: string;
+  username: string;
+  email: string;
+};
+
+export type Todo = {
+  id: number;
+  title: string;
+  completed: boolean;
+  userId: number;
+  user?: User | null;
+};
+
+const getUserById = (userId: number): User | null => {
+  return (
+    usersFromServer.find(userFromServer => userId === userFromServer.id) || null
+  );
+};
+
+export const todos: Todo[] = todosFromServer.map(todo => {
+  return { ...todo, user: getUserById(todo.userId) };
+});
 
 export const App = () => {
+  const [visibleTodos, setVisibleTodos] = useState(todos);
+  const [title, setTitle] = useState('');
+  const [selectedUserId, setSelectedUserId] = useState('0');
+  const [titleError, setTitleError] = useState(false);
+  const [userError, setUserError] = useState(false);
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    let isValid = true;
+
+    if (title === '') {
+      setTitleError(true);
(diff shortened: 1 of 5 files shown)


RuslanMudryi (approved):
good job

src/App.tsx
```diff
@@ -1,61 +1,149 @@
...
+    } else {
+      setTitleError(false);
+    }
+
+    if (selectedUserId === '0') {
+      setUserError(true);
+      isValid = false;
+    } else {
+      setUserError(false);
+    }
```
BudnikOleksii, line 57: Try to simplify.
- You don't need to reset errors, you do it when a user types something in related input
- Prefer const over let
- You don't need to compare with an empty string, just check if a value exists

src/App.tsx
```diff
@@ -1,61 +1,149 @@
...
+
+    if (!isValid) {
+      return;
+    }
+
+    const getId = () => {
```
BudnikOleksii, line 63: Move this function outside `handleSubmit`

src/App.tsx
```diff
@@ -1,61 +1,149 @@
...
+      id: getId(),
+      title: title,
+      completed: false,
+      userId: +selectedUserId,
+      user: getUserById(parseInt(selectedUserId)) || null,
+    };
+
+    setVisibleTodos(prevTodos => [...prevTodos, newTodo]);
+    setTitle('');
+    setSelectedUserId('0');
```
BudnikOleksii, line 81: Create a function to reset values

src/App.tsx
```diff
@@ -1,61 +1,149 @@
...
+              type="text"
+              value={title}
+              data-cy="titleInput"
+              placeholder="Enter a title"
+              onChange={event => {
+                setTitle(event.target.value);
+                if (event.target.value !== '') {
+                  setTitleError(false);
+                }
+              }}
```
BudnikOleksii, line 103: Create a function instead of passing anonim one to `onChange`

src/App.tsx
```diff
@@ -1,61 +1,149 @@
...
+export const todos: Todo[] = todosFromServer.map(todo => {
+  return { ...todo, user: getUserById(todo.userId) };
+});
 
 export const App = () => {
+  const [visibleTodos, setVisibleTodos] = useState(todos);
+  const [title, setTitle] = useState('');
+  const [selectedUserId, setSelectedUserId] = useState('0');
+  const [titleError, setTitleError] = useState(false);
+  const [userError, setUserError] = useState(false);
```
BudnikOleksii, line 38: Boolean variables should start from `is` or `has`

src/App.tsx
```diff
@@ -1,61 +1,149 @@
...
+              placeholder="Enter a title"
+              onChange={event => {
+                setTitle(event.target.value);
+                if (event.target.value !== '') {
+                  setTitleError(false);
+                }
+              }}
+            />
+          </label>
+          {titleError === false ? null : (
```
BudnikOleksii, line 106:
```suggestion
          {titleError && (
```
You don't need to use the ternary operator if you return nothing in the second case

BudnikOleksii (changes requested):
Good work 👍 
Left a few comments to improve your solution.

PR #2894, opened 2024-12-15
Task solution (read description)

AnderZerfall:
My tests are completely broken because of a mysterious issue — I'm still getting the error "Failed to fetch the module...". So, idk, if it passes the test or not. I compared my code to others and it should work fine... 
At least the demo seems to be fine.

[DEMO LINK](https://anderzerfall.github.io/react_add-todo-form/)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
       },
       "devDependencies": {
         "@cypress/react18": "^2.0.1",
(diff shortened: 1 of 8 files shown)


AnderZerfall:
Some checks were not successful. I'll fix that asap

BudnikOleksii (approved):
Good work 👍 
But pay attention to naming

PR #2930, opened 2025-01-17
Develop

djtyrf312:
[DEMO LINK](https://djtyrf312.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -16,40 +16,42 @@ const todosWithUser: TodoWithUser[] = todosFromServer.map(
     };
   },
 );
-const defaultOptionIndex = '0';
+const defaultUserId = '0';
 
 export const App = () => {
   const [todos, setTodos] = useState<TodoWithUser[]>([...todosWithUser]);
   const [title, setTitle] = useState('');
-  const [userName, setUserName] = useState(defaultOptionIndex);
+  const [userId, setUserId] = useState(defaultUserId);
   const [hasTitleError, setHasTitleError] = useState(false);
   const [hasUserError, setHasUserError] = useState(false);
 
   const clearForm = () => {
     setTitle('');
-    setUserName(defaultOptionIndex);
+    setUserId(defaultUserId);
   };
 
-  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+  const handleSubmit = (
+    event: React.FormEvent<HTMLFormElement>,
+  ): null | void => {
     event.preventDefault();
     if (!title) {
       setHasTitleError(true);
     }
 
-    if (userName === defaultOptionIndex) {
+    if (userId === defaultUserId) {
       setHasUserError(true);
     }
 
-    if (title && userName !== defaultOptionIndex) {
+    if (title || userId !== defaultUserId) {
+      const user = usersFromServer.find(({ id }) => id === +userId) as User;
+      const newId = Math.max(...todos.map(todo => todo.id)) + 1;
+
       const newTodo: TodoWithUser = {
-        userId: usersFromServer.find(user => user.name === userName)
-          ?.id as number,
-        id: Math.max(...todos.map(todo => todo.id)) + 1,
+        userId: +userId,
+        id: newId,
         title,
         completed: false,
-        user: usersFromServer.find(
-          user => user.id === Number(userName),
-        ) as User,
+        user,
       };
 
       setTodos([...todos, newTodo]);
@@ -85,13 +87,13 @@ export const App = () => {
         <div className="field">
           <select
             data-cy="userSelect"
-            value={userName}
+            value={userId}
             onChange={event => {
-              setUserName(event.target.value);
+              setUserId(event.target.value);
               setHasUserError(false);
             }}
           >
-            <option value={defaultOptionIndex} disabled>
+            <option value={defaultUserId} disabled>
               Choose a user
             </option>
             {usersFromServer.map(user => {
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,61 +1,127 @@
...
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { TodoWithUser, User } from './App.types';
+import { TodoList } from './components/TodoList';
+
+const todosWithUser: TodoWithUser[] = todosFromServer.map(
+  (todo): TodoWithUser => {
+    const user = usersFromServer.find(({ id }) => id === todo.userId) as User;
```
naziamoff: I understand that this is for TS purposes. 
But what if the user with the provided id is not found? Shit like that sometimes happens and we want to handle it instead of casting types

src/App.tsx
```diff
@@ -1,61 +1,127 @@
...
+
+const validateTitle = (title: string): boolean => titleRegexp.test(title);
 
 export const App = () => {
+  const [todos, setTodos] = useState<TodoWithUser[]>([...todosWithUser]);
+  const [title, setTitle] = useState('');
+  const [userId, setUserId] = useState(defaultUserId);
+  const [hasTitleError, setHasTitleError] = useState(false);
+  const [hasUserError, setHasUserError] = useState(false);
+  const isNotSelectedUser = userId === defaultUserId;
```
naziamoff: when we're dealing with booleans, we usually make the logic from positive, e.g. `isSelectedUser`, and use with `NOT (!)` operator.

It is difficult to read stuff like `!isNotSelectedUser`, since I have to compile in my mind what this variable means. In my mind it would sound like `not is not selected user`, this is difficult to read

src/App.tsx
```diff
@@ -1,61 +1,127 @@
...
+  const hasForbiddenSymbolsInTitle = !validateTitle(title);
+  const titleErrorMessage = hasForbiddenSymbolsInTitle
```
naziamoff: from this thing, it looks like there must be some error message anyway. Would be better to have it as a function, which can also return null if there's no error

src/App.tsx
```diff
@@ -1,61 +1,127 @@
...
+      setHasTitleError(true);
+    }
+
+    if (isNotSelectedUser) {
+      setHasUserError(true);
+    }
+
+    if (title && !isNotSelectedUser) {
+      const user = usersFromServer.find(({ id }) => id === +userId) as User;
+      const newId = Math.max(...todos.map(todo => todo.id)) + 1;
```
naziamoff, line 70: you don't have to use spread operator here, since `map` is not mutable

src/App.tsx
```diff
@@ -1,61 +1,127 @@
...
+    if (!title || hasForbiddenSymbolsInTitle) {
+      setHasTitleError(true);
+    }
+
+    if (isNotSelectedUser) {
+      setHasUserError(true);
+    }
+
+    if (title && !isNotSelectedUser) {
+      const user = usersFromServer.find(({ id }) => id === +userId) as User;
```
naziamoff: converting to a number with `+` is not a mistake, but not considered a clean approach. To make it easier to understand we usually convert the type explicitly, like `Number(userId)`

src/App.tsx
```diff
@@ -1,61 +1,127 @@
...
+      const newId = Math.max(...todos.map(todo => todo.id)) + 1;
+
+      const newTodo: TodoWithUser = {
+        userId: +userId,
+        id: newId,
+        title,
+        completed: false,
```
naziamoff, line 76: `isCompleted` would be a cleaner approach

src/App.tsx
```diff
@@ -1,61 +1,127 @@
...
+            type="text"
+            value={title}
+            data-cy="titleInput"
+            placeholder="Please enter a title"
+            onChange={event => {
+              setTitle(event.target.value);
+              setHasTitleError(false);
+            }}
+          />
+          {hasTitleError && <span className="error">{titleErrorMessage}</span>}
```
naziamoff: this thing is difficult to read since there's a lot of stuff happening. There's a render condition, tag name, classname, text node.

Here's how I would approach it:

```suggestion
          {hasTitleError && (
            <span className="error">
              {titleErrorMessage}
            </span>
          )}
```

src/App.tsx
```diff
@@ -1,61 +1,127 @@
...
+            <option value={defaultUserId} disabled>
               Choose a user
             </option>
+            {usersFromServer.map(user => {
+              return (
+                <option key={user.id} value={user.id}>
+                  {user.name}
+                </option>
+              );
+            })}
```
naziamoff: We can return right away instead of return statement if there's no logic happening inside the callback

```suggestion
            {usersFromServer.map(user => (
              <option key={user.id} value={user.id}>
                {user.name}
              </option>
             )
            )}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,16 @@
-export const TodoList = () => {};
+import { TodoWithUser } from '../../App.types';
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: TodoWithUser[];
+};
+
+export const TodoList = ({ todos }: Props) => {
```
naziamoff: I mentioned `return-right-away` approach somewhere above, this component is a great place to practice it, since we have 2 return statement that can be avoided here

naziamoff (approved):

src/App.tsx
djtyrf312: Without this, I would have an error with the types. So maybe better to leave it to split the array to separate arguments for math max
image

PR #2948, opened 2025-01-27
init commit

nex1994:
[DEMO LINK](https://nex1994.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React, { FormEventHandler, useEffect, useState } from 'react';
+import React, { FormEventHandler, useState } from 'react';
 import './App.scss';
 
 import todosFromServer from './api/todos';
@@ -11,6 +11,10 @@ import { TodoList } from './components/TodoList';
 function getUserById(userId: number): User | null {
   const foundUser = usersFromServer.find(user => user.id === userId);
 
+  if (foundUser === null) {
+    throw new Error('No user found');
+  }
+
   // if there is no user with a given userId
   return foundUser || null;
 }
@@ -21,32 +25,18 @@ export const todos: Todo[] = todosFromServer.map(todo => ({
 }));
 
 export const App: React.FC = () => {
-  const [user, setUser] = useState('');
-  const [title, setTitle] = useState('');
-  const [userError, setUserError] = useState('');
-  const [titleError, setTitleError] = useState('');
+  const [userInput, setUserInput] = useState('');
+  const [titleInput, setTitleInput] = useState('');
+  const [userInputError, setUserInputError] = useState('');
+  const [titleInputError, setTitleInputError] = useState('');
   const [isSubmited, setIsSubmited] = useState(false);
-
-  useEffect(() => {
-    if (user === '') {
-      setUserError('Please choose a user');
-    } else {
-      setUserError('');
-    }
-
-    if (title === '') {
-      setTitleError('Please enter a title');
-    } else {
-      setTitleError('');
-    }
-
-    // return () => {
-    //   setIsSubmited(false);
-    // };
-  }, [user, title]);
 
   const getUserByName = (name: string): User | null => {
     const foundUser = usersFromServer.find(users => users.name === name);
+
+    if (foundUser === null) {
+      throw new Error('No user found');
+    }
 
     return foundUser || null;
   };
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+  ...todo,
+  user: getUserById(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [user, setUser] = useState('');
+  const [title, setTitle] = useState('');
+  const [userError, setUserError] = useState('');
+  const [titleError, setTitleError] = useState('');
+  const [isSubmited, setIsSubmited] = useState(false);
```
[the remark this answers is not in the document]
zibi-dev: True. You can remove useEffect and handle errors in the `handleSubmit `function.

src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+    (max, users) => (users.id > max ? users.id : max),
+    todos[0].id,
+  );
+
+  const handleSubmit: FormEventHandler = event => {
+    event.preventDefault();
+
+    setIsSubmited(true);
+
+    if (titleError !== '' || userError !== '') {
```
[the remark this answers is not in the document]
zibi-dev: +1

src/App.tsx
```diff
@@ -1,61 +1,134 @@
...
+  return foundUser || null;
+}
+
+export const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUserById(todo.userId),
+}));
+
+export const App: React.FC = () => {
+  const [user, setUser] = useState('');
```
zibi-dev: User more descriptive names. For example here it would good to name it `userNameInput` I would know right away what purpose is for it

src/App.tsx
```diff
@@ -1,61 +1,179 @@
...
+    event.preventDefault();
+
+    if (titleError !== '' || userError !== '') {
+      return;
+    }
+
+    todos.push({
+      title: title,
+      user: getUserByName(user),
+      userId: getUserByName(user)?.id,
```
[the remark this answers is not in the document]
zibi-dev: You can do it one go before push.

```
const submittedUser = getUserByName(user)

  todos.push({
      title: title,
      user: submittedUser,
      userId: submittedUser?.id,
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,16 @@
...
+
+export const TodoList = ({ todos }: Props) => {
+  return (
+    <section className="TodoList">
+      {todos.map(todo => {
+        return <TodoInfo key={todo.id} todo={todo} />;
+      })}
+    </section>
+  );
+};
```
zibi-dev:
```suggestion
export const TodoList = ({ todos }: Props) => {
  return (
    <section className="TodoList">
      {todos.map(todo =><TodoInfo key={todo.id} todo={todo} />;
      )}
    </section>
  );
};
```

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,13 @@
...
+  user: User | null;
+};
+
+export const UserInfo = ({ user }: Prop) => {
+  return (
+    <a className="UserInfo" href={`mailto:${user?.email}`}>
+      {user?.name}
+    </a>
+  );
+};
```
zibi-dev: This should handle if user is null. Maybe display "User not found"

zibi-dev (changes requested):
Good job with the implementation! However, there are a few areas that could be improved. Additionally, consider taking advantage of the opportunity to destructure some objects that are being passed as props.

src/App.tsx
```diff
@@ -61,20 +51,31 @@ export const App: React.FC = () => {
 
     setIsSubmited(true);
 
-    if (titleError !== '' || userError !== '') {
-      return;
+    if (userInput === '') {
+      setUserInputError('Please choose a user');
+    } else {
+      setUserInputError('');
     }
```
zibi-dev: Also text input should be trimmed. I can by pass this check by using empty spaces.

src/App.tsx
```diff
@@ -61,20 +51,31 @@ export const App: React.FC = () => {
...
-      user: getUserByName(user),
-      userId: getUserByName(user)?.id,
-      completed: false,
-      id: findHighestId + 1,
-    });
+    if (titleInput === '') {
+      setTitleInputError('Please enter a title');
+    } else {
+      setTitleInputError('');
+    }
```
zibi-dev: Those else block here and above can be removed errors are cleaned on every change in input and select elements

src/App.tsx
```diff
@@ -11,6 +11,10 @@ import { TodoList } from './components/TodoList';
 function getUserById(userId: number): User | null {
   const foundUser = usersFromServer.find(user => user.id === userId);
 
+  if (foundUser === null) {
+    throw new Error('No user found');
+  }
```
zibi-dev: Its not caught anywhere so it will break the app. You can remove it.

zibi-dev (changes requested):

zibi-dev (approved):

PR #3023, opened 2025-03-14
sol

dvdmsk:
- [DEMO LINK](https://dvdmsk.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -52,7 +52,7 @@ export const App = () => {
 
   const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
     event.preventDefault();
-    if (title && userSelect) {
+    if (title.trim() && userSelect) {
       setTodos(prev => [
         ...prev,
         {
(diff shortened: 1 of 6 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,21 @@
...
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  return (
+    <article
+      data-id={todo.id}
+      className={classNames('TodoInfo', {
+        'TodoInfo--completed': todo.completed,
+      })}
+    >
+      <h2 className="TodoInfo__title">{todo.title}</h2>
```
volodymyr-soltys97: Use destructuring for `todo`

src/utils/types.ts
```diff
@@ -0,0 +1,25 @@
...
+export interface ToDo {
+  id: number;
+  title: string;
+  completed: boolean;
+  user?: {
+    id: number;
+    name: string;
+    username: string;
+    email: string;
+  };
```
volodymyr-soltys97:
```suggestion
  user?: User;
```

volodymyr-soltys97 (changes requested):
Good job :
Let's improve your code
The title cannot be only spaces 
image

Also add a label text to input as on the working example here 
image

src/App.tsx
```diff
@@ -1,61 +1,132 @@
...
+    if (user) {
+      return {
+        user: { ...user },
+        ...todo,
+      };
+    }
+
+    return {
+      ...todo,
+    };
```
etojeDenys, line 24:
```suggestion
    return {
      ...todo,
      user
    };
```

etojeDenys (approved):

PR #3030, opened 2025-03-18
todo resolving

reb0rned:
https://reb0rned.github.io/react_add-todo-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,121 @@
 import './App.scss';
+import  usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { User } from './types/types';
+import { TodoList } from './components/TodoList';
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,61 +1,129 @@
...
+  }
+
+  const submitHandler = (event: React.FormEvent) => {
+    event.preventDefault()
+    setTitleHasError(title.trim() === '')
+    setUserHasError(userId === 0)
+
+    if (title === '' || userId === 0) {
+      return;
+    }
```
R-Ohirok: You remove spaces when setting an error, but do not remove them when checking for an error, this can cause the following situation: 
image
image

R-Ohirok (approved):
good job! But there may be an error

PR #3055, opened 2025-03-21
Solution

ENigmusSS:
[DEMO LINK](https://enigmusss.github.io/react_add-todo-form/)


maxbilfed (approved):
Good solution

ENigmusSS:
Recheck that. Data preparation should be in App file, but outside component function as constants. I also have zero oneletter variables in the code, despite that e for event is widespread standard. Also, ids is numbers, so default value let be too, they autoconverted after all, and everything works just fine.

PR #3295, opened 2025-08-12
add task solution

isapq:
https://isapq.github.io/react_add-todo-form/


pedro-ruas (changes requested):
Some tests are failing, one of then is because there is a missing `placeholder` in the input field.
This is required by the task

PR #3332, opened 2025-08-24
Solution

CyberDogFK:
[DEMO LINK](https://cyberdogfk.github.io/react_add-todo-form/)


src/App.tsx
```diff
@@ -1,61 +1,96 @@
...
+    }
+
+    const newToDo: ToDo = {
+      id: Math.max(...posts.map(post => post.id)) + 1,
+      title,
+      completed: false,
+      userId,
+    };
+
+    setPosts([...posts, newToDo]);
```
BudnikOleksii, line 35: Use functional setState to set state depending on the previous one

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,24 @@
...
+  return (
+    <article
+      data-id={todo.id}
+      className={cs('TodoInfo', {
+        'TodoInfo--completed': todo.completed,
+      })}
+    >
+      <h2 className="TodoInfo__title">{todo.title}</h2>
+
+      <UserInfo user={findUserById(todo.userId)} />
```
BudnikOleksii, line 21: It's better to pass a todo with a user info directly to the `TodoInfo` component instead of doing it there.
This component should be just a presentational("dummy") component, accept props and render UI, that's all

BudnikOleksii (approved):
Good work 👍 
Approved, but left a few comments

PR #3388, opened 2025-10-15
task solution

Pobedasg:
[DEMO LINK](https://Pobedasg.github.io/react_add-todo-form/)

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -98,6 +98,18 @@ module.exports = {
         },
       },
     ],
+
+    // Import
+    'import/extensions': [
+      'error',
+      'ignorePackages',
+      {
+        ts: 'never',
+        tsx: 'never',
+        js: 'never',
+        jsx: 'never',
+      },
+    ],
   },
   ignorePatterns: ['dist', '.eslintrc.cjs', 'vite.config.ts', 'src/vite-env.d.ts', 'cypress'],
   settings: {
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -98,18 +98,6 @@ module.exports = {
         },
       },
     ],
-
-    // Import
-    'import/extensions': [
-      'error',
-      'ignorePackages',
-      {
-        ts: 'never',
-        tsx: 'never',
-        js: 'never',
-        jsx: 'never',
-      },
-    ],
   },
   ignorePatterns: ['dist', '.eslintrc.cjs', 'vite.config.ts', 'src/vite-env.d.ts', 'cypress'],
   settings: {
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,6 +1,6 @@
 import classNames from 'classnames';
 import { UserInfo } from '../UserInfo/UserInfo';
-import { TodoWithUser } from '../../types/todoWithUser';
+import { TodoWithUser } from '../../types/todoWithUser.js';
 interface Props {
   todo: TodoWithUser;
 }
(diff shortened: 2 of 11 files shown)


MarinaKramarchuk (approved):
looks great, unfortunately there is no demo link to test

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,32 @@
...
+  todo: TodoWithUser;
+}
+
+export const TodoInfo = ({ todo }: Props) => {
+  const { id, title, completed, user } = todo;
+
+  return (
+    <article
+      data-id={id}
+      className={`TodoInfo${completed ? ' TodoInfo--completed' : ''}`}
```
Moroz-Dmytro: use classnames

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,34 @@
...
+  id: number;
+  title: string;
+  completed: boolean;
+  userId: number;
+  user?: {
+    id: number;
+    name: string;
+    username: string;
+    email: string;
+  };
```
Moroz-Dmytro: Create separate interface for User

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,34 @@
...
+}
+
+interface Props {
+  todos: Todo[];
+}
```
Moroz-Dmytro: Better move all interfaces to separate file

src/components/TodoList/TodoList.tsx
```diff
@@ -1 +1,34 @@
...
+  const todosWithUser = todos.filter(todo => todo.user);
+
+  if (!todosWithUser.length) {
+    return null;
+  }
+
+  return (
+    <section className="TodoList">
+      {todosWithUser.map(todo => (
+        <TodoInfo key={todo.id} todo={todo as Required<Todo>} />
```
Moroz-Dmytro:
```suggestion
        <TodoInfo key={todo.id} todo={todo} />
```

Try to avoid useing `as`, try to fix it in another way

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1 +1,16 @@
-export const UserInfo = () => {};
+interface Props {
+  user: {
+    id: number;
+    name: string;
+    username: string;
+    email: string;
+  };
```
Moroz-Dmytro: Create interface for User
Or check maybe it is alreadt created somewhere

Moroz-Dmytro (changes requested):

.eslintrc.cjs
```diff
@@ -98,6 +98,18 @@ module.exports = {
...
+    'import/extensions': [
+      'error',
+      'ignorePackages',
+      {
+        ts: 'never',
+        tsx: 'never',
+        js: 'never',
+        jsx: 'never',
+      },
+    ],
```
mkyloa: why do you need this?

src/components/TodoList/TodoList.tsx
mkyloa: I agree with @Moroz-Dmytro, let's avoid typecasting as hard as we can

src/types/Todo.ts
```diff
@@ -0,0 +1,9 @@
+import { User } from './user';
+
+export interface Todo {
+  id: number;
+  title: string;
+  completed: boolean;
+  userId: number;
+  user?: User;
```
mkyloa, line 8: is user really optional?

mkyloa (changes requested):

.eslintrc.cjs
Pobedasg: I need to configure import/extensions with that values so that ESLint (specifically, the eslint-plugin-import rule) does not require specifying file extensions (.ts, .js, etc.) for local imports. At least, I thought it would help me... I must have missed something somewhere, I'll go and take a closer look at the code.

mkyloa (approved):

PR #3402, opened 2025-10-17
add task solution

JackSavin:
[DEMO LINK](https://jacksavin.github.io/react_add-todo-form/)


S-Khraban (approved):
Good Job!

src/App.tsx
```diff
@@ -1,61 +1,143 @@
...
+
+    if (title.trim() === '') {
+      setTitleError(true);
+      hasError = true;
+    }
+
+    if (selectedUserId === '') {
+      setUserError(true);
+      hasError = true;
+    }
```
781calmstate, line 78: if you just want to return if `hasError` is true, then just return from `if`'s


```suggestion
    if (title.trim() === '') {
      setTitleError(true);
     return;
    }

    if (selectedUserId === '') {
      setUserError(true);
      return;
    }
```

src/App.tsx
```diff
@@ -1,61 +1,143 @@
...
+
+    if (hasError) {
+      return;
+    }
+
+    const numericUserId = Number(selectedUserId);
+    const newId =
+      currentTodos.length > 0
+        ? Math.max(...currentTodos.map(todo => todo.id)) + 1
+        : 1;
```
781calmstate, line 88: `1` is a magic number, i don't understand the purpose of it, consider to move those magic values into consts with descriptive naming

781calmstate (changes requested):
Don't be afraid to split the logic into components, for example you could move into separated component input with label and error, similarly for your select

PR #3404, opened 2025-10-18
add solution

vtmnkmarta:
[DEMO LINK](https://vtmnkmarta.github.io/react_add-todo-form/)


YuliiaKulish (approved):

src/types/nullable-type.ts
781calmstate, line 1: you capitalize only those files where you declare your components, the types folder should be in lower case as well as files inside of it.

you also should move the types folder from components. they should be on the same level like this:

src: 
- components
- types

src/types/todo.ts
781calmstate, line 1: same

src/components/Types/TodoAggregate.ts
781calmstate, line 1: same

src/types/user.ts
781calmstate, line 1: same

781calmstate (changes requested):
Pay more attention to commits naming. For now it's bad and non-descriptive. Read more about good practices [here](https://www.conventionalcommits.org/en/v1.0.0/)

Moroz-Dmytro (approved):

PR #3405, opened 2025-10-18
add task solution

YuliiaKulish:
[DEMO LINK](https://YuliiaKulish.github.io/react_add-todo-form/)


andriiLink (approved):
Great job! 
But it's better to make a new component for AddTodoForm

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1 +1,16 @@
...
+import Todo from '../../type/Todo';
+import { UserInfo } from '../UserInfo';
+interface Props {
+  todo: Todo;
+}
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => (
+  <article
+    data-id={todo.id}
+    className={`TodoInfo${todo.completed ? ' TodoInfo--completed' : ''}`}
```
781calmstate: use `classnames` lib for conditional classes

src/type/Todo.ts
```diff
@@ -0,0 +1,9 @@
+import User from './User';
+
+export default interface Todo {
```
781calmstate: types usually are not exported with `default`, for example in the future you could have more types for your todos, you don't need to create a new file for each new type

src/types/todo.ts
781calmstate, line 1: file names should only be capitalize if it's component's file or component's folder

src/types/user.ts
781calmstate, line 1: file names should only be capitalize if it's component's file or component's folder

src/type/User.ts
```diff
@@ -0,0 +1,6 @@
+export default interface User {
```
781calmstate: same about default

src/App.tsx
```diff
@@ -1,61 +1,137 @@
...
+    };
+  })
+  .filter((todo): todo is Todo => todo !== null);
+
+export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>(initialTodos);
+  const [selectedUserId, setSelectedUserId] = useState<number>(0);
+  const [title, setTitle] = useState<string>('');
+  const [titleError, setTitleError] = useState<boolean>(false);
+  const [userError, setUserError] = useState<boolean>(false);
```
781calmstate: variables that store boolean values should be named so they create some kind of question which can be answered with `true` or `false`. For example: `hasData`, `isVisible`, `matchesPattern` etc

781calmstate: we could also reduce two `useStates` to a single one that contains object.

For example:
```
const [erros, setErrors] = useState({ 
   user: 'You need to select user',
   title: 'You need to fill the title'
})
```

something similar to this

src/App.tsx
```diff
@@ -1,61 +1,137 @@
...
+          <label htmlFor="title">Title</label>
+          <input
+            id="title"
+            type="text"
+            data-cy="titleInput"
+            value={title}
+            onChange={(event: React.ChangeEvent<HTMLInputElement>) => {
+              setTitle(event.target.value);
+              setTitleError(false);
+            }}
```
781calmstate: if your inline event handler requires more than one line of code, it's probably should be moved into declared handle function

src/App.tsx
```diff
@@ -1,61 +1,137 @@
...
+          <select
+            id="user"
+            data-cy="userSelect"
+            value={selectedUserId}
+            onChange={(event: React.ChangeEvent<HTMLSelectElement>) => {
+              const newId = Number(event.target.value);
+
+              setSelectedUserId(newId);
+              setUserError(false);
+            }}
```
781calmstate: same about inline event handlers

781calmstate (changes requested):
Don't be afraid to split logic into components. React enhances component approach. In your case you could split into components logic for your fields.
For example `TitleField` and `UserField` each containing their label, field and error

781calmstate (approved):
Great job!

PR #3490, opened 2026-01-16
Solution

Br0No:
https://Br0No.github.io/react_add-todo-form/


src/App.tsx
```diff
@@ -1,61 +1,131 @@
...
+  const [selectedUser, setSelectedUser] = useState(0);
+  const [hasErrorSelectedUser, setHasErrorSelectedUser] = useState(false);
+  const [todos, setTodos] = useState<TodoWithUser[]>(
+    todosFromServer
+      .map(todo => {
+        const user = usersFromServer.find(person => person.id === todo.userId);
+
+        if (!user) {
+          return null;
+        }
```
Dimononon, line 22: I'm worried about this implementation. If you not found user you just return `null `instead of `Todo` with `user=null`?

Dimononon (approved):
I don't like  what you do with `Todo` when you not found `User`, but we don't have  instructions for this case so it's fine ig. Other than this all great!

PR #3494, opened 2026-01-16
implement the ability to add todos to the list

illia-001:
[DEMO LINK](https://illia-001.github.io/react_add-todo-form/)


Milana-Danylivska (approved):
Good job! The solution is clean and well-structured

PR #3509, opened 2026-01-19
Solution

mark-chep:


dmytro-halieba (approved):
Great job on this, the code works perfectly! I have a couple of small suggestions to make it even cleaner:

For handleChange, we could pass +event.target.value directly in the call. This way, we avoid doing the type conversion inside the function itself.

Also, if we attach the User object to the todos as soon as we set them, we won't have to manually add it every time it's needed later on.

PR #3589, opened 2026-03-25
commit message

alexsand-r:


Anton-Kuchmasov (changes requested):
Hey, to get your solution approved, you need to pass all tests. Take a look at those two:

Image

Would you mind to fix them?

If you're stuck, feel free to ask help in fe_chat

Anton-Kuchmasov (approved):
Seems like test stage on Github failed with unknown reason.
By the way, your solution is correct

PR #3632, opened 2026-05-05
implement solution

whomngmnt:
[DEMO LINK]  https://whomngmnt.github.io/react_add-todo-form/


src/App.tsx
```diff
@@ -1,61 +1,125 @@
...
+            value={title}
+            onChange={event => {
+              const normalizedTitle = event.target.value.replace(
+                /[^0-9A-Za-zА-Яа-яІіЇїЄєҐґ\s]/g,
+                '',
+              );
+
+              setTitle(normalizedTitle);
+              setIsTitleError(false);
+            }}
```
781calmstate, line 85: better to move into separate handlers inline functions that take more than one line.

related to all similar situations in your code

781calmstate (approved):

PR #3640, opened 2026-05-08
Solution

Tysiachnyk:
[DEMO LINK](https://Tysiachnyk.github.io/react_add-todo-form/)


src/App.tsx
```diff
@@ -1,61 +1,138 @@
...
+import { TodoList } from './components/TodoList';
+import { PreparedTodo, Todo } from './types/Todo';
+import { User } from './types/User';
+
+function getUser(users: User[], query: number): User | null {
+  return users.find(user => user.id === query) || null;
+}
+
+function getPreparedTodos(todos: Todo[]): PreparedTodo[] {
+  const copyTodos = [...todos];
```
DmitriyS2k, line 15: No need to copy the array here — `.map()` doesn't mutate the original, so [...todos] is redundant.

3. [CODE KNOWLEDGE] - If you are using a non-mutating array method, you **don't** need to create a copy of the array

DmitriyS2k (approved):
Overall great work!

dothearts (approved):

PR #3643, opened 2026-05-08
Solution

Misharkus:
[DEMO LINK](https://Misharkus.github.io/react_add-todo-form/)


src/App.tsx
```diff
@@ -1,61 +1,124 @@
 import './App.scss';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
+import { AggregatedTodo } from './types/Types';
+import { TodoList } from './components/TodoList';
+import { FormEvent, useState } from 'react';
+// import { event } from 'cypress/types/jquery';
```
dothearts: You don't need it, remove the comment

src/App.tsx
```diff
@@ -1,61 +1,124 @@
...
+
+const aggregatedTodos: AggregatedTodo[] = todosFromServer.map(todo => {
+  const foundUser =
+    usersFromServer.find(user => user.id === todo.userId) ?? null;
+
+  if (!foundUser) {
+    throw new Error(`Missing user in todo: ${todo.id}`);
```
dothearts: You don't need throw error in this case. Only need handle `null` value in component that render todo.

src/App.tsx
```diff
@@ -1,61 +1,124 @@
...
+    }
+
+    if (!newUserId) {
+      setSelectError('Please choose a user');
+      hasErrors = true;
+    }
+
+    if (hasErrors) {
+      return;
+    }
```
dothearts, line 53: You can create validation func, which return true or false. 
In this case submit function overload.

dothearts (changes requested):
In general, you did all right. 
Fix some small issues. 
Many thanks!

dothearts (approved):

PR #3657, opened 2026-05-13
add task solution

A-Melnykk:
[DEMO LINK](https://A-Melnykk.github.io/react_add-todo-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,123 @@
-import './App.scss';
+import React, { useState, useEffect } from 'react';
+import { Todo } from './types/Todo';
+import { User } from './types/User';
+import { TodoList } from './components/TodoList/TodoList';
+import usersFromServer from './api/users';
+import todosFromServer from './api/todos';
 
-// import usersFromServer from './api/users';
-// import todosFromServer from './api/todos';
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,61 +1,125 @@
...
-          <select data-cy="userSelect">
-            <option value="0" disabled>
-              Choose a user
-            </option>
-          </select>
+          <div className="control">
+            <button
+              type="submit"
+              data-cy="submitButton"
+              className={`button is-primary ${isSubmitting ? 'is-loading' : ''}`}
```
Denys-Kravchuk9988: It's better to use `classNames` here

Denys-Kravchuk9988 (changes requested):
Good job! 

A few things to improve:
1. There is needed to write errors that fields are required when user clicked `Add Todo` and not filled fields. And these messages should disapper once inputs are edited.
Image

2. User should be clickable and open mailbox
Image
Image

3. I would recommend to follow design as in example here
https://mate-academy.github.io/react_add-todo-form/

Also, pay attention to AI comments and try to pass all tests. If you need help - feel free to ask in the chat

brespect (approved):
Good job!

PR #3664, opened 2026-05-17
Solution

lilp3ka:
https://lilp3ka.github.io/react_add-todo-form/
demo


andriy-stetsula (approved):
Good job, but you need to break the code into components!

